PR #52, opened 2018-02-28
:hammer: Fixes #49 refactor template specialisation to function overloading

### LINKED ISSUE #49: Template specialisation with an integer doesn't work
Template specialisation with the dimension parameter doesn't work, it always calls the 3D case, even for the 2D case. The following case should call 2D, instead the call is to the 3D function
```
template <>
mpm::Cell<2> function(){ std::cout << "2D\n";}

template <>
mpm::Cell<2> function(){ std::cout << "3D\n";}
```
Expected behaviour for
```
auto cell = std::make_shared<mpm::Cell<3>>()
cell->function();
```
is to print `2D` instead prints `3D`.

Needs template specialisation done through SFINAE

kks32:
Fixes #49 by replacing `template <> cell<2>` and `template <> cell<3>` with an overloading of functions, except for compute volume where a switch statement with dimension as argument and some static_casting is required to keep Eigen happy.

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -138,118 +138,140 @@ void mpm::Cell<Tdim>::remove_particle_id(Index id) {
                    particles_.end());
 }
 
-//! Compute volume of a 2D cell
+//! Compute volume of a cell
 //! \retval volume Cell volume / area
 //! \tparam Tdim Dimension
-template <>
-inline void mpm::Cell<2>::compute_volume() {
+template <unsigned Tdim>
+inline void mpm::Cell<Tdim>::compute_volume() {
   try {
     Eigen::VectorXi indices = shapefn_->corner_indices();
-    // Quadrilateral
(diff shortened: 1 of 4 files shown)


include/cell.tcc
```diff
@@ -138,119 +138,141 @@ void mpm::Cell<Tdim>::remove_particle_id(Index id) {
...
+          //       | / h        | / g
+          //       |/           |/
+          //       *_ _ _ _ _ _ *
+          //     e               f
+          //
+          // Calculation of hexahedron volume from
+          // https://arc.aiaa.org/doi/pdf/10.2514/3.9013
+
+          const Eigen::Matrix<double, 3, 1> a =
+              static_cast<Eigen::Matrix<double, 3, 1>>(
```
js947: 😭

js947 (approved):

PR #70, opened 2018-04-26
Add gauss points for hex and quad elements, fix #68

### LINKED ISSUE #68: Gauss Points Implementation
Prepare the necessary files and functions for gauss point integration scheme in MPM

- Make a new class Gauss
- Change necessary functions in `particle.h` and `node.h`

Follow this [link ](https://docs.google.com/document/d/1UUTEbbgZsPed40KOf-oLHb2kaHg8N2F9FPSdR18ca9k/edit) for further documentation.

ezrayst:


include/gauss.h
```diff
@@ -0,0 +1,154 @@
+#ifndef MPM_PARTICLE_H_
```
kks32: Why is `gauss.h` has the Particle class?

CMakeLists.txt
```diff
@@ -55,7 +55,7 @@ add_executable(mpm
 SET(test_src
    ${mpm_SOURCE_DIR}/tests/cell_container_test.cc
    ${mpm_SOURCE_DIR}/tests/cell_test.cc
-   ${mpm_SOURCE_DIR}/tests/quad_shapefn_test.cc
+   ${mpm_SOURCE_DIR}/tests/hex_quad_test.cc
```
kks32: could we call quadrature as `quadrature` instead of `quad`

include/hex_quad.tcc
```diff
@@ -0,0 +1,217 @@
...
+template <unsigned Tdim, unsigned Tnquadratures>
+mpm::HexahedronQuadrature<Tdim, Tnquadratures>::HexahedronQuadrature()
+    : mpm::QuadratureBase<Tdim, Tnquadratures>() {
+
+  static_assert(Tdim == 3, "Invalid dimension for a 3D hexahedron element");
+  static_assert(
+      (Tnquadratures == 1) || (Tnquadratures == 8) || (Tnquadratures == 27),
+      "Invalid number of quadratures");
+
+  if (Tnquadratures == 1) {
```
kks32: Could we do partial specialisation instead?

include/quad_base.h
```diff
@@ -0,0 +1,48 @@
...
+#define MPM_QUADRATURE_BASE_H_
+
+#include <algorithm>
+#include <iostream>
+#include <limits>
+#include <vector>
+
+#include <Eigen/Dense>
+
+namespace mpm {
```
kks32: No need to predefine a class

include/quad_base.h
```diff
@@ -0,0 +1,48 @@
...
+template <unsigned Tdim, unsigned Tnquadratures>
+class QuadratureBase;
+}
+
+// Quadrature base class
+//! \brief Base class for quadrature
+//! \tparam Tdim Dimension
+//! \tparam Tnquadratures number of quadratures
+template <unsigned Tdim, unsigned Tnquadratures>
+class mpm::QuadratureBase {
```
kks32: This should be an abstract class

include/quad_base.h
```diff
@@ -0,0 +1,48 @@
...
+  //! Destructor
+  virtual ~QuadratureBase() {}
+
+  //! Return quadrature points
+  //! \param[out] qpoints Quadrature points in local coordinates
+  Eigen::Matrix<double, Tnquadratures, Tdim> quadratures() { return qpoints_; }
+
+  //! Return weights
+  //! \param[out] weights Weights for quadrature points
+  std::vector<double> weights() { return weights_; }
```
kks32: All vectors are Eigen vectors in the MPM code.

include/quad_base.h
```diff
@@ -0,0 +1,48 @@
+#ifndef MPM_QUADRATURE_BASE_H_
```
kks32: This file should be just called `quadrature.h`

kks32 (changes requested):

ezrayst:
I have not changed to partial specalization, but I am having trouble in getting the test cases to run. Will check again but it is giving seg-fault now.

kks32:
Hi @ezrayst Thank you for these. Could you try running your test cases with `./mpmtest -s` and `./mpmtest` without `-s`, sometimes it might be because your constructor fails. Do run `gdb --args ./mpmtest -s [quadrature]` if you want to just test the quadrature. you can also do `[2D][quadrature]` which will only test 2D cases with quadrature. In gdb `run` and then do `bt` to see where it fails. I'll have a look as well on your branch.

kks32:
Hi @ezrayst Your last commit seems to have removed / renamed files incorrectly, which results in compilation issues. Could you please check? Thanks!

include/gauss.h
kks32: Is this file required? It seems to be a copy of `Particle` class

include/hex_quadrature.tcc
```diff
@@ -0,0 +1,243 @@
...
+    weights(7) = 1.;
+
+  return weights;
+}
+
+// Getting the quadratures for Tnquadratures = 27
+template <>
+inline Eigen::MatrixXd mpm::HexahedronQuadrature<3, 27>::quadratures() {
+  Eigen::Matrix<double, 27, 3> quadratures;
+    const double Qpoint_a = std::sqrt(3. / 5.);
```
kks32: Hi @ezrayst could you please use a lower case for all variables, we no longer use capitalisation of `const` as `const`-ing is now very common than before. Thanks!

include/quadrature.h
```diff
@@ -0,0 +1,40 @@
...
+// Quadrature base class
+//! \brief Base class for quadrature
+//! \tparam Tdim Dimension
+//! \tparam Tnquadratures number of quadratures
+template <unsigned Tdim, unsigned Tnquadratures>
+class QuadratureBase {
+ public:
+  //! Constructor
+  //! Assign variables to zero
+  QuadratureBase() {}
```
kks32: Set this to default.

kks32 (changes requested):

include/hex_quadrature.tcc
ezrayst: Understood and changed!

kks32 (approved):

PR #76, opened 2018-06-12
JSON input for MPM - Fix #75

### LINKED ISSUE #75: Use JSON to handle MPM input parameters
* Read input for MPM (mesh files and input parameters) from a JSON file

ezrayst:
First push is up, but having trouble with the initialization with `json` [input](https://github.com/ezrayst/mpm/blob/67ded92806fad1c6b8af47dedf4992d78be4eff3/tests/io_test.cc#L45) - thus `mpmtest` does not run and it will fail circleci

@kks32, can you please comment on the structure too?


README.md
```diff
@@ -25,7 +25,7 @@
 ## Compile and Run
 > See https://mpm-doc.cb-geo.com/ for more detailed instructions. 
 
-0. Run `mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release /path/to/CMakeLists.txt`.
+0. Run `mkdir build && cmake -B /path/to/build/folder -DCMAKE_BUILD_TYPE=Release /path/to/CMakeLists.txt`.
```
kks32: Could you leave it as such. You create a build directory, you can `cd` to build to compile and build the files.

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+using json = nlohmann::json;
+
+//! \brief Input/Output handler
+class IO {
+ public:
+  //! Constructor
+  IO(json json_filename);
+
+  //! Return input mesh file name
+  std::string inputMeshFileName() const { return inputMeshFileName_; }
```
kks32: Please stick to our convention of naming variables and function names.

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+
+  //! Return input sub mesh file name
+  std::string inputSubMeshFileName() const { return inputSubMeshFileName_; }
```
kks32: Can we store the file names in a map and return the value for each key instead of multiple file names.

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+#define MPM_IO_H_
+
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <utility>
+
+#include <boost/filesystem.hpp>
+#include <eigen3/Eigen/Dense>
```
kks32: Why do we need eigen?

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+
+//! Alias for JSON
+#include "json.hpp"
+using json = nlohmann::json;
```
kks32: `Json`

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+  double dampingRatio() const { return dampingRatio_; }
+
+ private:
+  //! Input directory
+  std::string working_dir_;
+
+  //! Input json object
+  json json_;
+
+  //! Material properties json object
```
kks32: These variables are redundant if you store the `Json` object.

include/io.tcc
```diff
@@ -0,0 +1,261 @@
+#include "io.h"
```
kks32: This is not a templatised class, so please use `.cc`

kks32 (changes requested):
Please have a look at: https://github.com/cb-geo/lem/blob/develop/include/io.h

kks32:
I would suggest just pushing your input `json` file to start with, let's agree on the input JSON, before writing rest of the code.

README.md
ezrayst: OK! Understood!

ezrayst:
@kks32, I have added JSON file as an example. Will work from this.

bin/mpm.json
```diff
@@ -0,0 +1,30 @@
+{
+  "title": "Example JSON Input for MPM",
+  "input_directory" : "/inputFiles/",
```
kks32: `/inputFiles/` to `./path/to/input_files/`

bin/mpm.json
```diff
@@ -0,0 +1,30 @@
+{
+  "title": "Example JSON Input for MPM",
+  "input_directory" : "/inputFiles/", 
+  "output_directory" : "/results/",
```
kks32: Use a `post_processing` object. Include `output_directory` and `output_steps`. Change `/results` to `./results`

bin/mpm.json
```diff
@@ -0,0 +1,30 @@
...
+  "input_files" : {
```
kks32: Please keep all file names lower case

bin/mpm.json
```diff
@@ -0,0 +1,30 @@
...
+    "soil_particle_filename" : "soilParticles.dat",
+    "init_stress_soil_particle_filename" : "initStressSoilP.dat",
+    "material_filename" : "material.dat",
+    "traction_filename" : "Traction.dat"
+  },
+  "analysis" : {
+    "dt" : 0.001,
+    "number_steps" : 1000,
+    "number_substeps" : 10,
+    "gravity_flag" : 0,
```
kks32: Flags should be boolean

kks32 (changes requested):

bin/mpm.json
ezrayst: Yes, thus we have 0 or 1? Or do you want "true" or "false"?

include/io.h
```diff
@@ -0,0 +1,149 @@
...
+
+//! Alias for JSON
+#include "json.hpp"
+using json = nlohmann::json;
+
+//! \brief Input/Output handler
+class IO {
+ public:
+  //! Constructor
+  IO(json json_filename);
```
kks32: IO constructor should be `  IO(const int argc, char** argv);`

kks32 (changes requested):

PR #86, opened 2018-06-20
:hammer: Fix incorrect input in 3D Read mesh and Add 2D test case

kks32:


tests/read_mesh_ascii_test.cc
```diff
@@ -7,7 +7,195 @@ using Json = nlohmann::json;
...
+      }
+      file << "\n";
+    }
+
+    file.close();
+
+    // Check read mesh nodes
+    SECTION("Check read mesh nodes") {
+      // Dimension
+      const unsigned dim = 2;
```
cbgeo: Do we need to repeat `dim` and Tolerance?

tests/read_mesh_ascii_test.cc
```diff
@@ -7,7 +7,195 @@ using Json = nlohmann::json;
...
+    for (const auto& coord : coordinates) {
+      for (unsigned i = 0; i < coord.size(); ++i) {
+        file << coord[i] << "\t";
+      }
+      file << "\n";
+    }
+
+    // Write cell node ids
+    for (const auto& cell : cells) {
+      for (const auto nid : cell) {
```
cbgeo: Why is this not passed by reference

cbgeo (changes requested):

cbgeo (approved):

PR #87, opened 2018-06-21
Mesh create nodes from a list of coordinates

kks32:
* Use a factory to generate the correct type of nodes

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -1,0 +1,7 @@
+#include "node.h"
+#include "factory.h"
+#include "node_base.h"
+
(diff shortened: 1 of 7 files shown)


src/node.cc
```diff
@@ -0,0 +1,11 @@
+#include "node.h"
+#include "factory.h"
+#include "node_base.h"
+
```
cbgeo: Add comments to describe what these nodes are?

cbgeo (changes requested):

cbgeo (approved):

PR #110, opened 2018-07-19
:dart: Add additional test cases for newton raphson

kks32:
* Test points in negative coordinates


kks32:
PR contains other files not part of test case, remove and resubmit PR

PR #127, opened 2018-07-23
:computer: Get a particular property of a material

kks32:
Add a function to return density and other material properties.


cbgeo (changes requested):
Maybe move Json properties to base class?

cbgeo (approved):

PR #222, opened 2018-08-14
:checkered_flag: Assign velocity boundary to cells

### LINKED ISSUE #207: Velocity constraint on non-Cartesian elements
Current implementation applies velocity constraints at the nodes. Input includes `node id`, `direction = 0, 1 2` and `prescribed velocity`. This does not work for non uniform mesh.

Proposed summary of changes:
- Input would be `cell id`, `planeid`, `direction` in respect to the normal of face and `prescribed velocity` with positive being into the element
- `plane id` would be added to each element which contains 6 planes for hexahedron and 4 planes for quadrilateral, each with 1 dimension less than the dimension of the element. For instance, plane 0 of hexahedron consist of the plane made of nodes 0, 1, 4 and 5. 
- Currently, velocity constraint is stored in `nodes` but this `cell velocity constraint` will be in `mesh`
- It is applied during the mapping back from `nodes` to `particles`. So the application of velocity constraint depends on which element a particle belongs to
- During application, there will be rotation of the current velocity and acceleration to the constraint, and rotate back to the cartesian

ezrayst:
- [x] Add face to `elements`
- [x] Assign velocity boundary to `cell` 
- [x] Rotation matrix
- [x] Find normal of face having velocity boundary
- [ ] Apply boundary conditions to velocity and acceleration mapping from node to particle
- [ ] Apply boundary conditions when calculating strain


include/cell.h
```diff
@@ -54,7 +54,7 @@ class Cell {
   bool initialise();
 
   //! Return the initialisation status of cells
-  //! \retval initialisation_status Cell has nodes, element type and volumes
+  //! \retval initialisation_status Cell has nodes, shape functions and volumes
```
kks32: This should be `element type`

include/cell.h
```diff
@@ -218,6 +218,35 @@ class Cell {
...
+  //! Assign velocity constraint
+  //! Directions can take values between 0 and Dim-1
+  //! \param[in] face_id Face of cell of velocity constraint
+  //! \param[in] dir Direction of velocity constraint
+  //! \param[in] velocity Applied velocity constraint
+  bool assign_velocity_constraint(unsigned face_id, unsigned dir,
+                                  double velocity);
+
+  //! Compute normal vector
+  //! \param[in] face_id Face of cell of velocity constraint
```
kks32: This looks inconsistent, there are no parameters passed to this function.

include/cell.h
```diff
@@ -218,6 +218,35 @@ class Cell {
...
+  void compute_normal();
+
+  //! Return number of normal vectors
+  //! \retval number of normal vectors
+  unsigned nnormal() const { return velocity_constraints_normals_.size(); }
+
+  //! Return unit normal vector, positive pointing outside of the element
+  //! \param[in] id of constraint
```
kks32: What id? Face id?

include/cell.h
```diff
@@ -245,8 +274,16 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  std::vector<std::tuple<unsigned, unsigned, double>> velocity_constraints_;
```
kks32: Could you add a comment on what these unsigned variables are?

include/cell.h
```diff
@@ -245,8 +274,16 @@ class Cell {
...
+
+  //! Normal of face with velocity constraints
+  std::vector<Eigen::VectorXd> velocity_constraints_normals_;
```
kks32: Shall we just call these `face_normals_`?

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+//! Assign velocity constraint
+//! Constrain directions can take values between 0 and Dim-1
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
+                                                 double velocity) {
+  bool status = true;
+  try {
+    //! Constrain directions can take values between 0 and Dim-1
+    if (dir >= 0 && dir < (Tdim)) {
```
kks32: Why can't we have to `Tnphases * Tdim`?

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    indices = element_->face_indices(face_id);
+
+    // Compute the vector to calculate normal (perpendicular)
+    // a = node(0) - node(1)
+    a = (this->nodes_[indices(0)])->coordinates() -
+        (this->nodes_[indices(1)])->coordinates();
+
+    // Compute normal and make unit vector
+    // normal = a x b
+    // Note that definition of a and b are such that normal is always out of
```
kks32: This is not very clear to me on what you are doing.

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    // page
+    normal_vector(0) = -a(1);
+    normal_vector(1) = a(0);
+    normal_vector /= normal_vector.norm();
+
+    // Store to private variable
+    velocity_constraints_normals_.emplace_back(normal_vector);
```
kks32: It would be easier to have all normals to all faces. Or have a map of normal to each face.

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+//! Constrain directions can take values between 0 and Dim-1
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
+                                                 double velocity) {
+  bool status = true;
+  try {
+    //! Constrain directions can take values between 0 and Dim-1
+    if (dir >= 0 && dir < (Tdim)) {
+      this->velocity_constraints_.emplace_back(
+          std::tuple<unsigned, unsigned, double>(face_id, dir, velocity));
```
kks32: Could you store it as a map of `unsigned` and a `pair`?

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    // Get face_id
+    face_id = std::get<0>(velocity_constraint);
+
+    // Get the nodes of the face
+    indices = element_->face_indices(face_id);
+
+    // Compute two vectors to calculate normal
+    // a = node(1) - node(0)
+    // b = node(3) - node(0)
+    a = (this->nodes_[indices(1)])->coordinates() -
```
kks32: These are consts

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    // Store to private variable
+    velocity_constraints_normals_.emplace_back(normal_vector);
+  }
+}
+
+//! Compute normal 3d
+template <>
+inline void mpm::Cell<3>::compute_normal() {
+
+  Eigen::Matrix<double, 3, 1> a, b, normal_vector;
```
kks32: Please write like C++, use RAII

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+  Eigen::VectorXi indices;
+  mpm::Index face_id;
+
+  for (const auto& velocity_constraint : this->velocity_constraints_) {
+    // Get face_id
+    face_id = std::get<0>(velocity_constraint);
```
kks32: const auto

include/cell.tcc
kks32: I see what you mean by confusion with () and []. Let's use [] for vectors and () for matrices

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+Eigen::VectorXd mpm::Cell<Tdim>::normal(unsigned id) {
+
+  Eigen::Matrix<double, Tdim, 1> normal_vector;
+  normal_vector.setZero();
+
+  try {
+    // Check if id is within the range of normal vector
+    if (id < this->velocity_constraints_normals_.size()) {
+      // return normal vector depending on id
+      normal_vector = this->velocity_constraints_normals_[id];
```
kks32: shouldn't this be .at(id)?

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+  return normal_vector;
+}
+
+//! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
+template <>
+inline Eigen::MatrixXd mpm::Cell<2>::compute_inverse_rotation_matrix(
+    double alpha, double beta, double gamma) {
+
+  Eigen::Matrix<double, 3, 3> rotation_matrix_3d;
```
kks32: Declare where you use it.

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+  // beta = 0 implies no rotation on z axis
+  if (fabs(beta) > 1.E-12) beta = 0;
```
kks32: use `std::fabs`

kks32: beta = 0.

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+}
+
+//! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
+template <>
+inline Eigen::MatrixXd mpm::Cell<2>::compute_inverse_rotation_matrix(
+    double alpha, double beta, double gamma) {
+
+  Eigen::Matrix<double, 3, 3> rotation_matrix_3d;
+
+  // beta = 0 implies no rotation on z axis
```
kks32: // Why is beta z-axis?

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+  if (fabs(beta) > 1.E-12) beta = 0;
+
+  // clang-format off
+  rotation_matrix_3d << cos(alpha)*cos(gamma) - sin(alpha)*cos(beta)*sin(gamma),  -cos(alpha)*sin(gamma) - sin(alpha)*cos(beta)*cos(gamma),   sin(beta)*sin(alpha),
```
kks32: Precompute cos and sin

include/cell.tcc
```diff
@@ -926,3 +926,158 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+                        sin(alpha)*cos(gamma) + cos(alpha)*cos(beta)*sin(gamma),  -sin(alpha)*sin(gamma) + cos(alpha)*cos(beta)*cos(gamma),  -sin(beta)*cos(alpha),
+                        sin(beta)*sin(gamma),                                      sin(beta)*cos(gamma),                                      cos(beta);                                 
+  // clang-format on  
+
+  // Take block 2x2 matrix 
+  Eigen::Matrix<double, 2, 2> inverse_rotation_matrix;
+  inverse_rotation_matrix = rotation_matrix_3d.block<2, 2>(0, 0);
```
kks32, line 1078: Why not just construct a 2x2 matrix?

include/hexahedron_element.h
```diff
@@ -2,6 +2,7 @@
 #define MPM_HEXAHEDRON_ELEMENT_H_
 
 #include <exception>
+#include <iostream>
```
kks32: Remove

include/hexahedron_element.tcc
```diff
@@ -466,4 +466,69 @@ inline Eigen::MatrixXi
   //clang-format on
   return indices;
 }
+
+
+//!            Behind: F4
```
kks32, line 523: You would need to move this comment to the header

include/hexahedron_element.tcc
```diff
@@ -466,4 +466,69 @@ inline Eigen::MatrixXi
...
+//!    F3 |  |         |  |   F1
+//!       |  0_ _ _ _ _|_ 0
+//!       | / 0        | / 1
+//!       |/     F0    |/
+//!       0_ _ _ _ _ _ 0
+//!     4               5  
+//!         Front: F5 
+//!
+//!
+//! Description:
```
kks32: REmove this description line

include/hexahedron_element.tcc
```diff
@@ -466,4 +466,69 @@ inline Eigen::MatrixXi
...
+//!       |  0_ _ _ _ _|_ 0
+//!       | / 0        | / 1
+//!       |/     F0    |/
+//!       0_ _ _ _ _ _ 0
+//!     4               5  
+//!         Front: F5 
+//!
+//!
+//! Description: 
+//! Bottom face: F0
```
kks32: Can you make this a single or a double line
//! {Bottom: F0, Right: F1, ...}

include/hexahedron_element.tcc
```diff
@@ -466,4 +466,69 @@ inline Eigen::MatrixXi
...
+//! Top face: F2
+//! Left face: F3
+//! Behind face: F4
+//! Front face: F5
+//! Return indices of a face of the element
+template <unsigned Tdim, unsigned Tnfunctions>
+inline Eigen::VectorXi
+    mpm::HexahedronElement<Tdim, Tnfunctions>::face_indices(unsigned face_id) const {
   
+  Eigen::Matrix<int, 4, 1> indices;
```
kks32: Store it as an Eigen::MatrixXd and return the row.

include/mesh.h
```diff
@@ -1,6 +1,7 @@
 #ifndef MPM_MESH_H_
 #define MPM_MESH_H_
 
+#include <iostream>
```
kks32, line 4: Remove

include/mesh.h
```diff
@@ -148,12 +149,19 @@ class Mesh {
...
+  //! Assign velocity constraints to nodes
   //! \param[in] velocity_constraints Constraint at node, dir, and velocity
   bool assign_velocity_constraints(
       const std::vector<std::tuple<mpm::Index, unsigned, double>>&
           velocity_constraints);
 
+  //! Assign velocity constraints to cells
+  //! \param[in] velocity_constraints Constraint at cell id, phase id, dir, and
+  //! velocity
+  bool assign_velocity_constraints(
```
kks32: Rename to add cells to function `assign_cell_velocity_constraints`

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
+  bool status = false;
+  try {
+    if (nodes_.size()) {
+      for (const auto& velocity_constraint : velocity_constraints) {
+        // Cell id
+        mpm::Index cell_id = std::get<0>(velocity_constraint);
+        // Face id
+        mpm::Index face_id = std::get<1>(velocity_constraint);
+        // Direction of the local coordinate system of the face
+        // Tdim = 2: Normal = y local axis, = 1
```
kks32, line 366: What do you mean?

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
+          throw std::runtime_error(
+              "Cell or face or velocity constraint is invalid");
+      }
+    } else {
+      throw std::runtime_error(
+          "No cells have been assigned in mesh, cannot assign velocity "
+          "constraints");
+    }
+  } catch (std::exception& exception) {
+    std::cerr << exception.what() << "\n";
```
kks32: Use console_ logger

include/quadrilateral_element.h
```diff
@@ -2,6 +2,7 @@
 #define MPM_QUADRILATERAL_ELEMENT_H_
 
 #include <exception>
+#include <iostream>
```
kks32: Remove this

include/quadrilateral_element.tcc
```diff
@@ -360,3 +360,45 @@ inline Eigen::MatrixXi
   return indices;
 }
   
+//!          F2
```
kks32, line 415: Move this to header

include/quadrilateral_element.tcc
```diff
@@ -360,3 +360,45 @@ inline Eigen::MatrixXi
...
+//!     |          |
+//!     |          |
+//!   0 0----------0 1
+//!          F0
+//! Return indices of a face of the element
+template <unsigned Tdim, unsigned Tnfunctions>
+inline Eigen::VectorXi
+    mpm::QuadrilateralElement<Tdim, Tnfunctions>::face_indices(unsigned face_id) const {
+  
+  Eigen::Matrix<int, 2, 1> indices;
```
kks32: Same as Hexahedron, change to matrices

kks32 (changes requested):
Over all looks good, Could you please do the requested changes.

include/cell.tcc
kks32: I think the reason why I did square braces was that it looked consistent with the nodes access. So can we use [] for indices, thanks

kks32 (changes requested):

include/cell.tcc
kks32: Ignore previous comments about (), leave it as is.

kks32 (changes requested):

include/cell.tcc
ezrayst: cannot be const I think, it is within a for loop

include/cell.tcc
ezrayst: Because `Tnphases` is not a template parameter of `cell`. Should we add it?

ezrayst: Anyways, the value should be between 0 and Tdim * Tnphases - 1 to be more exact

kks32: That's purely a check, so no need to add a template parameter, just don't do that check

kks32 (changes requested):

include/cell.h
```diff
@@ -222,6 +222,34 @@ class Cell {
...
+  //! \param[in] velocity Applied velocity constraint
+  bool assign_cell_velocity_constraint(unsigned face_id, unsigned dir,
+                                       double velocity);
+
+  //! Compute normal vector
+  void compute_normal();
+
+  //! Return number of normal vectors
+  //! \retval number of normal vectors
+  unsigned nnormal() const { return face_normals_.size(); }
```
kks32: Why do we need this?

include/cell.h
```diff
@@ -222,6 +222,34 @@ class Cell {
...
+  //! Assign velocity constraint
+  //! Directions can take values between 0 and Dim-1
+  //! \param[in] face_id Face of cell of velocity constraint
+  //! \param[in] dir Direction of velocity constraint
+  //! \param[in] velocity Applied velocity constraint
+  bool assign_cell_velocity_constraint(unsigned face_id, unsigned dir,
+                                       double velocity);
+
+  //! Compute normal vector
+  void compute_normal();
```
kks32: just 1 normal?

include/cell.h
```diff
@@ -222,6 +222,34 @@ class Cell {
...
+
+  //! Return number of normal vectors
+  //! \retval number of normal vectors
+  unsigned nnormal() const { return face_normals_.size(); }
+
+  //! Return unit normal vector, positive pointing outside of the element
+  //! \param[in] face_id of constraint
+  //! \retval unit normal vector
+  Eigen::VectorXd normal(unsigned face_id);
```
kks32, line 243: Why is this a public function?

include/cell.h
```diff
@@ -222,6 +222,34 @@ class Cell {
...
+
+  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! \param[in] alpha Euler alpha angle in radians
+  //! \param[in] beta Euler beta angle in radians
+  //! \param[in] gamma Euler gamma angle in radians
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::MatrixXd compute_inverse_rotation_matrix(double alpha, double beta,
```
kks32, line 250: I thought we were moving these to `geometry.h`

include/cell.h
```diff
@@ -249,8 +277,17 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
+  std::vector<std::tuple<unsigned, unsigned, double>> velocity_constraints_;
+
+  //! Normal of face with velocity constraints
+  std::vector<Eigen::VectorXd> face_normals_;
```
kks32: We discussed to have this as a map?

include/cell.h
```diff
@@ -249,8 +277,17 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
+  std::vector<std::tuple<unsigned, unsigned, double>> velocity_constraints_;
```
kks32: this should be a map

kks32 (changes requested):

ezrayst:
Krishna, have not changed everything I need to change. I would contact you once I do that.

include/cell.h
ezrayst: Not sure actually we can do this...

Because the first `unsigned` is face id, and for a face id, there can be multiple sets of direction and velocity. Thus it might not work when we call them? What do you think?

kks32: Can we not have a map of a pair?

kks32:
@ezrayst I get what you mean by you can't use a map, thanks!

ezrayst:
Thanks. I think a `cell` would only have maximum of 6 faces, so O(6) is not too bad compared to O(1). Is `vector` of `tuple` OK then?

kks32:
It will be O(18) as the worst case performance for each time we try to find the element. So technically it's not a single O(n), but each time the worst performnace is O(n), I'm fine with vector of tupled.

kks32:
@ezrayst should we close this PR?

kks32:
Closing this PR as #242 supersedes this. @ezrayst Let me know if this is closed incorrectly.

PR #230, opened 2018-08-21
Add face id in element

ezrayst:
- Node indices for face in element
- Make them in namespace
- Add tests

### DIFF
--- a/include/element.h
+++ b/include/element.h
@@ -132,6 +132,11 @@ class Element {
   //! to check if a point is inside /outside of a hedron
   //! \retval indices Indices that form sub-tetrahedrons
   virtual Eigen::MatrixXi inhedron_indices() const = 0;
+
+  //! Return indices of a face of an element
+  //! \param[in] face_id given id of the face
+  //! \retval indices Indices that make the face
+  virtual Eigen::VectorXi face_indices(unsigned face_id) const = 0;
 };
 
 }  // namespace mpm
--- a/include/element.h
+++ b/include/element.h
@@ -138,7 +138,7 @@ class Element {
   //! Return indices of a face of an element
   //! \param[in] face_id given id of the face
   //! \retval indices Indices that make the face
-  virtual Eigen::MatrixXi face_indices(unsigned face_id) const = 0;
+  virtual Eigen::VectorXi face_indices(unsigned face_id) const = 0;
 };
 
 }  // namespace mpm
--- a/include/element.h
+++ b/include/element.h
@@ -1,6 +1,8 @@
 #ifndef MPM_ELEMENT_H_
 #define MPM_ELEMENT_H_
 
+#include <exception>
+#include <map>
 #include <vector>
 
 #include <Eigen/Dense>
--- a/include/element.h
+++ b/include/element.h
@@ -136,8 +136,7 @@ class Element {
   //! Return indices of a face of an element
   //! \param[in] face_id given id of the face
   //! \retval indices Indices that make the face
-  virtual Eigen::Matrix<unsigned, 2 * (Tdim - 1), 1> face_indices(
-      unsigned face_id) const = 0;
+  virtual Eigen::MatrixXi face_indices(unsigned face_id) const = 0;
 };
 
 }  // namespace mpm
--- a/include/element.h
+++ b/include/element.h
@@ -136,7 +136,8 @@ class Element {
   //! Return indices of a face of an element
   //! \param[in] face_id given id of the face
   //! \retval indices Indices that make the face
-  virtual Eigen::VectorXi face_indices(unsigned face_id) const = 0;
+  virtual Eigen::Matrix<unsigned, 2 * (Tdim - 1), 1> face_indices(
+      unsigned face_id) const = 0;
 };
 
 }  // namespace mpm
--- a/include/hexahedron_element.h
+++ b/include/hexahedron_element.h
@@ -242,13 +242,16 @@ class HexahedronElement : public Element<Tdim> {
 
  private:
   //! Constants of face id and associate nodes
-  const std::map<unsigned, Eigen::Matrix<int, 4, 1>> face_indices_hexahedron_{
-      {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
-      {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
-      {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
-      {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
-      {4, Eigen::Matrix<int, 4, 1>(0, 1, 2, 3)},
-      {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}};
+  const std::map<unsigned, Eigen::Matrix<int, 8, 1>> face_indices_hexahedron_{
+      {0, (Eigen::Matrix<int, 8, 1>() << 0, 1, 5, 4, 8, 12, 16, 10).finished()},
+      {1,
+       (Eigen::Matrix<int, 8, 1>() << 5, 1, 2, 0, 12, 11, 14, 18).finished()},
+      {2,
+       (Eigen::Matrix<int, 8, 1>() << 7, 6, 2, 3, 19, 14, 13, 15).finished()},
+      {3, (Eigen::Matrix<int, 8, 1>() << 0, 4, 7, 3, 10, 17, 15, 9).finished()},
+      {4, (Eigen::Matrix<int, 8, 1>() << 1, 0, 3, 2, 8, 9, 13, 11).finished()},
+      {5,
+       (Eigen::Matrix<int, 8, 1>() << 4, 5, 6, 7, 16, 18, 19, 17).finished()}};
 
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
--- a/include/hexahedron_element.h
+++ b/include/hexahedron_element.h
@@ -2,6 +2,7 @@
 #define MPM_HEXAHEDRON_ELEMENT_H_
 
 #include <exception>
+#include <map>
 
 #include <Eigen/Dense>
 
@@ -80,6 +81,43 @@ namespace mpm {
 //!
 //!
 //! </pre>
+//!
+//! Face numbering for 8-node, 20-node and 27-node Hexaheron Element \n
+//!
+//! <pre>
+//!            Behind: F4
+//!        3      F2       2
+//!          0_ _ _ _ _ _0
+//!         /|           /|
+//!        / |          / |
+//!     7 0_ |_ _ _ _ _0 6|
+//!       |  |         |  |
+//!    F3 |  |         |  |   F1
+//!       |  0_ _ _ _ _|_ 0
+//!       | / 0        | / 1
+//!       |/     F0    |/
+//!       0_ _ _ _ _ _ 0
+//!     4               5
+//!         Front: F5
+//!
+//!
+//! Bottom face: F0, Right face: F1, Top face: F2,
+//! Left face: F3, Behind face: F4, Front face: F5
+//! </pre>
+//!
+//! Namespace containing constants of face id
+
+const std::map<unsigned, Eigen::VectorXi> face_indices_hexahedron{
+
+    {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
+    {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
+    {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
+    {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
+    {4, Eigen::Matrix<int, 4, 1>(0, 1, 2, 3)},
+    {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}
+
+};
+//!
 //! \tparam Tdim Dimension
 //! \tparam Tnfunctions Number of functions
 template <unsigned Tdim, unsigned Tnfunctions>
@@ -209,6 +247,11 @@ class HexahedronElement : public Element<Tdim> {
   //! to check if a point is inside /outside of a hedron
   //! \retval indices Indices that form sub-tetrahedrons
   Eigen::MatrixXi inhedron_indices() const override;
+
+  //! Return indices of a face of an element
+  //! \param[in] face_id given id of the face
+  //! \retval indices Indices that make the face
+  Eigen::VectorXi face_indices(unsigned face_id) const override;
 
  private:
   //! Logger
--- a/include/hexahedron_element.h
+++ b/include/hexahedron_element.h
@@ -105,19 +105,6 @@ namespace mpm {
 //! Left face: F3, Behind face: F4, Front face: F5
 //! </pre>
 //!
-//! Namespace containing constants of face id
-
-const std::map<unsigned, Eigen::VectorXi> face_indices_hexahedron{
-
-    {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
-    {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
-    {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
-    {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
-    {4, Eigen::Matrix<int, 4, 1>(0, 1, 2, 3)},
-    {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}
-
-};
-//!
 //! \tparam Tdim Dimension
 //! \tparam Tnfunctions Number of functions
 template <unsigned Tdim, unsigned Tnfunctions>
@@ -254,6 +241,15 @@ class HexahedronElement : public Element<Tdim> {
   Eigen::VectorXi face_indices(unsigned face_id) const override;
 
  private:
+  //! Namespace containing constants of face id
+  const std::map<unsigned, Eigen::VectorXi> face_indices_hexahedron{
+      {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
+      {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
+      {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
+      {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
+      {4, Eigen::Matrix<int, 4, 1>(0, 1, 2, 3)},
+      {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}};
+
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
 };
--- a/include/hexahedron_element.h
+++ b/include/hexahedron_element.h
@@ -238,17 +238,18 @@ class HexahedronElement : public Element<Tdim> {
   //! Return indices of a face of an element
   //! \param[in] face_id given id of the face
   //! \retval indices Indices that make the face
-  Eigen::VectorXi face_indices(unsigned face_id) const override;
+  Eigen::Matrix<unsigned, 2 * (Tdim - 1), 1> face_indices(
+      unsigned face_id) const override;
 
  private:
-  //! Namespace containing constants of face id
-  const std::map<unsigned, Eigen::VectorXi> face_indices_hexahedron{
-      {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
-      {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
-      {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
-      {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
-      {4, Eigen::Matrix<int, 4, 1>(0, 1, 2, 3)},
-      {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}};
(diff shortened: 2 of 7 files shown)


include/quadrilateral_element.h
```diff
@@ -53,6 +54,31 @@ namespace mpm {
...
+//!     |          |
+//!     |          |
+//!   0 0----------0 1
+//!          F0
+//! </pre>
+//!
+//!
+//! Namespace containing constants of face id
+
+const std::map<unsigned, Eigen::VectorXi> face_indices_quadrilateral{
```
kks32: This isn't even declared in the class?

include/quadrilateral_element.tcc
```diff
@@ -412,3 +412,33 @@ inline Eigen::MatrixXi
...
+  // Make return_indices
+  Eigen::Matrix<int, 2, 1> return_indices;
+  try {
+    // Check if face_id is within range
+    if (face_id < face_indices_quadrilateral.size()) {
+      return_indices = face_indices_quadrilateral.find(face_id)->second;
+    } else {
+      return_indices = face_indices_quadrilateral.find(0)->second;
+      throw std::runtime_error(
+          "Face ID is undefined, using default value of 0.");
```
kks32: I don't see a reason to set a default value.

include/quadrilateral_element.tcc
```diff
@@ -412,3 +412,33 @@ inline Eigen::MatrixXi
...
+template <unsigned Tdim, unsigned Tnfunctions>
+inline Eigen::VectorXi
+    mpm::QuadrilateralElement<Tdim, Tnfunctions>::face_indices(unsigned face_id) const {
+  
+  // Make return_indices
+  Eigen::Matrix<int, 2, 1> return_indices;
+  try {
+    // Check if face_id is within range
+    if (face_id < face_indices_quadrilateral.size()) {
+      return_indices = face_indices_quadrilateral.find(face_id)->second;
```
kks32: you are iterating over a map, no reason to have a try block, map will throw

kks32 (changes requested):
face indices should be defined in the class of each element not outside.

include/quadrilateral_element.tcc
ezrayst: Understood. Just throw error then

include/quadrilateral_element.h
ezrayst: I will make it to private variable then?

include/hexahedron_element.h
```diff
@@ -167,7 +192,21 @@ class HexahedronElement : public Element<Tdim> {
...
   Eigen::MatrixXi inhedron_indices() const override;
 
+  //! Return indices of a face of an element
+  //! \param[in] face_id given id of the face
+  //! \retval indices Indices that make the face
+  Eigen::VectorXi face_indices(unsigned face_id) const override;
+
  private:
+  //! Namespace containing constants of face id
+  const std::map<unsigned, Eigen::VectorXi> face_indices_hexahedron{
```
kks32: This is a private variable so end with `_`

kks32: Great! Nice design making this a const private

include/hexahedron_element.tcc
```diff
@@ -518,4 +518,44 @@ inline Eigen::MatrixXi
...
+//!       0_ _ _ _ _ _ 0
+//!     4               5  
+//!         Front: F5 
+//!
+//!
+//! Bottom face: F0, Right face: F1, Top face: F2, 
+//! Left face: F3, Behind face: F4, Front face: F5
+//! Return indices of a face of the element
+template <unsigned Tdim, unsigned Tnfunctions>
+inline Eigen::VectorXi
```
kks32: Return `Eigen::Matrix<int, 4, 1>` `int` or `unsigned`?

include/quadrilateral_element.tcc
```diff
@@ -412,3 +412,32 @@ inline Eigen::MatrixXi
...
+//!   3 0----------0 2
+//!     |          |
+//!  F3 |          | F1
+//!     |          |
+//!     |          |
+//!   0 0----------0 1
+//!          F0
+//! Return indices of a face of the element
+template <unsigned Tdim, unsigned Tnfunctions>
+inline Eigen::VectorXi
```
kks32: Return `Eigen::Matrix<int, 2, 1>`

kks32: or <Tdim, 1>

tests/hexahedron_element_test.cc
```diff
@@ -764,6 +764,18 @@ TEST_CASE("Hexahedron elements are checked", "[hex][element][3D]") {
...
       REQUIRE(indices(11, 2) == 2);
     }
+
+    SECTION("Eight noded hexahedron shape function for face indices") {
+      // Check for face indices
+      unsigned face_id = 0;
+      Eigen::VectorXi indices = hex->face_indices(face_id);
+      REQUIRE(indices.size() == 4);
+
+      REQUIRE(indices(0) == 0);
```
kks32: Don't you have to check all of the faces?

kks32 (changes requested):

include/hexahedron_element.h
ezrayst: Thanks. Understood. I forgot about `_`

include/hexahedron_element.tcc
ezrayst: I think the best is to use `unsigned`. Because of previous design using Eigen's typedef like `VectorXi` or `MatrixXi`, I used `int`. I have reverted back to `unsigned`

include/quadrilateral_element.tcc
ezrayst: Not <`Tdim`, 1>. If we want to make it general, it's `(Tdim - 1) *2`

tests/hexahedron_element_test.cc
ezrayst: OK! I will check for all faces then

include/element.h
```diff
@@ -88,6 +88,12 @@ class Element {
   //! to check if a point is inside /outside of a hedron
   //! \retval indices Indices that form sub-tetrahedrons
   virtual Eigen::MatrixXi inhedron_indices() const = 0;
+
+  //! Return indices of a face of an element
+  //! \param[in] face_id given id of the face
+  //! \retval indices Indices that make the face
+  virtual Eigen::Matrix<unsigned, 2 * (Tdim - 1), 1> face_indices(
```
kks32: Keep this as Xd

kks32 (changes requested):
Shouldn't this be applied on all the nodes on a face? 

In which case, we have to use Eigen::MatrixXd

include/element.h
ezrayst: you mean Xi?

ezrayst: Then we have `int` instead of `unsigned`

cbgeo: Xi, so int

ezrayst: updated btw... seems to always fail in `codecov`

ezrayst:
@kks32 , any more comments on this one?

kks32:
@ezrayst 
Could you fix my previous comment?
> kks32 requested changes a day ago

```
Shouldn't this be applied on all the nodes on a face?

In which case, we have to use Eigen::MatrixXd
```

ezrayst:
I thought I solved it using MatrixXi. Or do you here refer to something else? I might misunderstand you

kks32:
@ezrayst There are 3 different quadrilateral elements and two hexahedrons, the face indices are not the same. Your solution only applies to linear elements.

ezrayst:
@kks32, I wonder whether we can find one normal for higher order elements. Assuming a general surface or curve, we potentially have a function of normal right?

Current assumption: higher order elements would still retain linear hexahedron/quadrilateral shape only with an extra nodes. Thus the face would still be defined by only corner nodes.

kks32:
@ezrayst Irrespective of how you find the normals in an element, this function should return all the nodes on a given face. This function has nothing to do with finding the normals (although it is primarily used to do that, but also you would need this to apply on all the nodes on a given face, in which case you are missing all the other non-corner nodes).

Finding a normal on a curved surface becomes tricky, and varies from one point to another, so I'm happy for you to make an assumption than elements remain linear to find the normal. We need to state this so it is not forgotten. Since we won't use highly distorted elements, it is not a bad assumption to make. 

I would return all the nodes on a given face, unless you have a different way of applying your velocities to a given face in a cell without knowing all the nodes? Won't you use this same function again for finding which nodes to apply the boundary conditions on a given face?

ezrayst:
Ok, understood your point. I did not consider the applying part in the later stages properly for higher order elements. I will change them accordingly then. 

Would you be OK with numbering following what we have, corner first, then inside elements?

kks32:
The numbering should be the same as what we have used before, it shouldn't change. Yes, corners first then the inside elements.

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,68 @@ inline Eigen::MatrixXi
...
+template <>
+inline Eigen::MatrixXi
+    mpm::HexahedronElement<3, 8>::face_indices(unsigned face_id) const {
+  
+  // Make return_indices
+  Eigen::Matrix<int, 4, 1> return_indices;
+  try {
+    // Check if face_id is within range
+    if (face_id < face_indices_hexahedron_.size()) {
+      return_indices = face_indices_hexahedron_.at(face_id).head(4);
```
kks32: Since you are chopping, you might as well define a const variable locally for each template specialisation and return the correct size. Create local variables of face indices for different element types (template specailsation).

include/hexahedron_element.h
```diff
@@ -210,7 +235,24 @@ class HexahedronElement : public Element<Tdim> {
...
   Eigen::MatrixXi inhedron_indices() const override;
 
+  //! Return indices of a face of an element
+  //! \param[in] face_id given id of the face
+  //! \retval indices Indices that make the face
+  Eigen::MatrixXi face_indices(unsigned face_id) const override;
+
  private:
+  //! Constants of face id and associate nodes
+  const std::map<unsigned, Eigen::Matrix<int, 8, 1>> face_indices_hexahedron_{
```
kks32: Remove this

kks32 (changes requested):
Define the face_indices locally within the function

include/hexahedron_element.tcc
ezrayst: But we need a library to store all data right? We need input `face_id` to know which set to return too. Does it mean we have to create the library over and over? I think storing it as private member would be better.

kks32: You are incurring more cost of creating another Eigen::Matrix by chopping, it would be more expensive than a function that has const variable returned each time. I don't understand what you mean by a library to store all data?

ezrayst: Say for 2D quadrilateral 4-node, we need to know that for face 0, the associated nodes are 0 and 1. Then for face 1, the associated nodes are 1 and 2. What I meant by library is the data that map face id to the associated nodes. 

The function has input `face_id`. If we make a local variable, we need to make the 'library' every time and only use one in the end. 

I think having a private member is better. Yes, truncation of matrix has cost, but it's only for one matrix instead of making 4 (for 2D) and 6 (for 3D) each time the function is being called.

kks32: It's a const function, so it is known at compile time, there is nothing more to do. It's more expensive to truncate and reconstruct a matrix than to just return a const. It's probably `inlined` this way of using `const` than your way of truncation. So, private member variable is not better and is more expensive.

kks32: If you are editing a private member variable, it is no longer a `const` function.

ezrayst: Alright. Let me do it

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,81 @@ inline Eigen::MatrixXi
   //clang-format on
   return indices;
 }
+
+
+//!            Behind: F4
```
kks32: You can remove this figure, as it appears in the header

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,81 @@ inline Eigen::MatrixXi
...
+      {0, Eigen::Matrix<int, 4, 1>(0, 1, 5, 4)},
+      {1, Eigen::Matrix<int, 4, 1>(5, 1, 2, 0)},
+      {2, Eigen::Matrix<int, 4, 1>(7, 6, 2, 3)},
+      {3, Eigen::Matrix<int, 4, 1>(0, 4, 7, 3)},
+      {4, Eigen::Matrix<int, 4, 1>(1, 0, 3, 2)},
+      {5, Eigen::Matrix<int, 4, 1>(4, 5, 6, 7)}};
+
+  try {
+    // Check if face_id is within range
+    if (face_id < face_indices_hexahedron.size()) {
```
kks32: We don't need to check, map `at` operation should handle this

include/quadrilateral_element.tcc
```diff
@@ -455,3 +455,85 @@ inline Eigen::MatrixXi
   return indices;
 }
   
+//!          F2
```
kks32: Remove the figure here

tests/hexahedron_element_test.cc
```diff
@@ -912,6 +912,30 @@ TEST_CASE("Hexahedron elements are checked", "[hex][element][3D]") {
...
+                 7, 6, 2, 3, 
+                 0, 4, 7, 3, 
+                 1, 0, 3, 2, 
+                 4, 5, 6, 7;
+      // clang-format on      
+
+      // Check for all face indices
+      for (unsigned i = 0; i < indices.rows(); ++i) {
+        auto check_indices =
+            hex->face_indices(i);
```
kks32: Remove extra line break

tests/hexahedron_element_test.cc
```diff
@@ -912,6 +912,30 @@ TEST_CASE("Hexahedron elements are checked", "[hex][element][3D]") {
...
+                 5, 1, 2, 0, 
+                 7, 6, 2, 3, 
+                 0, 4, 7, 3, 
+                 1, 0, 3, 2, 
+                 4, 5, 6, 7;
+      // clang-format on      
+
+      // Check for all face indices
+      for (unsigned i = 0; i < indices.rows(); ++i) {
+        auto check_indices =
```
kks32: const

kks32 (changes requested):

include/hexahedron_element.tcc
ezrayst: Yes! Are you OK with using `map` instead of `Eigen::Matrix`? I think I like `map` better.

kks32: It has always been a map!

include/hexahedron_element.h
```diff
@@ -80,6 +81,30 @@ namespace mpm {
...
+//!       |  0_ _ _ _ _|_ 0
+//!       | / 0        | / 1
+//!       |/     F0    |/
+//!       0_ _ _ _ _ _ 0
+//!     4               5
+//!         Front: F5
+//!
+//!
+//! Bottom face: F0, Right face: F1, Top face: F2,
+//! Left face: F3, Behind face: F4, Front face: F5
```
kks32: Instead of behind, use `rear` or `back`

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,50 @@ inline Eigen::MatrixXi
...
   return indices;
 }
+
+//! Return indices of a face of the element
+//! 8-noded hexahedron
+template <>
+inline Eigen::MatrixXi
+    mpm::HexahedronElement<3, 8>::face_indices(unsigned face_id) const {
   
+  //! Constants of face id and associate nodes
```
kks32: `//! face id and its associated nodal indices`

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,50 @@ inline Eigen::MatrixXi
...
+}
+
+//! Return indices of a face of the element
+//! 20-noded hexahedron
+template <>
+inline Eigen::MatrixXi
+    mpm::HexahedronElement<3, 20>::face_indices(unsigned face_id) const {
+  
+  //! Constants of face id and associate nodes
+  const std::map<unsigned, Eigen::Matrix<int, 8, 1>> face_indices_hexahedron{
```
kks32: Could you clang-format off this please and align them?

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,50 @@ inline Eigen::MatrixXi
...
+  }
+}
+
+//! Return indices of a face of the element
+//! 20-noded hexahedron
+template <>
+inline Eigen::MatrixXi
+    mpm::HexahedronElement<3, 20>::face_indices(unsigned face_id) const {
+  
+  //! Constants of face id and associate nodes
```
kks32: `//! face id and its associated nodal indices`

include/quadrilateral_element.h
```diff
@@ -2,6 +2,7 @@
 #define MPM_QUADRILATERAL_ELEMENT_H_
 
 #include <exception>
+#include <map>
```
kks32: include map in element

include/quadrilateral_element.h
```diff
@@ -2,6 +2,7 @@
 #define MPM_QUADRILATERAL_ELEMENT_H_
 
 #include <exception>
```
kks32: remove / move to element.h

include/hexahedron_element.h
```diff
@@ -2,6 +2,7 @@
 #define MPM_HEXAHEDRON_ELEMENT_H_
 
 #include <exception>
```
kks32: remove and move to element.h and map as well

include/quadrilateral_element.h
```diff
@@ -2,6 +2,7 @@
...
+#include <map>
 
 #include <Eigen/Dense>
```
kks32: Remove this as well

include/quadrilateral_element.tcc
```diff
@@ -455,3 +455,59 @@ inline Eigen::MatrixXi
   return indices;
 }
   
+//! Return indices of a face of the element
+//! 4-noded quadrilateral
+template <>
+inline Eigen::MatrixXi
+    mpm::QuadrilateralElement<2, 4>::face_indices(unsigned face_id) const {
+  
+  //! Constants of face id and associate nodes
```
kks32: `//! face id and its associated nodal indices`

include/quadrilateral_element.tcc
```diff
@@ -455,3 +455,59 @@ inline Eigen::MatrixXi
...
+  }
+}
+
+//! Return indices of a face of the element
+//! 8-noded quadrilateral
+template <>
+inline Eigen::MatrixXi
+    mpm::QuadrilateralElement<2, 8>::face_indices(unsigned face_id) const {
+  
+  //! Constants of face id and associate nodes
```
kks32: `//! face ids and its associated nodal indices`

kks32: Please note, in other places I have said `id` instead of ids, please use face ids in all places

kks32 (changes requested):
Just minor changes

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,48 @@ inline Eigen::MatrixXi
...
+  //! Face ids and its associated nodal indices
+  // clang-format off
+  const std::map<unsigned, Eigen::Matrix<int, 8, 1>> face_indices_hexahedron{
+      {0, (Eigen::Matrix<int, 8, 1>() << 0, 1, 5, 4,  8, 12, 16, 10).finished()},
+      {1, (Eigen::Matrix<int, 8, 1>() << 5, 1, 2, 0, 12, 11, 14, 18).finished()},
+      {2, (Eigen::Matrix<int, 8, 1>() << 7, 6, 2, 3, 19, 14, 13, 15).finished()},
+      {3, (Eigen::Matrix<int, 8, 1>() << 0, 4, 7, 3, 10, 17, 15,  9).finished()},
+      {4, (Eigen::Matrix<int, 8, 1>() << 1, 0, 3, 2,  8,  9, 13, 11).finished()},
+      {5, (Eigen::Matrix<int, 8, 1>() << 4, 5, 6, 7, 16, 18, 19, 17).finished()}};
+  //clang-format on
```
kks32: There should be a space between `//` and `clang-format on` otherwise it won't work

include/quadrilateral_element.h
```diff
@@ -1,8 +1,6 @@
 #ifndef MPM_QUADRILATERAL_ELEMENT_H_
 #define MPM_QUADRILATERAL_ELEMENT_H_
 
-#include <exception>
-
 #include <Eigen/Dense>
```
kks32: You can remove this

include/hexahedron_element.h
```diff
@@ -1,8 +1,6 @@
 #ifndef MPM_HEXAHEDRON_ELEMENT_H_
 #define MPM_HEXAHEDRON_ELEMENT_H_
 
-#include <exception>
-
 #include <Eigen/Dense>
```
kks32: You can remove  this

kks32 (changes requested):

include/hexahedron_element.tcc
```diff
@@ -563,4 +563,48 @@ inline Eigen::MatrixXi
...
+  try {
+    return face_indices_hexahedron.at(face_id);
+  } catch (std::exception& exception) {
```
kks32: I wouldn't do a try block here, because if you catch  an exception here, you are not returning anything to hope for it to continue, which will create more problems than handling this exception

kks32 (changes requested):
Remove all exception handlings in face_indices

include/hexahedron_element.tcc
ezrayst: I agree. Basically the analysis stops right? But do you want to throw something to indicate it's about this issue?

kks32 (approved):

PR #231, opened 2018-08-22
Make geometry class for coordinate system rotation matrix

ezrayst:
- Make geometry class to calculate any geometrical mathematics
- Add function to provide rotation matrix
- Add test functions

### DIFF
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -19,10 +19,10 @@ class Geometry {
   //! Constructor
   Geometry(){};
 
-  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! Return inverse rotation matrix for orthogonal axis coordinate system
   //! \param[in] angles Rotation angles depending on dimension
   //! \retval inverse of Euler rotation matrix R
-  Eigen::Matrix<double, Tdim, Tdim> compute_inverse_rotation_matrix(
+  Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles);
 
  private:
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -5,14 +5,13 @@
 #include <memory>
 
 #include "Eigen/Dense"
-#include "Eigen/LU"
 
 #include "logger.h"
 
 namespace mpm {
 
 //! Geometry class
-//! \brief Base class that computes geometry mathematics
+//! \brief Base class that performs geometric manipulations / operations
 //! \tparam Tdim Dimension
 template <unsigned Tdim>
 class Geometry {
@@ -21,12 +20,12 @@ class Geometry {
   Geometry(){};
 
   //! Compute inverse of rotation matrix for orthogonal axis coordinate system
-  //! \param[in] agnles Rotation angles depending on dimension
+  //! \param[in] angles Rotation angles depending on dimension
   //! \retval inverse of Euler rotation matrix R
-  Eigen::MatrixXd compute_inverse_rotation_matrix(
-      const Eigen::VectorXd& angles);
+  Eigen::Matrix<double, Tdim, Tdim> compute_inverse_rotation_matrix(
+      const Eigen::Matrix<double, Tdim, 1>& angles);
 
- protected:
+ private:
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
 
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -6,8 +6,6 @@
 
 #include "Eigen/Dense"
 
-#include "logger.h"
-
 namespace mpm {
 
 //! Geometry class
@@ -24,10 +22,6 @@ class Geometry {
   //! \retval inverse of Euler rotation matrix R
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles) const;
-
- private:
-  //! Logger
-  std::unique_ptr<spdlog::logger> console_;
 
 };  // Geometry class
 }  // namespace mpm
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -1,0 +1,38 @@
+#ifndef MPM_GEOMETRY_H_
+#define MPM_GEOMETRY_H_
+
+#include <limits>
+#include <memory>
+
+#include "Eigen/Dense"
+#include "Eigen/LU"
+
+#include "logger.h"
+
+namespace mpm {
+
+//! Geometry class
+//! \brief Base class that computes geometry mathematics
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class Geometry {
+ public:
+  //! Constructor
+  Geometry(){};
+
+  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! \param[in] agnles Rotation angles depending on dimension
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::MatrixXd compute_inverse_rotation_matrix(
+      const Eigen::VectorXd& angles);
+
+ protected:
+  //! Logger
+  std::unique_ptr<spdlog::logger> console_;
+
+};  // Geometry class
+}  // namespace mpm
+
+#include "geometry.tcc"
+
+#endif  // MPM_GEOMETRY_H_
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -17,13 +17,13 @@ template <unsigned Tdim>
 class Geometry {
  public:
   //! Constructor
-  Geometry(){};
+  Geometry() = default;
 
   //! Return inverse rotation matrix for orthogonal axis coordinate system
   //! \param[in] angles Rotation angles depending on dimension
   //! \retval inverse of Euler rotation matrix R
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
-      const Eigen::Matrix<double, Tdim, 1>& angles);
+      const Eigen::Matrix<double, Tdim, 1>& angles) const;
 
  private:
   //! Logger
--- a/include/geometry.tcc
+++ b/include/geometry.tcc
@@ -9,11 +9,10 @@ inline Eigen::Matrix<double, 2, 2> mpm::Geometry<2>::inverse_rotation_matrix(
   const double cos_beta = cos(angles(1));
   const double sin_beta = sin(angles(1));
 
-  Eigen::Matrix<double, 2, 2> rotation_matrix;
-
   // clang-format off
-  rotation_matrix << cos_alpha*cos_beta - sin_alpha*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_beta,
-                     sin_alpha*cos_beta + cos_alpha*sin_beta,  -sin_alpha*sin_beta + cos_alpha*cos_beta;                                 
+  const Eigen::Matrix<double, 2, 2> rotation_matrix = 
+    (Eigen::Matrix<double, 2, 2>() << cos_alpha*cos_beta - sin_alpha*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_beta,
+                                      sin_alpha*cos_beta + cos_alpha*sin_beta,  -sin_alpha*sin_beta + cos_alpha*cos_beta).finished();                                 
   // clang-format on            
 
   // Return inverted rotation matrix
@@ -33,12 +32,12 @@ inline Eigen::Matrix<double, 3, 3> mpm::Geometry<3>::inverse_rotation_matrix(con
   const double cos_gamma = cos(angles(2));
   const double sin_gamma = sin(angles(2));
 
-  Eigen::Matrix<double, 3, 3> rotation_matrix;
-
   // clang-format off
-  rotation_matrix << cos_alpha*cos_beta - sin_alpha*cos_gamma*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_gamma*cos_beta,   sin_gamma*sin_alpha,
-                     sin_alpha*cos_beta + cos_alpha*cos_gamma*sin_beta,  -sin_alpha*sin_beta + cos_alpha*cos_gamma*cos_beta,  -sin_gamma*cos_alpha,
-                     sin_gamma*sin_beta,                                  sin_gamma*cos_beta,                                  cos_gamma;
+  const Eigen::Matrix<double, 3, 3> rotation_matrix = 
+
+    (Eigen::Matrix<double, 3, 3>() << cos_alpha*cos_beta - sin_alpha*cos_gamma*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_gamma*cos_beta,   sin_gamma*sin_alpha,
+                                      sin_alpha*cos_beta + cos_alpha*cos_gamma*sin_beta,  -sin_alpha*sin_beta + cos_alpha*cos_gamma*cos_beta,  -sin_gamma*cos_alpha,
+                                      sin_gamma*sin_beta,                                  sin_gamma*cos_beta,                                  cos_gamma).finished();
   // clang-format on
 
   // Return inverted rotation matrix
--- a/include/geometry.tcc
+++ b/include/geometry.tcc
@@ -1,7 +1,7 @@
 //! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
 template <>
 inline Eigen::Matrix<double, 2, 2> mpm::Geometry<2>::inverse_rotation_matrix(
-    const Eigen::Matrix<double, 2, 1>& angles) {
+    const Eigen::Matrix<double, 2, 1>& angles) const {
 
   // Get cos and sin of angles
   const double cos_alpha = cos(angles(0));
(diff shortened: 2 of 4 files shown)


include/geometry.h
```diff
@@ -0,0 +1,38 @@
...
+//! \brief Base class that computes geometry mathematics
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class Geometry {
+ public:
+  //! Constructor
+  Geometry(){};
+
+  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! \param[in] agnles Rotation angles depending on dimension
```
kks32: `angles` not `agnles`

include/geometry.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "Eigen/Dense"
+#include "Eigen/LU"
+
+#include "logger.h"
+
+namespace mpm {
+
+//! Geometry class
+//! \brief Base class that computes geometry mathematics
```
kks32: Base class that performs geometric manipulations / operations

include/geometry.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MPM_GEOMETRY_H_
+#define MPM_GEOMETRY_H_
+
+#include <limits>
+#include <memory>
+
+#include "Eigen/Dense"
+#include "Eigen/LU"
```
kks32: What's LU for?

include/geometry.h
```diff
@@ -0,0 +1,38 @@
...
+  //! Constructor
+  Geometry(){};
+
+  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! \param[in] agnles Rotation angles depending on dimension
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::MatrixXd compute_inverse_rotation_matrix(
+      const Eigen::VectorXd& angles);
+
+ protected:
```
kks32: Why protected?

include/geometry.tcc
```diff
@@ -0,0 +1,52 @@
...
+
+  Eigen::Matrix<double, 2, 2> rotation_matrix;
+
+  // clang-format off
+  rotation_matrix << cos_alpha*cos_beta - sin_alpha*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_beta,
+                     sin_alpha*cos_beta + cos_alpha*sin_beta,  -sin_alpha*sin_beta + cos_alpha*cos_beta;                                 
+  // clang-format on            
+
+  // Invert rotation matrix
+  Eigen::Matrix<double, 2, 2> inverse_rotation_matrix;
```
kks32: What's the reasoning behind this? Can you just do it in retrun value?

include/geometry.tcc
```diff
@@ -0,0 +1,52 @@
...
+inline Eigen::MatrixXd mpm::Geometry<2>::compute_inverse_rotation_matrix(
+    const Eigen::VectorXd& angles) {
+
+  // Get cos and sin of angles
+  const double cos_alpha = cos(angles(0));
+  const double sin_alpha = sin(angles(0));
+  const double cos_beta = cos(angles(1));
+  const double sin_beta = sin(angles(1));
+
+  Eigen::Matrix<double, 2, 2> rotation_matrix;
```
kks32: Construct it as a const matrix?

include/geometry.tcc
```diff
@@ -0,0 +1,52 @@
+//! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
+template <>
+inline Eigen::MatrixXd mpm::Geometry<2>::compute_inverse_rotation_matrix(
+    const Eigen::VectorXd& angles) {
```
kks32: Can you pass `Eigen::Matrix<double, Tdim, 1>` instead of VectorXd?

include/geometry.tcc
```diff
@@ -0,0 +1,52 @@
+//! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
+template <>
+inline Eigen::MatrixXd mpm::Geometry<2>::compute_inverse_rotation_matrix(
```
kks32: Set size of return matrix `Eigen::Matrix<double, Tdim, Tdim>`

include/geometry.tcc
```diff
@@ -0,0 +1,52 @@
...
+
+  // Get cos and sin of angles
+  const double cos_alpha = cos(angles(0));
+  const double sin_alpha = sin(angles(0));
+  const double cos_beta = cos(angles(1));
+  const double sin_beta = sin(angles(1));
+  const double cos_gamma = cos(angles(2));
+  const double sin_gamma = sin(angles(2));
+
+  Eigen::Matrix<double, 3, 3> rotation_matrix;
```
kks32: Can you make this const

tests/geometry_test.cc
```diff
@@ -0,0 +1,95 @@
...
+TEST_CASE("Geometry is checked for 2D case", "[geometry][2D]") {
+
+  // Dimension
+  const unsigned Dim = 2;
+  // Tolerance
+  const double Tolerance = 1.E-7;
+
+  SECTION("Check inverse rotation matrix") {
+
+    auto geometry = std::make_shared<mpm::Geometry<Dim>>();
```
kks32: Why make_shared?

tests/geometry_test.cc
```diff
@@ -0,0 +1,95 @@
...
+    // clang-format off
+    inverse_rotation_matrix <<  0.435595740399158,  0.789149130992431,  0.433012701892219,
+                               -0.659739608441171, -0.047367172745376,  0.75,
+                                0.612372435695794, -0.612372435695795,  0.5;
+    // clang-format on
+    auto check_inverse_rotation_matrix =
+        geometry->compute_inverse_rotation_matrix(angles);
+    REQUIRE(check_inverse_rotation_matrix.cols() == 3);
+    REQUIRE(check_inverse_rotation_matrix.rows() == 3);
+    REQUIRE(check_inverse_rotation_matrix(0, 0) ==
```
kks32: Please write a for loop!

tests/geometry_test.cc
```diff
@@ -0,0 +1,95 @@
...
+    Eigen::Matrix<double, 2, 2> inverse_rotation_matrix;
+    // clang-format off
+    inverse_rotation_matrix <<  0.258819045102521, 0.965925826289068,
+                               -0.965925826289068, 0.258819045102521;
+    // clang-format on
+    auto check_inverse_rotation_matrix =
+        geometry->compute_inverse_rotation_matrix(angles);
+    REQUIRE(check_inverse_rotation_matrix.cols() == 2);
+    REQUIRE(check_inverse_rotation_matrix.rows() == 2);
+    REQUIRE(check_inverse_rotation_matrix(0, 0) ==
```
kks32: Write a loop!

kks32 (changes requested):

include/geometry.tcc
ezrayst: Wanted to be clear that it is the `inverse` not the `rotation matrix`. But we can return value for performance better

include/geometry.h
```diff
@@ -0,0 +1,37 @@
...
+template <unsigned Tdim>
+class Geometry {
+ public:
+  //! Constructor
+  Geometry(){};
+
+  //! Compute inverse of rotation matrix for orthogonal axis coordinate system
+  //! \param[in] angles Rotation angles depending on dimension
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::Matrix<double, Tdim, Tdim> compute_inverse_rotation_matrix(
```
kks32: Rename to `inverse_rotation_matrix`

tests/geometry_test.cc
```diff
@@ -0,0 +1,81 @@
...
+    // clang-format off
+    angles << 45 * M_PI / 180,   // alpha
+              30 * M_PI / 180;   // beta
+
+    Eigen::Matrix<double, 2, 2> inverse_rotation_matrix;
+    // clang-format off
+    inverse_rotation_matrix <<  0.258819045102521, 0.965925826289068,
+                               -0.965925826289068, 0.258819045102521;
+    // clang-format on
+    auto check_inverse_rotation_matrix =
```
kks32, line 33: const

tests/geometry_test.cc
```diff
@@ -0,0 +1,81 @@
...
+  // Tolerance
+  const double Tolerance = 1.E-7;
+
+  SECTION("Check inverse rotation matrix") {
+
+    auto geometry = std::make_unique<mpm::Geometry<Dim>>();
+
+    Eigen::Matrix<double, 3, 1> angles;
+    // clang-format off
+    angles << 45 * M_PI / 180,   // alpha
```
kks32: Always use `.` after numbers to denote floating points

kks32 (changes requested):

include/geometry.tcc
```diff
@@ -0,0 +1,45 @@
...
+    const Eigen::Matrix<double, 2, 1>& angles) {
+
+  // Get cos and sin of angles
+  const double cos_alpha = cos(angles(0));
+  const double sin_alpha = sin(angles(0));
+  const double cos_beta = cos(angles(1));
+  const double sin_beta = sin(angles(1));
+
+  // clang-format off
+  const Eigen::Matrix<double, 2, 2> rotation_matrix =
```
kks32: I saw somewhere you can just use `()`, not sure if it works for a matrix, could you check? https://eigen.tuxfamily.org/dox/group__TutorialMatrixArithmetic.html

include/geometry.h
```diff
@@ -0,0 +1,37 @@
...
+template <unsigned Tdim>
+class Geometry {
+ public:
+  //! Constructor
+  Geometry(){};
+
+  //! Return inverse rotation matrix for orthogonal axis coordinate system
+  //! \param[in] angles Rotation angles depending on dimension
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
```
kks32, line 23: const the function

include/geometry.h
```diff
@@ -0,0 +1,37 @@
...
+namespace mpm {
+
+//! Geometry class
+//! \brief Base class that performs geometric manipulations / operations
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class Geometry {
+ public:
+  //! Constructor
+  Geometry(){};
```
kks32: Set to default

kks32 (changes requested):

include/geometry.tcc
ezrayst: It's the default constructor. https://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html#title7

kks32: No argument can only be a default constructor, sorry, what I meant was can we do something like 
`Eigen::Matrix<double, 2, 2> test (1.0, 2.0, 3.0, 4.0);`

ezrayst: I don't think it's the way we initialize `Eigen`. https://forum.kde.org/viewtopic.php?f=74&t=107441

ezrayst: The default constructor `()` should be filled with the size of row and cols, not with the elements of the matrix. Thus I feel it's OK to use `()`. Another option is to use this:

`= (Eigen::MatrixXd rotation_matrix(2, 2) << 1.0, 2.0, 3.0, 4.0).finished()`

If not, we can always not used `const` and used comma-initialization.

ezrayst:
Anyway, I found these functions within `Eigen`: https://eigen.tuxfamily.org/dox/group__TutorialGeometry.html

And even functions to rotate 3D: https://eigen.tuxfamily.org/dox/classEigen_1_1AngleAxis.html

Now should we continue with what we have or should we utilize it? (I kind of feel bad)

Anyway, they have affine transformation too...

kks32:
I'd use Eigen's rotation functions, if they do what we want instead of our own functions. 

With regards to Affine transformation, our code does pre-computation, which makes it faster than Eigen.

kks32:
Having looked at their implementation, Eigen's implementation and ours are exactly identical. I'd say ours would be a bit more flexible, because we don't call the 2D rotation `Rotation2D` which is useful to have a single function to specialise for different dimensions.

include/geometry.tcc
kks32: Yes, I know we can do `.finished()`, if you look at affine_trasform.cc, we use that. I know we can do `Eigen::Vector3d test(1.0, 2.0, 3.0)`, my question was is it only specific to vectors or does it extend to matrices?

include/geometry.h
```diff
@@ -0,0 +1,37 @@
+#ifndef MPM_GEOMETRY_H_
+#define MPM_GEOMETRY_H_
+
+#include <limits>
+#include <memory>
+
+#include "Eigen/Dense"
+
+#include "logger.h"
```
kks32: I'm not throwing, why do I need a logger?

include/geometry.h
```diff
@@ -0,0 +1,37 @@
...
+
+  //! Return inverse rotation matrix for orthogonal axis coordinate system
+  //! \param[in] angles Rotation angles depending on dimension
+  //! \retval inverse of Euler rotation matrix R
+  Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
+      const Eigen::Matrix<double, Tdim, 1>& angles) const;
+
+ private:
+  //! Logger
+  std::unique_ptr<spdlog::logger> console_;
```
kks32: Remove this

include/geometry.tcc
```diff
@@ -0,0 +1,45 @@
...
+  const double cos_alpha = cos(angles(0));
+  const double sin_alpha = sin(angles(0));
+  const double cos_beta = cos(angles(1));
+  const double sin_beta = sin(angles(1));
+  const double cos_gamma = cos(angles(2));
+  const double sin_gamma = sin(angles(2));
+
+  // clang-format off
+  const Eigen::Matrix<double, 3, 3> rotation_matrix = 
+
```
kks32: Remove empty line

include/geometry.tcc
```diff
@@ -0,0 +1,45 @@
+//! Compute inverse of 2d rotation matrix for orthogonal axis coordinate system
+template <>
+inline Eigen::Matrix<double, 2, 2> mpm::Geometry<2>::inverse_rotation_matrix(
+    const Eigen::Matrix<double, 2, 1>& angles) const {
+
+  // Get cos and sin of angles
+  const double cos_alpha = cos(angles(0));
```
kks32: It would be better if we precompute `cos_alpha * cos_beta` and other combinations instead of these?

kks32 (changes requested):

ezrayst:
I would say their euler angle is slightly different notation from ours. But well, I prefer our own implementation actually, for the same reason that we have one function to call, instead of having `if else`.

kks32:
@ezrayst Have you looked at their implementation?

include/geometry.h
ezrayst: Well not now, but we will use it in the future as we are adding more functions right? Or we always assume that we never `throw` here?

kks32:
There are 24 different ways Euler angles can be represented!

include/geometry.h
kks32: Add it when you use it

include/geometry.tcc
ezrayst: We have quite a few combinations. I prefer to precompute the element in the matrix separately so we end of with 9 (for 3D) and 4 (for 2D) combinations only, instead of potential 36 (for 3D) and 16 (for 2D).

I thought the most expensive part is computing the trigonometry function. Multiplication is quite cheap right?

tests/geometry_test.cc
```diff
@@ -0,0 +1,81 @@
...
+
+  SECTION("Check inverse rotation matrix") {
+
+    const auto geometry = std::make_unique<mpm::Geometry<Dim>>();
+
+    Eigen::Matrix<double, 2, 1> angles;
+    // clang-format off
+    angles << 45. * M_PI / 180.,   // alpha
+              30. * M_PI / 180.;   // beta
+
```
kks32: Turn on clang-format

kks32 (changes requested):

include/geometry.tcc
ezrayst: I tried that in matrix, I don't think it extends to matrices unfortunately. To my understanding, if we do `Eigen::Matrix<double, 3, 3> test(...)` the `(...)` is reserved to construct the size of the matrix instead of the value of elements of matrix.

include/geometry.tcc
kks32: For 2D it will be only 4 combinations!, so for me it doesn't make sense to store a temporary value if we can store a product of `cos_alpha * cos_beta`

kks32: What do you mean by precompute the element matrix? Where is the precomputation happening in this code?

kks32 (changes requested):

include/geometry.tcc
```diff
@@ -0,0 +1,45 @@
...
+  const double sin_alpha = sin(angles(0));
+  const double cos_beta = cos(angles(1));
+  const double sin_beta = sin(angles(1));
+  const double cos_gamma = cos(angles(2));
+  const double sin_gamma = sin(angles(2));
+
+  // clang-format off
+  const Eigen::Matrix<double, 3, 3> rotation_matrix = 
+
+    (Eigen::Matrix<double, 3, 3>() << cos_alpha*cos_beta - sin_alpha*cos_gamma*sin_beta,  -cos_alpha*sin_beta - sin_alpha*cos_gamma*cos_beta,   sin_gamma*sin_alpha,
```
kks32: Precompute the set of the same 4 components you would do in 2D, `cos_a * cos _b`, `sin_a * sin_b`, `cos_a * sin_b`, and `sin_a * cos_b`, keep cos and sin of `gamma` as is

kks32 (changes requested):

include/geometry.tcc
kks32: Interesting, considering `Eigen::VectorXd` is an alias for `Eigen::Matrix<double, XX, 1>`, so technically both should have the same function calls. Let's stick with `.finished()` like we use in other parts of the code.

kks32 (changes requested):

ezrayst:
I will take a look at their implementation tomorrow. I have briefly read the descriptions though. I will update you more on it.

kks32 (approved):

PR #235, opened 2018-08-23
Compute angle between two vectors

ezrayst:
- Add function in `geometry` to compute angle between two vectors
- Input is `VectorXd` so this is good for 3D and 2D
- Test cases are made ready

### DIFF
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -27,7 +27,7 @@ class Geometry {
   //! \param[in] vector_a First vector
   //! \param[in] vecotr_b Second vector
   //! \retval anti-clockwise angle from vector_a to vector_b
-  double angle_between_two_vectors(
+  const double angle_between_two_vectors(
       const Eigen::Matrix<double, Tdim, 1>& vector_a,
       const Eigen::Matrix<double, Tdim, 1>& vector_b);
 
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -27,7 +27,7 @@ class Geometry {
   //! \param[in] vector_a First vector
   //! \param[in] vecotr_b Second vector
   //! \retval anti-clockwise angle from vector_a to vector_b
-  const double angle_between_two_vectors(
+  const double angle_between_vectors(
       const Eigen::Matrix<double, Tdim, 1>& vector_a,
       const Eigen::Matrix<double, Tdim, 1>& vector_b);
 
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -23,6 +23,13 @@ class Geometry {
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles) const;
 
+  //! Return angle between two vectors in radians
+  //! \param[in] vector_a First vector
+  //! \param[in] vecotr_b Second vector
+  //! \retval anti-clockwise angle from vector_a to vector_b
+  double angle_between_two_vectors(const Eigen::VectorXd& vector_a,
+                                   const Eigen::VectorXd& vector_b);
+
 };  // Geometry class
 }  // namespace mpm
 
--- a/include/geometry.h
+++ b/include/geometry.h
@@ -27,8 +27,9 @@ class Geometry {
   //! \param[in] vector_a First vector
   //! \param[in] vecotr_b Second vector
   //! \retval anti-clockwise angle from vector_a to vector_b
-  double angle_between_two_vectors(const Eigen::VectorXd& vector_a,
-                                   const Eigen::VectorXd& vector_b);
+  double angle_between_two_vectors(
+      const Eigen::Matrix<double, Tdim, 1>& vector_a,
+      const Eigen::Matrix<double, Tdim, 1>& vector_b);
 
 };  // Geometry class
 }  // namespace mpm
--- a/include/geometry.tcc
+++ b/include/geometry.tcc
@@ -48,20 +48,10 @@ inline Eigen::Matrix<double, 3, 3> mpm::Geometry<3>::inverse_rotation_matrix(con
 }
 
 //! Return angle between two vectors in radians
-template <>
-inline const double mpm::Geometry<2>::angle_between_two_vectors(
-    const Eigen::Matrix<double, 2, 1>& vector_a,
-    const Eigen::Matrix<double, 2, 1>& vector_b) {
-
-  // return angle between vector a and b = arccos( a dot b / ||a|| ||b||)
-  return acos((vector_a.normalized()).dot((vector_b.normalized())));
-}
-
-//! Return angle between two vectors in radians
-template <>
-inline const double mpm::Geometry<3>::angle_between_two_vectors(
-    const Eigen::Matrix<double, 3, 1>& vector_a,
-    const Eigen::Matrix<double, 3, 1>& vector_b) {
+template <unsigned Tdim>
+inline const double mpm::Geometry<Tdim>::angle_between_vectors(
+    const Eigen::Matrix<double, Tdim, 1>& vector_a,
+    const Eigen::Matrix<double, Tdim, 1>& vector_b) {
 
   // return angle between vector a and b = arccos( a dot b / ||a|| ||b||)
   return acos((vector_a.normalized()).dot((vector_b.normalized())));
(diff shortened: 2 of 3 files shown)


include/geometry.h
```diff
@@ -23,6 +23,13 @@ class Geometry {
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles) const;
 
+  //! Return angle between two vectors in radians
+  //! \param[in] vector_a First vector
+  //! \param[in] vecotr_b Second vector
+  //! \retval anti-clockwise angle from vector_a to vector_b
+  double angle_between_two_vectors(const Eigen::VectorXd& vector_a,
```
kks32: Why can't it be `Vector<double, Tdim, 1>`?

include/geometry.tcc
```diff
@@ -45,4 +45,16 @@ inline Eigen::Matrix<double, 3, 3> mpm::Geometry<3>::inverse_rotation_matrix(con
...
+//! Return angle between two vectors in radians
+template <unsigned Tdim>
+double mpm::Geometry<Tdim>::angle_between_two_vectors(
+    const Eigen::VectorXd& vector_a, const Eigen::VectorXd& vector_b) {
+
+  // angle between vector a and b = arccos( a dot b / ||a|| ||b||)
+  const double angle =
+      acos((vector_a.normalized()).dot((vector_b.normalized())));
+
+  return angle;
```
kks32: Use a return value optimisation, and make this function inline

tests/geometry_test.cc
```diff
@@ -17,6 +17,7 @@ TEST_CASE("Geometry is checked for 2D case", "[geometry][2D]") {
 
   SECTION("Check inverse rotation matrix") {
```
kks32: Remove empty line

tests/geometry_test.cc
```diff
@@ -41,6 +42,21 @@ TEST_CASE("Geometry is checked for 2D case", "[geometry][2D]") {
...
+    // Make geometry
+    const auto geometry = std::make_unique<mpm::Geometry<Dim>>();
+
+    Eigen::Matrix<double, 2, 1> vector_a;
+    vector_a << 3, 0;
+    Eigen::Matrix<double, 2, 1> vector_b;
+    vector_b << -2, 2;
+    const double angle = 2.356194490192345;
+
+    auto check_angle = geometry->angle_between_two_vectors(vector_a, vector_b);
```
kks32: Remove unnecessary `check_angle`

kks32 (changes requested):

include/geometry.h
ezrayst: We can do that too

include/geometry.tcc
```diff
@@ -45,4 +45,14 @@ inline Eigen::Matrix<double, 3, 3> mpm::Geometry<3>::inverse_rotation_matrix(con
 
   // Return inverted rotation matrix
   return rotation_matrix.inverse();
-}
\ No newline at end of file
+}
+
+//! Return angle between two vectors in radians
+template <unsigned Tdim>
+double mpm::Geometry<Tdim>::angle_between_two_vectors(
```
kks32: Please make this a const function and inline

tests/geometry_test.cc
```diff
@@ -79,4 +94,19 @@ TEST_CASE("Geometry is checked for 3D case", "[geometry][3D]") {
       }
     }
   }
+
+  SECTION("Check angle between two vectors") {
+    // Make geometry
+    const auto geometry = std::make_unique<mpm::Geometry<Dim>>();
+
+    Eigen::Matrix<double, 3, 1> vector_a;
+    vector_a << 3, 0, 0;
```
kks32: Could you use non-zero third dimension? Otherwise it is the same as 2D

kks32 (changes requested):
Why is the final commit called ` optimize initialization`? What part of initialisation is being optimised?

I don't understand the reasoning behind this function? What if the 2 vectors are not on the same plane?

tests/geometry_test.cc
ezrayst: Oh, yes!

ezrayst:
We need to get the Euler angles (our definition) to define new coordinate system. For that, we need to compute 2 angles for 2D and 3 angles for 3D.

For instance, the definition of `gamma` angle would be between vector [0 0 1] to the normal to the face. Once obtained, we can rotate coordinate system.

kks32:
But this function returns only one angle, you mentioned you need to compute 2 angles for 2D and 3 for 3D?

ezrayst:
As the title suggests, this function only computes an angle between two vectors. 2 angles for 2D would be done in other functions (in `cell`) that call this function.

kks32:
I don't understand which angle you are computing if the two vectors are not on the same plane. In other words, where do you intend to use this function?

ezrayst:
I can always get an angle between any two vectors. The angle from vector a to vector b and the angle vice versa should add up to 360. It is in geometry, so it should be general mathematics. Simply calculating an angle between two random vectors.

I intend to call this function to calculate angle `alpha`, `beta` and `gamma`. Once we get the normal to the face (the new `z` axis) and specify the new `x` axis, thus automatically get the new `y` axis (it has to be orthogonal to both `x` and `z`), then we can call this function to get all angles. Then we feed them to rotation matrix to transform velocity.

kks32:
This basically computes the angle between two vectors `on a given plane`

include/geometry.h
```diff
@@ -23,6 +23,14 @@ class Geometry {
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles) const;
 
+  //! Return angle between two vectors in radians
+  //! \param[in] vector_a First vector
+  //! \param[in] vecotr_b Second vector
+  //! \retval anti-clockwise angle from vector_a to vector_b
+  const double angle_between_two_vectors(
```
kks32: Can we rename to `angle_between_vectors`

include/geometry.h
```diff
@@ -23,6 +23,14 @@ class Geometry {
   Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix(
       const Eigen::Matrix<double, Tdim, 1>& angles) const;
 
+  //! Return angle between two vectors in radians
```
kks32, line 26: Here say that on the shared plane to make it more obvious?

include/geometry.h
```diff
@@ -23,6 +23,14 @@ class Geometry {
...
+  //! \param[in] vector_a First vector
+  //! \param[in] vecotr_b Second vector
+  //! \retval anti-clockwise angle from vector_a to vector_b
```
kks32, line 29: There is no more `\retval`

include/geometry.tcc
```diff
@@ -45,4 +45,24 @@ inline Eigen::Matrix<double, 3, 3> mpm::Geometry<3>::inverse_rotation_matrix(con
 
   // Return inverted rotation matrix
   return rotation_matrix.inverse();
+}
+
+//! Return angle between two vectors in radians
+template <>
+inline const double mpm::Geometry<2>::angle_between_two_vectors(
```
kks32: Why do we specialise when we have the same implementation?

tests/geometry_test.cc
```diff
@@ -79,4 +94,19 @@ TEST_CASE("Geometry is checked for 3D case", "[geometry][3D]") {
       }
     }
   }
+
+  SECTION("Check angle between two vectors") {
```
kks32, line 98: add `on the shared plane`

kks32 (changes requested):

include/geometry.h
ezrayst: what do you mean there is no more? We don't use it or the function does not return value?

include/geometry.h
ezrayst: I think it's good. We can't have angle between more than 2 vectors anyway

include/geometry.h
ezrayst: I am fine in adding it, but seems unnecessary. If not on the shared plane, the angle is either 0 or 360 right?

kks32 (approved):

PR #238, opened 2018-08-27
:hammer: Fix geometry documentations

kks32:


cbgeo (approved):

ezrayst:
Can you explain about `\retval` and why we seem not to use them?

kks32:
There is no variable being returned here, so we can't use `\retval`, since we are doing return value optimisation. The first argument in `\retval` is the variable name, we don't have a variable to say what it is.

ezrayst:
Understood!

PR #242, opened 2018-08-29
Assign velocity boundary to cell and compute normal

ezrayst:
- Make functions to assign velocity boundary
- Compute and store normals
- Make test cases

Note: need to merge with `face_id` first


include/cell.h
```diff
@@ -222,6 +224,29 @@ class Cell {
...
+
+  //! Apply velocity constraints
+  void apply_cell_velocity_constraints();
+
+  //! Compute normal vector
+  void compute_normals();
+
+  //! Return number of normal vectors
+  //! \retval number of normal vectors
+  unsigned nnormal() const { return face_normals_.size(); }
```
kks32: Shouldn't this be equal to the number of faces a cell has?

include/cell.h
```diff
@@ -222,6 +224,29 @@ class Cell {
...
+
+  //! Return unit normal vector, positive pointing outside of the element
+  //! \param[in] face_id of constraint
+  //! \retval unit normal vector
+  Eigen::VectorXd normal(unsigned face_id);
```
kks32: I need to check where this is used?

include/cell.h
```diff
@@ -249,8 +274,19 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
```
kks32: comments don't match the description

include/cell.h
```diff
@@ -249,8 +274,19 @@ class Cell {
...
+  // std::vector<std::tuple<unsigned, unsigned, double>> velocity_constraints_;
+  std::map<unsigned, std::vector<std::tuple<unsigned, double>>>
```
kks32: You couldn't use a pair?

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
 
   return acceleration;
 }
+
+//! Assign velocity constraint
+//! Constrain directions can take values between 0 and Dim-1
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::assign_cell_velocity_constraint(unsigned face_id,
```
kks32: It's in the cell just call it `assign_velocity_constraint`

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+                                                      unsigned dir,
+                                                      double velocity) {
+  bool status = true;
+  try {
+    //! Constrain directions can take values between 0 and Dim * Nphases - 1
+    if (dir >= 0 && dir < Tdim) {
+      if (velocity_constraints_.find(face_id) == velocity_constraints_.end())
```
kks32: It is stored as a map, you don't need to do use find.

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+        this->velocity_constraints_.insert(
+            std::make_pair<unsigned, std::vector<std::tuple<unsigned, double>>>(
+                static_cast<unsigned>(face_id),
+                std::vector<std::tuple<unsigned, double>>()));
+
+      this->velocity_constraints_.at(face_id).emplace_back(
```
kks32: What is this doing compared to line #966?

kks32 (changes requested):

include/cell.h
ezrayst: Right, will use pair. I was in a rush just now, and stick with what I had before :(

include/cell.h
ezrayst: No, better to only calculate normal of the face with boundary conditions

include/cell.tcc
ezrayst: Yes, but I want it to return `end()` if it is not found. So what this line and the next do are to check whether we have `face_id` for the `first`. If we do, simply `emplace_back` but if we don't, make a pair then emplace.

include/cell.h
ezrayst: We need to call the normal later to calculate euler angles before computing rotation matrix

include/cell.h
```diff
@@ -10,6 +11,7 @@
 
 #include "affine_transform.h"
 #include "element.h"
+// #include "geometry.h"
```
kks32: Remove if not used

include/cell.h
kks32: Why? It's going to be calculated once at the start, I don't see a performance issue. This way we store all normals and can use it for example to apply traction.

include/cell.h
```diff
@@ -249,8 +274,18 @@ class Cell {
...
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
+  std::map<unsigned, std::vector<std::pair<unsigned, double>>>
+      velocity_constraints_;
+
+  //! Normal of face with velocity constraints
+  std::vector<Eigen::VectorXd> face_normals_;
+
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
+
```
kks32: Remove space

include/cell.tcc
kks32: I see, You can remove this check and just do:
```
      this->velocity_constraints_[face_id].emplace_back(
          std::make_pair<unsigned, double>(static_cast<unsigned>(dir),
                                           static_cast<double>(velocity)));
```

We can check for duplicates when we read the input file.

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+//! Assign velocity constraint
+//! Constrain directions can take values between 0 and Dim-1
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
+                                                 double velocity) {
+  bool status = true;
+  try {
+    //! Constrain directions can take values between 0 and Dim * Nphases - 1
+    if (dir >= 0 && dir < Tdim) {
```
kks32: Maybe we don't check for Nphases, but accept this for (Dim * Nphases) direction possibilities?

kks32: Check can happen at the node level where we have that information?

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    } else
+      throw std::runtime_error("Constraint direction is out of bounds");
+  } catch (std::exception& exception) {
+    console_->error("{} #{}: {}\n", __FILE__, __LINE__, exception.what());
+    status = false;
+  }
+  return status;
+}
+
+//! Apply velocity constraints
```
kks32: Remove this until we add apply

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+    // Compute normal and make unit vector
+    // The normal vector n to vector a is defined such that the dot product
+    // between a and n is always 0 In 2D, n(0) = -a(1), n(1) = a(0) Note that
+    // the reverse does not work to produce normal that is positive pointing out
+    // of the element
+    Eigen::Matrix<double, 2, 1> normal_vector;
+    normal_vector(0) = -a(1);
+    normal_vector(1) = a(0);
+    normal_vector /= normal_vector.norm();
```
kks32: Can we use Eigen's `.normalized()` here?

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    // Store to private variable
+    face_normals_.emplace_back(normal_vector);
+  }
+}
+
+//! Compute normal 3d
+template <>
+inline void mpm::Cell<3>::compute_normals() {
+
+  for (const auto& velocity_constraint : this->velocity_constraints_) {
```
kks32: I would like normals to be computed for all faces than just on velocity constraints.

include/cell.h
```diff
@@ -249,8 +274,18 @@ class Cell {
...
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
+  std::map<unsigned, std::vector<std::pair<unsigned, double>>>
+      velocity_constraints_;
+
+  //! Normal of face with velocity constraints
+  std::vector<Eigen::VectorXd> face_normals_;
```
kks32: I'd store this as a map

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+}
+
+//! Return unit normal vector
+template <unsigned Tdim>
+Eigen::VectorXd mpm::Cell<Tdim>::normal(unsigned face_id) {
+
+  Eigen::Matrix<double, Tdim, 1> normal_vector;
+  normal_vector.setZero();
+
+  try {
```
kks32: When we have it as a map, we don't do this check

include/cell.tcc
```diff
@@ -951,3 +951,118 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+    // Compute normal and make unit vector
+    // normal = a x b
+    // Note that definition of a and b are such that normal is always out of
+    // page
+    Eigen::Matrix<double, 3, 1> normal_vector = a.cross(b);
+    normal_vector /= normal_vector.norm();
+
+    // Store to private variable
+    face_normals_.emplace_back(normal_vector);
```
kks32: Depending on the order of velocity constraint, this may not be in the order you expect it to be.

include/mesh.h
```diff
@@ -148,12 +149,19 @@ class Mesh {
...
   //! \param[in] velocity_constraints Constraint at node, dir, and velocity
   bool assign_velocity_constraints(
       const std::vector<std::tuple<mpm::Index, unsigned, double>>&
           velocity_constraints);
 
+  //! Assign velocity constraints to cells
+  //! \param[in] velocity_constraints Constraint at cell id, phase id, dir, and
+  //! velocity
+  bool assign_cell_velocity_constraints(
+      const std::vector<std::tuple<mpm::Index, mpm::Index, unsigned, double>>&
```
kks32: Why is phase id `mpm::Index`?

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
 }
 
+//! Assign velocity constraints to cells
+template <unsigned Tdim>
+bool mpm::Mesh<Tdim>::assign_cell_velocity_constraints(
+    const std::vector<std::tuple<mpm::Index, mpm::Index, unsigned, double>>&
+        velocity_constraints) {
+  bool status = false;
+  try {
+    if (nodes_.size()) {
```
kks32: I think we don't need to check this, if you are handling cells, it should have all the nodes. This check doesn't assume that the nodes are linked to the cells.

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
+      for (const auto& velocity_constraint : velocity_constraints) {
+        // Cell id
+        mpm::Index cell_id = std::get<0>(velocity_constraint);
+        // Face id
+        mpm::Index face_id = std::get<1>(velocity_constraint);
```
kks32: `face_id` is unsigned

tests/cell_test.cc
```diff
@@ -1,3 +1,4 @@
+#include <cmath>
 #include <limits>
```
kks32: Add an extra line to separate cincludes from regular c++ includes

tests/cell_test.cc
```diff
@@ -723,6 +724,74 @@ TEST_CASE("Cell is checked for 2D case", "[cell][2D]") {
                 Approx(interpolated_acceleration(i)).epsilon(Tolerance));
     }
   }
+
+  SECTION("Check velocity boundary conditions") {
+
```
kks32: Remove space

tests/cell_test.cc
```diff
@@ -723,6 +724,74 @@ TEST_CASE("Cell is checked for 2D case", "[cell][2D]") {
...
+    SECTION("Check normal vector") {
+
```
kks32: Remove space

kks32 (changes requested):

include/cell.h
kks32: Should this be public?

include/cell.tcc
ezrayst: will do so!

include/cell.tcc
ezrayst: I see, that would be quicker. Can we do it later though and leave it for now?

kks32: Sure, please place a  `// TODO` comment here, and add an issue to come back later, when this PR has been merged.

include/cell.h
ezrayst: Good idea! If we compute all normals this make sense

include/cell.h
```diff
@@ -222,6 +224,31 @@ class Cell {
...
+
+  //! Apply velocity constraints
+  void apply_velocity_constraints();
+
+  //! Compute normal vector
+  void compute_normals();
+
+  //! Return number of normal vectors
+  //! \retval number of normal vectors
+  unsigned nnormal() const { return face_normals_.size(); }
```
kks32: Do we still need this function?

include/cell.tcc
kks32: Could you fix this in this PR? Any particular reason why you want to do it later?

include/cell.tcc
```diff
@@ -949,3 +949,101 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+      throw std::runtime_error("Constraint direction is out of bounds");
+  } catch (std::exception& exception) {
+    console_->error("{} #{}: {}\n", __FILE__, __LINE__, exception.what());
+    status = false;
+  }
+  return status;
+}
+
+//! Apply velocity constraints
+template <unsigned Tdim>
```
kks32: Remove this  if this is not going to be part of the PR

include/cell.tcc
```diff
@@ -949,3 +949,101 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
+
+  // TODO
+}
+
+//! Compute all face normal 3d
+template <>
+inline void mpm::Cell<2>::compute_normals() {
+
+  for (unsigned face_id = 0; face_id < 4; ++face_id) {
```
kks32: could you `const` a variable to 4, which says why it is 4? Why can't it be 3? if it is a triangle?

include/cell.tcc
```diff
@@ -949,3 +949,101 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    face_normals_.insert(std::make_pair<unsigned, Eigen::VectorXd>(
+        static_cast<unsigned>(face_id), normal_vector));
+  }
+}
+
+//! Compute all face normal 3d
+template <>
+inline void mpm::Cell<3>::compute_normals() {
+
+  for (unsigned face_id = 0; face_id < 6; ++face_id) {
```
kks32: Should be generic for any 3D element

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
+
+        // Apply constraint
+        status = map_cells_[cell_id]->assign_velocity_constraint(face_id, dir,
+                                                                 velocity);
+
+        if (!status)
+          throw std::runtime_error(
+              "Cell or face or velocity constraint is invalid");
+      }
+    } else {
```
kks32: Check and throw at the start, no need to nest the for loop inside a big if.

include/mesh.tcc
```diff
@@ -347,6 +349,46 @@ bool mpm::Mesh<Tdim>::assign_velocity_constraints(
...
+    const std::vector<std::tuple<mpm::Index, unsigned, unsigned, double>>&
+        velocity_constraints) {
+  bool status = false;
+  try {
+    if (nodes_.size()) {
+      for (const auto& velocity_constraint : velocity_constraints) {
+        // Cell id
+        mpm::Index cell_id = std::get<0>(velocity_constraint);
+        // Face id
+        unsigned face_id = std::get<1>(velocity_constraint);
```
kks32: could we const these

kks32 (changes requested):

include/cell.tcc
ezrayst: We will never use triangle right? It's not that generic for 3D and 2D...

ezrayst:
Should we pass `Tnphases` as a template argument as well for `cell`?

include/cell.h
ezrayst: I rather still have it to get the number of normals rather than specify 4 or 6 everytime

include/cell.tcc
ezrayst: I plan to, but if you are ok with the rest, we can make it not part of the PR

include/mesh.tcc
ezrayst: Well there is a potential to throw within the for loop as well, thus we have the loop inside big if

include/cell.h
kks32: It's used only in the cell, you can call `face_normals_.size()`, without having to define a public function?

include/cell.h
```diff
@@ -222,6 +224,31 @@ class Cell {
...
+
+  //! Return unit normal vector, positive pointing outside of the element
+  //! \param[in] face_id of constraint
+  //! \retval unit normal vector
+  Eigen::VectorXd normal(unsigned face_id) const {
```
kks32: You don't need this function, if only `cell` is going to use it. It would be a lot clearer to just access the private variable.

include/cell.h
```diff
@@ -249,6 +276,16 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
+  std::map<unsigned, std::vector<std::pair<unsigned, double>>>
+      velocity_constraints_;
+
+  //! Normal of face with velocity constraints
```
kks32: Remove `with velocity constraints`

include/cell.tcc
kks32: Leave it if you are going to implement it.

include/cell.tcc
```diff
@@ -949,3 +949,107 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+  // TODO
+}
+
+//! Compute all face normal 2d
+template <>
+inline void mpm::Cell<2>::compute_normals() {
+
+  //! Set number of faces
+  const unsigned nfaces = 4;
```
kks32: This should be from the `element` class, cell should never define the number of faces.

include/mesh.h
```diff
@@ -184,6 +192,8 @@ class Mesh {
  private:
   // Locate a particle in mesh cells
   bool locate_particle_cells(std::shared_ptr<mpm::ParticleBase<Tdim>> particle);
+
+ protected:
```
kks32: Why `protected`?

include/mesh.tcc
kks32: My question is why can't you just check `if (!nodes_.size())` and throw, why should the for loop be inside the `if` block?

kks32 (changes requested):

include/cell.tcc
kks32: We may or may not use triangles, but that shouldn't affect how we define this. Get number of faces from the `element` class.

kks32:
Please do not pass `Tnphases`, because `cell` doesn't need to know how many phases it is handling. That's the problem with the `node` class. You can merely check if velocity constraints are greater than zero. It will fail, if you try to apply at the node. So no need to do `Tnphases`

include/mesh.tcc
ezrayst: Yes but line 377 has another throw. It's either nest it or to have 2 `throw`, one for the `nodes_size()` and the other for `status` from `assign_velocity_constraint`

include/cell.tcc
ezrayst: Will make them in element class then!

include/cell.tcc
ezrayst: Thanks

include/cell.tcc
ezrayst: Yes, will do it in this PR too!

include/cell.h
ezrayst: OK and done!

include/cell.h
ezrayst: Agree, and we have `element/nfaces()`, thus this is deprecated

ezrayst: Actually, I would still use it to check the number of faces. Or is it sufficient to know that whatever the element gives is right?

include/mesh.tcc
kks32: Could you please explain why nesting a `for` loop inside an `if` helps, when both are inside a `try` block? If you move the `for` loop out of `if` that should help in branch prediction. What is the potential benefit of keeping this structure?

include/cell.h
kks32: `Element` should give it right, you have no other way of checking it in the cell.

include/cell.h
```diff
@@ -249,6 +276,16 @@ class Cell {
 
   //! Shape function
   std::shared_ptr<const Element<Tdim>> element_{nullptr};
+
+  //! Velocity constraints
+  //! first-> face_id, second->pair of direction [0/1/2] and velocity value
```
kks32: `//! key:face_id, value: pair of direction [0/1/2] and velocity value`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+}
+
+//! Apply velocity constraints
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
+
+  // TODO
+}
+
+//! Compute all face normal 2d
```
kks32: normal`s`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+template <>
+inline void mpm::Cell<2>::compute_normals() {
+
+  //! Set number of faces from element
+  for (unsigned face_id = 0; face_id < element_->nfaces(); ++face_id) {
+    // Get the nodes of the face
+    Eigen::VectorXi indices = element_->face_indices(face_id);
```
kks32: `const`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    Eigen::Matrix<double, 2, 1> a = (this->nodes_[indices(0)])->coordinates() -
+                                    (this->nodes_[indices(1)])->coordinates();
+
+    // Compute normal and make unit vector
+    // The normal vector n to vector a is defined such that the dot product
+    // between a and n is always 0 In 2D, n(0) = -a(1), n(1) = a(0) Note that
+    // the reverse does not work to produce normal that is positive pointing out
+    // of the element
+    Eigen::Matrix<double, 2, 1> normal_vector;
+    normal_vector(0) = -a(1);
```
kks32: `normal_vector << -a(1), a(0);`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    normal_vector(1) = a(0);
+    normal_vector = normal_vector.normalized();
+
+    // Store to private variable
```
kks32: Remove comment

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    face_normals_.insert(std::make_pair<unsigned, Eigen::VectorXd>(
+        static_cast<unsigned>(face_id), normal_vector));
+  }
+}
+
+//! Compute all face normal 3d
```
kks32: `normals`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+template <>
+inline void mpm::Cell<3>::compute_normals() {
+
+  //! Set number of faces from element
+  for (unsigned face_id = 0; face_id < element_->nfaces(); ++face_id) {
+    // Get the nodes of the face
+    Eigen::VectorXi indices = element_->face_indices(face_id);
```
kks32: `const`

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+                                    (this->nodes_[indices(0)])->coordinates();
+
+    // Compute normal and make unit vector
+    // normal = a x b
+    // Note that definition of a and b are such that normal is always out of
+    // page
+    Eigen::Matrix<double, 3, 1> normal_vector = a.cross(b);
+    normal_vector = normal_vector.normalized();
+
+    // Store to private variable
```
kks32: Remove comment

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    face_normals_.insert(std::make_pair<unsigned, Eigen::VectorXd>(
+        static_cast<unsigned>(face_id), normal_vector));
```
kks32: why do you want to static_cast something which is already an unsigned?

include/cell.tcc
```diff
@@ -949,3 +949,97 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+    // the reverse does not work to produce normal that is positive pointing out
+    // of the element
+    Eigen::Matrix<double, 2, 1> normal_vector;
+    normal_vector(0) = -a(1);
+    normal_vector(1) = a(0);
+    normal_vector = normal_vector.normalized();
+
+    // Store to private variable
+    face_normals_.insert(std::make_pair<unsigned, Eigen::VectorXd>(
+        static_cast<unsigned>(face_id), normal_vector));
```
kks32: why static_cast?

include/element.h
```diff
@@ -139,6 +139,9 @@ class Element {
   //! \param[in] face_id given id of the face
   //! \retval indices Indices that make the face
   virtual Eigen::VectorXi face_indices(unsigned face_id) const = 0;
+
+  //! Return number of faces
```
kks32, line 143: Return `the` number of faces

include/hexahedron_element.h
```diff
@@ -235,9 +235,15 @@ class HexahedronElement : public Element<Tdim> {
   //! \retval indices Indices that make the face
   Eigen::VectorXi face_indices(unsigned face_id) const override;
 
+  //! Return number of faces
```
kks32: Return `the` number of faces

include/hexahedron_element.h
```diff
@@ -235,9 +235,15 @@ class HexahedronElement : public Element<Tdim> {
...
+  unsigned nfaces() const { return nfaces_; }
+
  private:
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
+
+  //! Number of faces = 6
+  const unsigned nfaces_{6};
```
kks32: No need for a private variable

include/hexahedron_element.h
```diff
@@ -235,9 +235,15 @@ class HexahedronElement : public Element<Tdim> {
   //! \retval indices Indices that make the face
   Eigen::VectorXi face_indices(unsigned face_id) const override;
 
+  //! Return number of faces
+  unsigned nfaces() const { return nfaces_; }
```
kks32: return 6; here

kks32 (changes requested):

include/cell.h
ezrayst: Understood

include/mesh.tcc
ezrayst: Understood. My bad, I didn't pay attention to the `try` block

include/cell.tcc
ezrayst: for some reason it does not work when I don't do `static_cast` :(

ezrayst:
But we need `Tnphases` to store the values of velocity boundary. Remember that the direction is `0 1 2` for one phase, but if you have two phases you will need directions `3 4 5` too, and so on. For that, we need `Tnphases`.

kks32:
Aren't we storing in a map? Why would we need Tnphases?

ezrayst:
Yes, a map of key and value, with the value being a vector of pair. The first key of the pair is `dir` which needs to be limited to `0 1 2` for one phase, and `0 1 2 3 4 5` for two phases and so on. 

We need to `throw` should the problem is only for one phase, but the input puts direction of, say, `4`. Thus we need `Tnphases`

kks32:
You will throw it in the node, if you are trying to apply something which is incorrect, there is no need to templatise cell.

ezrayst:
OK, got it

include/mesh.h
ezrayst: I am sorry. Not sure when I did this... Might be merging issue

include/cell.tcc
ezrayst: But we don't pass `Nphases`. This information is in `Tnphases`. 

Either way I agree, we don't have to check for them here. When we map to the nodes is when we have `throw` error

include/cell.h
ezrayst: I think I will use it later. Either way, we need to check them right? How do we test them without this function?

include/cell.tcc
ezrayst: Maybe we just allow it and don't check for this? We check in `nodes` as you mentioned?

include/cell.h
kks32: Please don't include functions, which you are not using. This function is not needed, you can get the face_normal directly within the cell. I can't see it being used outside the class.

include/cell.h
```diff
@@ -229,6 +231,27 @@ class Cell {
...
+  void compute_normals();
+
+  //! Return unit normal vector, positive pointing outside of the element
+  //! \param[in] face_id of constraint
+  //! \retval unit normal vector
+  Eigen::VectorXd normal(unsigned face_id) const {
+    return this->face_normals_.at(face_id);
+  }
+
  protected:
```
kks32: Can you change this to `private:`

include/cell.tcc
```diff
@@ -958,3 +958,94 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+
+//! Assign velocity constraint
+//! Constrain directions can take values between 0 and Dim-1
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
+                                                 double velocity) {
+  bool status = true;
+  try {
+    //! Constraint directions can take values between 0 and Dim * Nphases - 1
+    if (dir >= 0 && dir < Tdim) {
```
kks32: Check only if dir is greater than zero.

include/cell.tcc
```diff
@@ -958,3 +958,94 @@ Eigen::VectorXd mpm::Cell<Tdim>::interpolate_nodal_acceleration(
...
+      if (velocity_constraints_.find(face_id) == velocity_constraints_.end())
```
kks32: Remove this block, this is not needed.

include/mesh.h
```diff
@@ -202,6 +209,8 @@ class Mesh {
   Map<NodeBase<Tdim>> map_nodes_;
   //! Container of cells
   Container<Cell<Tdim>> cells_;
+  //! Map of cells for fast retrieval
```
kks32: Don't do this, instead iterate to find the cell pointer and then assign the velocity constraint.

include/mesh.tcc
```diff
@@ -130,6 +130,8 @@ bool mpm::Mesh<Tdim>::create_cells(
 template <unsigned Tdim>
 bool mpm::Mesh<Tdim>::add_cell(const std::shared_ptr<mpm::Cell<Tdim>>& cell) {
   bool insertion_status = cells_.add(cell);
+  // Add cell to map
+  if (insertion_status) map_cells_.insert(cell->id(), cell);
```
kks32: Please don't use `map_cells_` it's more data

include/mesh.tcc
```diff
@@ -378,6 +380,48 @@ bool mpm::Mesh<Tdim>::assign_particles_tractions(
...
+      const unsigned face_id = std::get<1>(velocity_constraint);
+      // Direction of the local coordinate system of the face
+      // Tdim = 2, Normal is y local axis, dir = 1
+      // Tdim = 3, Normal is z local axis, dir = 2
+      const unsigned dir = std::get<2>(velocity_constraint);
+      // Velocity
+      const double velocity = std::get<3>(velocity_constraint);
+
+      // Apply constraint
+      status = map_cells_[cell_id]->assign_velocity_constraint(face_id, dir,
```
kks32: Itreate over cells to `assign_velocity_constraint`

include/mesh.tcc
```diff
@@ -378,6 +380,48 @@ bool mpm::Mesh<Tdim>::assign_particles_tractions(
...
 
+//! Assign velocity constraints to cells
+template <unsigned Tdim>
+bool mpm::Mesh<Tdim>::assign_cell_velocity_constraints(
+    const std::vector<std::tuple<mpm::Index, unsigned, unsigned, double>>&
+        velocity_constraints) {
+  bool status = false;
+  try {
+    // Check the number of nodes
+    if (nodes_.size()) {
```
kks32: Use `!nodes_.size()`

tests/cell_test.cc
```diff
@@ -1,3 +1,5 @@
+#include <cmath>
```
kks32: Why do you need `cmath`?

tests/hexahedron_element_test.cc
```diff
@@ -934,6 +934,10 @@ TEST_CASE("Hexahedron elements are checked", "[hex][element][3D]") {
         for (unsigned j = 0; j < indices.cols(); ++j)
           REQUIRE(check_indices(j) == indices(i, j));
       }
+
+      // Check number of faces
+      const unsigned check_nfaces = 6;
+      REQUIRE(check_nfaces == hex->nfaces());
```
kks32: You check if LHS == 6, not other way around. no need to define `check_nfaces` just add the comment in the line above.

tests/hexahedron_element_test.cc
```diff
@@ -2427,6 +2431,10 @@ TEST_CASE("Hexahedron elements are checked", "[hex][element][3D]") {
...
```
kks32: Revise as above

tests/quadrilateral_element_test.cc
```diff
@@ -620,6 +620,10 @@ TEST_CASE("Quadrilateral elements are checked", "[quad][element][2D]") {
         for (unsigned j = 0; j < indices.cols(); ++j)
           REQUIRE(check_indices(j) == indices(i, j));
       }
+
+      // Check number of faces
+      const unsigned check_nfaces = 4;
+      REQUIRE(check_nfaces == quad->nfaces());
```
kks32: Same as above

tests/quadrilateral_element_test.cc
```diff
@@ -1351,6 +1355,10 @@ TEST_CASE("Quadrilateral elements are checked", "[quad][element][2D]") {
...
```
kks32: Same as above

tests/quadrilateral_element_test.cc
```diff
@@ -2110,6 +2118,10 @@ TEST_CASE("Quadrilateral elements are checked", "[quad][element][2D]") {
...
```
kks32: Same as above

kks32 (changes requested):

tests/cell_test.cc
ezrayst: I forgot to remove since I used it to check something before

include/cell.h
kks32: I thought you were going to remove this, no?

include/cell.h
```diff
@@ -229,7 +231,28 @@ class Cell {
                                     unsigned phase, double pvolume,
                                     const Eigen::Matrix<double, 6, 1>& pstress);
 
- protected:
+  //! Assign velocity constraint
+  //! Directions can take values between 0 and Dim-1
```
kks32: this is no longer `Dim -1`

include/cell.tcc
kks32: Are we going to have this function or no? in this PR?

include/mesh.tcc
```diff
@@ -138,8 +138,8 @@ template <unsigned Tdim>
 bool mpm::Mesh<Tdim>::remove_cell(
     const std::shared_ptr<mpm::Cell<Tdim>>& cell) {
   // Remove a cell if found in the container
-  bool status = cells_.remove(cell);
-  return status;
+  const mpm::Index id = cell->id();
```
kks32: what's the reasoning behind this line?

include/quadrilateral_element.h
```diff
@@ -198,6 +198,9 @@ class QuadrilateralElement : public Element<Tdim> {
   //! \retval indices Indices that make the face
   Eigen::VectorXi face_indices(unsigned face_id) const override;
 
+  //! Return the number of faces
```
kks32: //! Return the number of faces in a quadrilateral

tests/cell_test.cc
```diff
@@ -744,6 +744,64 @@ TEST_CASE("Cell is checked for 2D case", "[cell][2D]") {
...
+      Eigen::Matrix<double, 4, Dim> normal_vector;
+
+      // clang-format off
+      normal_vector <<  0.242535625036333, -0.970142500145332,
+                        0.447213595499958,  0.894427190999916,
+                        0.,                 1.,
+                       -1.,                 0.;
+      // clang-format on
+
+      for (unsigned i = 0; i < 4; ++i) {
```
kks32: No magic numbers, meaning, no standalone arbitrary numbers like `4` here, either add a comment above or a const variable.

tests/mesh_test.cc
```diff
@@ -612,6 +611,19 @@ TEST_CASE("Mesh is checked for 2D case", "[mesh][2D]") {
...
+        SECTION("Check assign velocity constraints to cells") {
+          // Vector of particle coordinates
+          std::vector<std::tuple<mpm::Index, unsigned, unsigned, double>>
+              velocity_constraints;
+          // Constraint
+          velocity_constraints.emplace_back(std::make_tuple(0, 3, 0, 10.5));
+          velocity_constraints.emplace_back(std::make_tuple(1, 2, 1, -10.5));
+
+          REQUIRE(mesh->assign_cell_velocity_constraints(
+                      velocity_constraints) == true);
```
kks32, line 625: what about the failure case?

tests/mesh_test.cc
```diff
@@ -1238,6 +1250,19 @@ TEST_CASE("Mesh is checked for 3D case", "[mesh][3D]") {
...
```
kks32, line 1268: failed case of dir <0

kks32 (changes requested):

tests/mesh_test.cc
ezrayst: but it's not `int`, it's `unsigned`

kks32: OK

tests/mesh_test.cc
ezrayst: I have added the check on `nfaces` as well, so we can have failure case.

include/mesh.tcc
ezrayst: I will revert back. I am not sure what I did :(

ezrayst:
If you will approve this soon, I will not add `Apply velocity constraints` in this PR. I will make the next one. I did add `TODO` within it

kks32:
Could you please remove `apply_velocity_constraints`

ezrayst:
OK

kks32 (approved):

PR #255, opened 2018-09-15
Apply velocity boundary conditions

ezrayst:
- [x] Refactor normal vector to local coordinates
- [x] Develop apply cell velocity constrains functions
- [ ] Create test functions

### DIFF
--- a/include/cell.h
+++ b/include/cell.h
@@ -241,8 +241,8 @@ class Cell {
   //! Apply velocity constraints
   void apply_velocity_constraints();
 
-  //! Compute normal and parallel vectors to the face of the element
-  void compute_normals_and_parallels();
+  //! Compute new coordinate axes of the element
+  void compute_new_coordinate_axes();
 
  private:
   //! cell id
@@ -277,13 +277,9 @@ class Cell {
   std::map<unsigned, std::vector<std::pair<unsigned, double>>>
       velocity_constraints_;
 
-  //! Normal of face
-  //! first-> face_id, second->vector of the normal
-  std::map<unsigned, Eigen::VectorXd> face_normals_;
-
   //! Parallel of face
-  //! first-> face_id, second->vector of the parallel
-  std::map<unsigned, Eigen::MatrixXd> face_parallels_;
+  //! first-> face_id, second->new coordinate axes
+  std::map<unsigned, Eigen::MatrixXd> new_coordinate_axes_;
 
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
--- a/include/cell.h
+++ b/include/cell.h
@@ -238,8 +238,11 @@ class Cell {
   bool assign_velocity_constraint(unsigned face_id, unsigned dir,
                                   double velocity);
 
-  //! Compute normal vector
-  void compute_normals();
+  //! Apply velocity constraints
+  void apply_velocity_constraints();
+
+  //! Compute normal and parallel vectors to the face of the element
+  void compute_normals_and_parallels();
 
  private:
   //! cell id
@@ -277,6 +280,10 @@ class Cell {
   //! Normal of face
   //! first-> face_id, second->vector of the normal
   std::map<unsigned, Eigen::VectorXd> face_normals_;
+
+  //! Parallel of face
+  //! first-> face_id, second->vector of the parallel
+  std::map<unsigned, Eigen::MatrixXd> face_parallels_;
 
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
--- a/include/cell.h
+++ b/include/cell.h
@@ -281,6 +281,16 @@ class Cell {
   //! first-> face_id, second->new coordinate axes
   std::map<unsigned, Eigen::MatrixXd> new_coordinate_axes_;
 
+  //! Constrained nodal velocity
+  std::map<unsigned, Eigen::VectorXd> constrained_nodal_velocity_;
+
+  //! Constrained nodal acceleration
+  std::map<unsigned, Eigen::VectorXd> constrained_nodal_acceleration_;
+
+  //! Geometry class to access functions
+  std::unique_ptr<Geometry<Tdim>> geometry_ =
+      std::make_unique<Geometry<Tdim>>();
+
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
 };  // Cell class
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -969,7 +969,65 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
 
 //! Apply velocity constraints
 template <unsigned Tdim>
-void mpm::Cell<Tdim>::apply_velocity_constraints() {}
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
+
+  // Assum phase = 0 - TODO: make it general
+  const unsigned phase = 0;
+
+  // Copy nodal velocities and accelerations
+  for (unsigned i = 0; i < this->nfunctions(); ++i) {
+    constrained_nodal_velocity_.insert(
+        std::make_pair<unsigned, Eigen::VectorXd>(i,
+                                                  nodes_[i]->velocity(phase)));
+    constrained_nodal_acceleration_.insert(
+        std::make_pair<unsigned, Eigen::VectorXd>(
+            i, nodes_[i]->acceleration(phase)));
+  }
+
+  // Set velocity constraint
+  for (auto iterator = this->velocity_constraints_.begin();
+       iterator != this->velocity_constraints_.end(); ++iterator) {
+    // Get face id
+    const auto face_id = iterator->first;
+    // Loop through all the velocity constraints
+    for (const auto& constraint : this->velocity_constraints_[face_id]) {
+      // Direction value in the constraint (0, Dim * Nphases)
+      const auto dir = constraint.first;
+      // Direction: dir % Tdim (modulus)
+      const auto direction = static_cast<unsigned>(dir % Tdim);
+      // Phase: Integer value of division (dir / Tdim)
+      // const auto phase = static_cast<unsigned>(dir / Tdim);
+
+      // Get inverse rotation matrix
+      Eigen::Matrix<double, Tdim, Tdim> inverse_rotation_matrix =
+          geometry_->inverse_rotation_matrix(geometry_->euler_angles_cartesian(
+              this->new_coordinate_axes_[face_id]));
+
+      // Get the nodes of the face
+      const Eigen::VectorXi node_indices = element_->face_indices(face_id);
+
+      // Apply it to the nodes
+      for (unsigned j = 0; j < node_indices.size(); ++j) {
+        auto nodal_velocity = constrained_nodal_velocity_[node_indices(j)];
+        auto nodal_acceleration =
+            constrained_nodal_acceleration_[node_indices(j)];
+
+        // Rotate velocity and acceleration vectors to new coordinate axes
+        nodal_velocity *= inverse_rotation_matrix;
+        nodal_acceleration *= inverse_rotation_matrix;
+
+        // Apply velocity boundary conditions in new coordinate axes
+        nodal_velocity(direction) = constraint.second;
+        nodal_acceleration(direction) = 0.;
+
+        // Rotate velocity and acceleration vectors back to original coordinate
+        // axes
+        nodal_velocity *= inverse_rotation_matrix.inverse();
+        nodal_acceleration *= inverse_rotation_matrix.inverse();
+      }
+    }
+  }
+}
 
 //! Compute all face normals and parallels 2d
 template <>
@@ -1026,7 +1084,7 @@ inline void mpm::Cell<3>::compute_new_coordinate_axes() {
                                     (this->nodes_[indices(0)])->coordinates();
 
     // Compute normal and make unit vector
-    // normal = a x b
+    // normal = a x b (non-commutative)
     // Note that definition of a and b are such that normal is always out of
     // page
     Eigen::Matrix<double, 3, 1> normal_vector = a.cross(b);
@@ -1034,6 +1092,7 @@ inline void mpm::Cell<3>::compute_new_coordinate_axes() {
 
     // Assume a as the new x-axis, and normal as the new z-axis, the new y-axis
     // has to be orthogonal to a and normal parallel_vector = normal_vector x a
+    // (non-commutative)
     Eigen::Matrix<double, 3, 3> new_coordinate_axes;
     new_coordinate_axes.col(0) = a.normalized();
     new_coordinate_axes.col(1) = (normal_vector.cross(a)).normalized();
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -967,37 +967,49 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
   return status;
 }
 
-//! Compute all face normals 2d
+//! Apply velocity constraints
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
+
+  // TODO
+}
+
+//! Compute all face normals and parallels 2d
 template <>
-inline void mpm::Cell<2>::compute_normals() {
+inline void mpm::Cell<2>::compute_normals_and_parallels() {
 
   //! Set number of faces from element
   for (unsigned face_id = 0; face_id < element_->nfaces(); ++face_id) {
     // Get the nodes of the face
     const Eigen::VectorXi indices = element_->face_indices(face_id);
 
-    // Compute the vector to calculate normal (perpendicular)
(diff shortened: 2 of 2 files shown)


include/cell.h
```diff
@@ -274,9 +277,19 @@ class Cell {
   std::map<unsigned, std::vector<std::pair<unsigned, double>>>
       velocity_constraints_;
 
-  //! Normal of face
-  //! first-> face_id, second->vector of the normal
-  std::map<unsigned, Eigen::VectorXd> face_normals_;
+  //! Parallel of face
```
kks32, line 280: What do you mean by `parallel of face`?

include/cell.h
```diff
@@ -274,9 +277,19 @@ class Cell {
...
+  //! first-> face_id, second->new coordinate axes
+  std::map<unsigned, Eigen::MatrixXd> new_coordinate_axes_;
+
+  //! Constrained nodal velocity
+  std::map<unsigned, Eigen::VectorXd> constrained_nodal_velocity_;
```
kks32, line 285: `nodal_velocity_constraints_`

include/cell.h
```diff
@@ -274,9 +277,19 @@ class Cell {
...
+
+  //! Constrained nodal acceleration
+  std::map<unsigned, Eigen::VectorXd> constrained_nodal_acceleration_;
```
kks32, line 288: `nodal_acceleration_constraints_`

include/cell.h
```diff
@@ -274,9 +277,19 @@ class Cell {
...
+
+  //! Geometry class to access functions
+  std::unique_ptr<Geometry<Tdim>> geometry_ =
```
kks32, line 291: Geometry should be a namespace, not an object.

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
   return status;
 }
 
-//! Compute all face normals 2d
+//! Apply velocity constraints
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
```
kks32, line 972: Can you not pass the `phase` here?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+    constrained_nodal_velocity_.insert(
+        std::make_pair<unsigned, Eigen::VectorXd>(i,
+                                                  nodes_[i]->velocity(phase)));
+    constrained_nodal_acceleration_.insert(
+        std::make_pair<unsigned, Eigen::VectorXd>(
+            i, nodes_[i]->acceleration(phase)));
+  }
+
+  // Set velocity constraint
+  for (auto iterator = this->velocity_constraints_.begin();
```
kks32, line 988: Use `itr`

include/cell.h
kks32: You already have `velocity_constraints_`, why do you need these two? why can't be local to the apply velocity constraints function?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
-//! Compute all face normals 2d
+//! Apply velocity constraints
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::apply_velocity_constraints() {
+
+  // Assum phase = 0 - TODO: make it general
+  const unsigned phase = 0;
+
+  // Copy nodal velocities and accelerations
+  for (unsigned i = 0; i < this->nfunctions(); ++i) {
```
kks32: Why do we need to copy this information from the nodes, is it not sufficient to tell the nodes what the velocities and accelerations should be? You shouldn't need to know what the velocity currently is, to do this.

include/cell.h
```diff
@@ -274,9 +277,19 @@ class Cell {
   std::map<unsigned, std::vector<std::pair<unsigned, double>>>
       velocity_constraints_;
 
-  //! Normal of face
-  //! first-> face_id, second->vector of the normal
-  std::map<unsigned, Eigen::VectorXd> face_normals_;
+  //! Parallel of face
+  //! first-> face_id, second->new coordinate axes
+  std::map<unsigned, Eigen::MatrixXd> new_coordinate_axes_;
```
kks32, line 282: Do you mean to say it as the local coordinates?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+    // Get face id
+    const auto face_id = iterator->first;
+    // Loop through all the velocity constraints
+    for (const auto& constraint : this->velocity_constraints_[face_id]) {
+      // Direction value in the constraint (0, Dim * Nphases)
+      const auto dir = constraint.first;
+      // Direction: dir % Tdim (modulus)
+      const auto direction = static_cast<unsigned>(dir % Tdim);
+      // Phase: Integer value of division (dir / Tdim)
+      // const auto phase = static_cast<unsigned>(dir / Tdim);
```
kks32, line 999: any reason why you can't do phase?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+      Eigen::Matrix<double, Tdim, Tdim> rotation_matrix =
+          geometry_->rotation_matrix(geometry_->euler_angles_cartesian(
+              this->new_coordinate_axes_[face_id]));
+
+      // Get the nodes of the face
+      const Eigen::VectorXi node_indices = element_->face_indices(face_id);
+
+      // Apply it to the nodes
+      for (unsigned j = 0; j < node_indices.size(); ++j) {
+        // Rotate velocity and acceleration vectors to new coordinate axes
```
kks32: You are resetting this value three lines below, why are we doing this?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+
+      // Get inverse rotation matrix
+      Eigen::Matrix<double, Tdim, Tdim> rotation_matrix =
+          geometry_->rotation_matrix(geometry_->euler_angles_cartesian(
+              this->new_coordinate_axes_[face_id]));
+
+      // Get the nodes of the face
+      const Eigen::VectorXi node_indices = element_->face_indices(face_id);
+
+      // Apply it to the nodes
```
kks32, line 1009: I can't see it being applied to any node here! There is no access to a single nodal pointer. Am I missing something?

kks32 (changes requested):
Could you please explain how the nodal velocities are getting updated here?

kks32 (changes requested):
Ignore this!

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+      for (unsigned j = 0; j < node_indices.size(); ++j) {
+        // Rotate velocity and acceleration vectors to new coordinate axes
+        constrained_nodal_velocity_[node_indices(j)] *=
+            rotation_matrix.inverse();
```
kks32, line 1013: Can we take the transpose instead?

include/cell.tcc
```diff
@@ -979,37 +979,106 @@ bool mpm::Cell<Tdim>::assign_velocity_constraint(unsigned face_id, unsigned dir,
...
+
+        // Apply velocity boundary conditions in new coordinate axes
+        constrained_nodal_velocity_[node_indices(j)](direction) =
+            constraint.second;
+        constrained_nodal_acceleration_[node_indices(j)](direction) = 0.;
+
+        // Rotate velocity and acceleration vectors back to original coordinate
+        // axes
+        constrained_nodal_velocity_[node_indices(j)] *= rotation_matrix;
```
kks32, line 1024: Why aren't we returning these variables, why is it stored in the cell level?

kks32 (changes requested):

ezrayst:
Proposed structure of the code for applying velocity boundary:
- [ ] The input file would still contain boundary constraints on `face`
- [ ] Boundary constraints are stored in `cell` because we need the data of the `face`
- [ ] Make a proxy to indicate the boundary are defined at the `face` or by the `nodes`
- [ ] Node's `compute_acceleration_velocity` should contain `apply_velocity_boundary` for node-based or cell-based boundary
- [ ] In cell based, we have to iterate through all the faces this boundary: (1) rotate the current velocity, (2) apply all boundaries in new local coordinate system, (3) rotate back to Cartesian

Note the following:
- Using Krishna's example of isoparametric Element 0 and Element 1, and roller support on both bottom faces, node 0 would not be constrained to move in all directions, so this method of applying boundary is, in my opinion, sound. 
- Using resultant to apply boundary would not be right, I think, as we are forcing certain components to be zero, which are not equivalent of making the resultant of that zero.

kks32:
* How do you plan to achieve ` Make a proxy to indicate the boundary are defined at the face or by the nodes`?

* I'd suggesting a boolean at the node to have `cell_velocity_constraints_` if true, we apply those. 

* Ideally it would be better for the node to have access, not necessarily store, all the constraints. If we do that, then it is easier to iterate.

* We need to consider the special case of 0, when zero velocity is applied on more than one direction, that plane will have zero velocity. Does this make it computationally faster? 

* We should never do resultant, rather it should be a linear summation of all constraints at a node.

* Furthermore, this would require changing `assign_cell_velocity_constraints` functions, before we modify the apply constraints.

kks32:
Could you provide more information on what kind of data structure will be used to store the constraints at the cell and the node level?

ezrayst:
- I am fine with `boolean` for `cell_velocity_constraints_`
- The nodes could have access, but the `face_id` should dictate it. `face_id` will give local numbering (cell level) of the nodes, which I think we can map to the global nodes.
- After thinking about it again, we should not apply boundary when calculating the acceleration and velocity. I think it's cleaner to apply them after all nodes have been calculated. So we iterate over all the constraints and apply them to all nodes.
- Agree on the resultant

> * We need to consider the special case of 0, when zero velocity is applied on more than one direction, that plane will have zero velocity. Does this make it computationally faster?

On this, I think let's just apply it as it is, then we can optimize later. What do you think? 

Data Structure:
- Storing constraints would remain the same for cell level now, which is a `std::map<unsigned, std::vector<std::pair<unsigned, double>>>` with the key being `face_id`, and the pair containing `direction` in new local coordinate and the value of the velocity
- I don't think we need to store the nodes. We just iterate over the `map`, and for each `face_id`, we will get all the affected local `node_id`, map them to global `node_id` and apply to all affected nodes
- Note that at each `apply` include: (1) rotate current velocity and acceleration, (2) apply the values of velocities and accelerations, (2) rotate back. So velocity and acceleration are always in Cartesian coordinate

kks32:
We apply nodal velocity constraints on Cartesian grids at the time of computation of acceleration and velocity, this way, when velocity is calculated at the node we always get the right velocity. I don't see why inclined constraints can't be applied at the same time. I do not get the argument of it being clean, because doing so will return velocities and accelerations which are incorrect (before restraints are applied). That's not good.

Nodes having access to all its constraints means we can optimise for the case of zero. Doing at the cell level, we won't have enough information to find which two planes have zero velocity. To me, node should have access to all the constraints, and this should be done at the assign_cell_constraints step. 

Let me know if you think this won't work.

kks32:
I'm not clear about `The nodes could have access, but the face_id should dictate it. face_id will give local numbering (cell level) of the nodes, which I think we can map to the global nodes.`, how this mapping will be done, what information will be transferred to the nodes, how and when will it be done?

kks32:
`I don't think we need to store the nodes. We just iterate over the map, and for each face_id, we will get all the affected local node_id, map them to global node_id and apply to all affected nodes` This would require adding an extra function and ensuring that this function is definitely called before returning the nodal velocities and acceleration. There is no guarantee that someone might forget to call this function, and hence will get unconstrained velocities and accelerations.

ezrayst:
Well, let's agree that the velocity boundaries will be defined on the face. Is this good? I don't see how we can do better in the other ways. 

If we are good on this, and you want to apply velocity and acceleration as we calculate them, we have to transfer those `face` velocities to `node` velocities, and use the latter when we apply them. We can write a function for this in `cell` to do that, which calls on something like `assign_cell_velocity_boundary` in `node` to store. I think the data structure should be a map of direction and a pair of direction and values.

But we are back at the issue I raised just now, about mapping from local node number we have from the face, to the global when we store them in `node`. What do you think?

kks32:
`Well, let's agree that the velocity boundaries will be defined on the face. Is this good? I don't see how we can do better in the other ways.`
- Agree the velocity boundaries for inclined planes have to be applied on the face. For each face constraint, we could store the rotation matrix, direction and constraint velocity at the cell level as a map with key being the faceid. This might reduce the number of times we calculate the rotation matrix at each node. For each cell with a face constraint, we have to do it at least 4 times. Storing it means we do it once. It's a balance between size vs efficiency. Since the shape of the cell doesn't change, it's better to store the rotation matrix, than to calculate it at each step.

`If we are good on this, and you want to apply velocity and acceleration as we calculate them, we have to transfer those face velocities to node velocities, and use the latter when we apply them. `
 - Agree we need to transfer it to the nodes, so each shared node with cell velocity constraints has all the information it needs. 

`We can write a function for this in cell to do that, which calls on something like assign_cell_velocity_boundary in node to store. I think the data structure should be a map of direction and a pair of direction and values.`
- Why not use the `assign_cell_velocity_constraints` function to create and assign the constraints to the nodes, why need a new function?

`But we are back at the issue I raised just now, about mapping from local node number we have from the face, to the global when we store them in node. What do you think?`
- I don't see any issue here, the cell has nodal pointers, what is the problem?

ezrayst:
> Agree the velocity boundaries for inclined planes have to be applied on the face. For each face constraint, we could store the rotation matrix, direction and constraint velocity at the cell level as a map with key being the faceid. This might reduce the number of times we calculate the rotation matrix at each node. For each cell with a face constraint, we have to do it at least 4 times. Storing it means we do it once. It's a balance between size vs efficiency. Since the shape of the cell doesn't change, it's better to store the rotation matrix, than to calculate it at each step.
- Yes, we need the rotation matrix too. I will add the `rotation matrix` as well to the map then.

> Why not use the assign_cell_velocity_constraints function to create and assign the constraints to the nodes, why need a new function?
- Agree, we can use that function as well. 

> I don't see any issue here, the cell has nodal pointers, what is the problem?
- I see! I am OK then. I can work with that


Then all we have is the `apply` part. I guess we will use `if` function. If we have `rotation matrix`, then we will rotate, apply and rotate back. If not, we will simply apply it as is.

When we have more phases, we will follow the current scheme with directions of 4, 5, 6 etc. That would be checked only when nodal constraint is passed to `nodes`, and would not be checked at `cell` because it does not have `Tnphases`. Would this be OK?

kks32:
`Yes, we need the rotation matrix too. I will add the rotation matrix as well to the map then.`
- Create a map of face_id to a shared_pointer of a struct object for the constraint.

ezrayst:
The struct object means the map containing rotation matrix, direction and velocity value?

kks32:
How are you planning to store `rotation matrix, direction and velocity`?

ezrayst:
I was thinking of a `map` of a `tuple`. The key is `face_id` and the tuple would consist `Eigen::MatrixXd rotation matrix`, `unsigned direction`, and `double velocity value`. So:
- `cell` stores a map with key: `face_id`, and the `tuple` as described above
- we make another private variable in `node` called `face_velocity_constraints_` which is a `vector` of `tuple` containing `rotation matrix` and a pair of `direction` and `velocity`.
- we have another function to `assign` to node constraint
- In `apply`, we can have `boolean` to see whether `face_velocity_constraints_` are empty. If it is, then go with the node scheme, otherwise, go with the cell-node scheme

kks32:
This would mean the rotation matrix and the constraints will be copied redundantly from cells to each node in which the constraint is applied. If we assume the rotation matrix has 9 components of doubles, that's 9 * 8 bytes * 4 (nodes on a face) 288 bytes of redundant storage just for the rotation matrices. Tuples are not the best way to pass the constraints from cell to nodes.

ezrayst:
> Create a map of face_id to a shared_pointer of a struct object for the constraint.
- So you mean to have pointer to the whole `cell`? I don't quite get it

kks32:
You can have a pointer to a constraint object. Create a constraint object in the cell, store it as a map <face_id, ptr_to_constraints>, `Nodes` have  a vector of pointers to constraints.

ezrayst:
I will try then. Let me try to implement.

ezrayst:
Wait, I will wait until our meeting today, then I will work on it again.

kks32:
Closing this PR as Velocity constraints have to be redesigned to be mapped directly from cell to nodes.

PR #262, opened 2018-09-24
Change registration name of element

ezrayst:
Since `node` uses registration name of `N2D` and `N3D` and `particle` uses `P2D` and `P3D`, it makes more sense for `elements` to use `E2D**` and `E3D**` instead of `ED2**` and `ED3**`

Thus I have changed them accordingly

### DIFF
--- a/src/element.cc
+++ b/src/element.cc
@@ -5,18 +5,18 @@
 
 // Quadrilateral 4-noded element
 static Register<mpm::Element<2>, mpm::QuadrilateralElement<2, 4>> quad4(
-    "ED2Q4");
+    "E2DQ4");
 
 // Quadrilateral 8-noded element
 static Register<mpm::Element<2>, mpm::QuadrilateralElement<2, 8>> quad8(
-    "ED2Q8");
+    "E2DQ8");
 
 // Quadrilateral 9-noded element
 static Register<mpm::Element<2>, mpm::QuadrilateralElement<2, 9>> quad9(
-    "ED2Q9");
+    "E2DQ9");
 
 // Hexahedron 8-noded element
-static Register<mpm::Element<3>, mpm::HexahedronElement<3, 8>> hex8("ED3H8");
+static Register<mpm::Element<3>, mpm::HexahedronElement<3, 8>> hex8("E3DH8");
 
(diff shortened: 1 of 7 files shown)


kks32 (changes requested):
The suggested change is internally inconsistent for the elements. Numbers come before the dimension and after the element type. I'd suggest we adopt Abaqus numbering, which people would be familiar http://web.mit.edu/calculix_v2.7/CalculiX/ccx_2.7/doc/ccx/node194.html

kks32:
http://www2.mae.ufl.edu/haftka/finite-element/ABAQUS-elements.pdf

ezrayst:
The numbering in Abaqus includes the model type: plane strain, plane stress, axissymetric. Do you want to include those as well?

I think it's more intuitive to have used Q and H for quadrilateral and hexahedron.

In addition, about numbering before the element, we can change it to `E3D4Q` if you want, but isn't normal for people to say 3D and to say Q4 or H8? I get it we are inconsistent about it, but it is consistent to the way we call them

kks32:
No, we can use `C3D8` for 8-noded hexahedron. `C2D4` for a 4-noded triangle to be similar to Abaqus naming. `Q` and `H` are redundant. If we end up having an element that is in conflict with our naming, we can think about that element later.

ezrayst:
I am fine as long as we are consistent. I still feel `C2DQ4` would be better because it is clear that it is a quadrilateral. It's intuitive. 

I still don't get what does `C` stand for? Continuum?

kks32:
I'm not sure what `C` stands for. Either we stick to how Abaqus calls it `C3D8` or if we really want to use `Q` or `H`, then we do `E3D8Q`, numbers before letters. We use `E` to denote element.

ezrayst:
I am fine with using `C` and then to use `Q` and `H` too. I will make the changes

kks32:
There was a reasoning behind numbering `ED2Q9`, where the first character refers to the type. `A0` sequence refers to Type and the Number the unit. Anything just a character will denote special case at the end. so `R` for reduced integration or `G` for GIMP element. Hence closing this PR.

PR #266, opened 2018-09-28
Fix yield condition from the square of tau to the trace of that

ezrayst:
- [x] Change yield condition. It used to be squared of tau only instead of the trace of that
- [x] Update test cases

### DIFF
--- a/tests/material/bingham_test.cc
+++ b/tests/material/bingham_test.cc
@@ -274,8 +274,8 @@ TEST_CASE("Bingham is checked in 2D", "[material][bingham][2D]") {
     jmaterial["density"] = 1000.;
     jmaterial["youngs_modulus"] = 1.0E+7;
     jmaterial["poisson_ratio"] = 0.3;
-    jmaterial["tau0"] = 771.8;
-    jmaterial["mu"] = 0.0451;
+    jmaterial["tau0"] = 200;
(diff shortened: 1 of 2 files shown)


tests/material/bingham_test.cc
```diff
@@ -274,8 +274,8 @@ TEST_CASE("Bingham is checked in 2D", "[material][bingham][2D]") {
     jmaterial["density"] = 1000.;
     jmaterial["youngs_modulus"] = 1.0E+7;
     jmaterial["poisson_ratio"] = 0.3;
-    jmaterial["tau0"] = 771.8;
-    jmaterial["mu"] = 0.0451;
+    jmaterial["tau0"] = 200;
+    jmaterial["mu"] = 200;
```
kks32, line 278: What's the unit here, there seem to be a big jump in the value here

kks32 (changes requested):

tests/material/bingham_test.cc
ezrayst: `tau0` is [Pa]
`mu` is [N s / m^2 or kg / m / s]
Both are written in `bingham.h`

This is a test case and in this test, we want to make sure it yields. Thus I have to find parameters to condition them

kks32 (approved):

PR #269, opened 2018-10-03
Conserve mass but change volume

ezrayst:
- [x] Change volume computation, using volumetric strain
- [x] I don't want to change `material`, so I add a private variable in `particle` called `density_correction_` so that `density_correction_ * density_from_material = density_updated`. 
- [x] Update the compute mass functions accordingly
- [ ] Update test functions 

This has been tested in 2d, with the material points crossing the cell.

### DIFF
--- a/include/particle.h
+++ b/include/particle.h
@@ -235,6 +235,8 @@ class Particle : public ParticleBase<Tdim> {
   Eigen::VectorXd shapefn_;
   //! B-Matrix
   std::vector<Eigen::MatrixXd> bmatrix_;
+  //! Density correction because of volume change
+  double density_correction_;
   //! Logger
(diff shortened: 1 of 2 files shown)


kks32 (commented):
Mass can be conserved without updating it, there is no reason to multiply and divide by the same factor to conserve mass.

On the other hand, volume I think should be calculated based on the volumetric strain calculated at the centre of the cell.

kks32:
Closing this PR because: 
Mass can be conserved without updating it, there is no reason to multiply and divide by the same factor to conserve mass.

On the other hand, volume I think should be calculated based on the volumetric strain calculated at the centre of the cell. 

PRs #272 fixes the first issue of compute mass and volume.

PR #275, opened 2018-10-03
:checkered_flag: Update volume based on volumetric strain fixes #176

### LINKED ISSUE #176: Update density/volume of material point
* Update the density of material point based on new volume, calculated using the volumetric strain of the cell.

kks32:
Update particle volume with strain rate centroid. Density is used only to calculate the particle mass at the start of the time-step. Hence, density is neither stored nor updated in the algorithm. If required, it could be added to `update_volume_strainrate`

### DIFF
--- a/include/particle.h
+++ b/include/particle.h
@@ -90,6 +90,11 @@ class Particle : public ParticleBase<Tdim> {
   //! \param[in] phase Index corresponding to the phase
   bool compute_volume(unsigned phase) override;
 
+  //! Update volume based on centre volumetric strain rate
+  //! \param[in] phase Index corresponding to the phase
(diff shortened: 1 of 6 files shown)


include/particle.h
```diff
@@ -227,6 +232,8 @@ class Particle : public ParticleBase<Tdim> {
   Eigen::Matrix<double, Tnphases, 1> volumetric_strain_centroid_;
   //! Strain rate
   Eigen::Matrix<double, 6, Tnphases> strain_rate_;
+  //! Strain rate centroid
```
cbgeo: Remove this!

cbgeo (changes requested):

cbgeo (approved):

PR #282, opened 2018-10-09
:checkered_flag: Fixes #281 GIMP issue on calculating strain at the centroid

### LINKED ISSUE #281: Cell needs particle size and deformation gradient to calculate the centroidal strain
GIMP shapefns require particle size and deformation gradient for calculating strain at the centroid of the cell. The call should be handled differently for GIMP and normal MPM. Since this is the centroid of the cell, GIMP neighbouring nodes will never be activated, hence, we can use Eigen::Matrix::Zero as size and deformation gradient.

cw646:
Changes to cell and particle calls of shapefn and bmatrix, add missing call to tnfunctions in gimp header.

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -849,8 +849,23 @@ Eigen::VectorXd mpm::Cell<Tdim>::compute_strain_rate_centroid(unsigned phase) {
   Eigen::Matrix<double, Tdim, 1> xi_centroid;
   xi_centroid.setZero();
 
-  // Get B-Matrix at the centroid
-  auto bmatrix = element_->bmatrix(xi_centroid, this->nodal_coordinates());
+  //! B-Matrix
+  std::vector<Eigen::MatrixXd> bmatrix;
+
+  if (element_->shapefn_type() == mpm::ShapefnType::GIMP) {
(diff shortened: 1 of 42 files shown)


include/cell.tcc
```diff
@@ -849,8 +849,23 @@ Eigen::VectorXd mpm::Cell<Tdim>::compute_strain_rate_centroid(unsigned phase) {
   Eigen::Matrix<double, Tdim, 1> xi_centroid;
   xi_centroid.setZero();
 
-  // Get B-Matrix at the centroid
-  auto bmatrix = element_->bmatrix(xi_centroid, this->nodal_coordinates());
+  //! B-Matrix
+  std::vector<Eigen::MatrixXd> bmatrix;
+
+  if (element_->shapefn_type() == mpm::ShapefnType::GIMP) {
```
kks32, line 855: It doesn't matter if it is GIMP or CPDI it's the same call.

kks32 (changes requested):

kks32 (changes requested):
Add tests in particle class to check for both normal and special shape function calls

cbgeo:
Superseded by #285 . Closing without merge.

PR #294, opened 2018-10-15
:checkered_flag: Use local sf and Jacobian for computing local cordinates in GIMP

kks32:
Affine transform and NR needs local sf and Jacobian in GIMP


cw646 (commented):
Did we need the

``` Eigen::VectorXd sf;
    sf.resize(indices.size());

    for (unsigned i = 0; i < indices.size(); ++i) sf(i) = sf1(i);
```
We added previously?

kks32:
we don't because we have the local shape function

cbgeo (approved):

PR #323, opened 2018-10-30
:checkered_flag: Read large files without checking for duplicate item ids in mesh generation

kks32:
* Reading large files take a huge time in creating meshes and cells, this reduces the time by having a flag to disable checking for duplicates in the `mesh` object in input configuration.

`check_duplicates` is set to true by default, if not specified is an optional parameter.


include/container.h
```diff
@@ -23,7 +23,8 @@ class Container {
 
   //! Add a pointer to an element
   //! \param[in] ptr A shared pointer
-  bool add(const std::shared_ptr<T>&);
+  //! \param[in[ check_duplicates Parameter to check duplicates
```
cbgeo:
```suggestion
  //! \param[in] check_duplicates Parameter to check duplicates
```

cbgeo (changes requested):

cbgeo (approved):

PR #324, opened 2018-10-31
Pressure smoothing

ezrayst:
- [x] Implement pressure smoothing which involves mapping and average weight.
- [x] test cases are made

image

Results: (top is normal, bottom is with pressure smoothing)
image

### DIFF
--- a/include/mpm_explicit_usf.tcc
+++ b/include/mpm_explicit_usf.tcc
@@ -144,10 +144,15 @@ bool mpm::MPMExplicitUSF<Tdim>::solve() {
     }
 #endif
 
-    // Iterate over each particle to compute pressure smoothing
-    mesh_->iterate_over_particles(
-        std::bind(&mpm::ParticleBase<Tdim>::compute_pressure_smoothing,
-                  std::placeholders::_1, phase));
+    // Iterate over each particle to compute pressure smoothing when specified
+    bool pressure_smoothing = false;
+    if (analysis_.find("pressure_smoothing") != analysis_.end()) {
+      pressure_smoothing = analysis_["pressure_smoothing"].template get<bool>();
+      if (pressure_smoothing)
+        mesh_->iterate_over_particles(
+            std::bind(&mpm::ParticleBase<Tdim>::compute_pressure_smoothing,
+                      std::placeholders::_1, phase));
+    }
 
     // Iterate over each particle to compute stress
     mesh_->iterate_over_particles(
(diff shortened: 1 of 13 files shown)


include/mpm_explicit_usf.tcc
```diff
@@ -117,6 +117,16 @@ bool mpm::MPMExplicitUSF<Tdim>::solve() {
         std::bind(&mpm::ParticleBase<Tdim>::compute_strain,
                   std::placeholders::_1, phase, dt_));
 
+    // Assign pressure to nodes
+    mesh_->iterate_over_particles(
+        std::bind(&mpm::ParticleBase<Tdim>::map_pressure_to_nodes,
+                  std::placeholders::_1, phase));
+
```
kks32, line 133: Need an MPI all_reduce call here

include/mpm_explicit_usf.tcc
```diff
@@ -117,6 +117,16 @@ bool mpm::MPMExplicitUSF<Tdim>::solve() {
         std::bind(&mpm::ParticleBase<Tdim>::compute_strain,
                   std::placeholders::_1, phase, dt_));
 
+    // Assign pressure to nodes
```
kks32, line 129: This whole block should be optional and done only if pressure_smoothing is enabled

kks32 (changes requested):
- [ ] Pressure smoothing should be an optional argument that should be disabled by default.
- [ ] You need an MPI all_reduce scalar on the pressure term
- [ ] How does pressure smoothing work when more than a single material type is used, does it affect all the materials? Shouldn't it only smooth those where pressure can be computed?
- [ ] Need both USF and USL to be updated

include/mpm_explicit_usf.tcc
```diff
@@ -124,6 +124,34 @@ bool mpm::MPMExplicitUSF<Tdim>::solve() {
...
+    // Run if there is more than a single MPI task
+    if (mpi_size > 1) {
+      // MPI all reduce nodal pressure
+      mesh_->allreduce_nodal_scalar_property(
+          std::bind(&mpm::NodeBase<Tdim>::pressure, std::placeholders::_1,
+                    phase),
+          std::bind(&mpm::NodeBase<Tdim>::update_pressure,
+                    std::placeholders::_1, false, phase, std::placeholders::_2,
+                    std::placeholders::_2));
+    }
```
ezrayst, line 144: I can't use `std::placeholders::_3` for some reason. Note that `update_pressure` has 4 input arguments: `update boolean`, `phase`, `mass` and `pressure`.

ezrayst:
I am not sure about what to do with different material id currently. Do we have only a pressure for all the phases? What if we have `Bingham` for solid and `Newtonian` for fluid in 2-phase MPM?

kks32:
#396 supersedes this PR so closing

PR #350, opened 2019-01-14
:construction: MohrCoulomb model

srhgk2:
Add new modified mohr coulomb material model. There is still work remaining in the softening part.

### DIFF
--- a/include/material/mohr_coulomb.h
+++ b/include/material/mohr_coulomb.h
@@ -62,12 +62,12 @@ class MohrCoulomb : public Material<Tdim> {
   bool compute_elastic_tensor();
   //! Compute j2, j3, rho, theta
   bool compute_rho_theta(const Vector6d stress, double& _j2, double& _j3,
-			 double& _rho, double& _theta);
+                         double& _rho, double& _theta);
   //! Compute dF/dSigma and dP/dSigma
-  bool compute_df_dp(const double _j2, const double _j3,
-		     const double _rho, const double _theta,
-		     const Vector6d stress, Vector6d& _dF_dSigma,
-		     Vector6d& _dP_dSigma, double& _softening);
+  bool compute_df_dp(const double _j2, const double _j3, const double _rho,
+                     const double _theta, const Vector6d stress,
+                     Vector6d& _dF_dSigma, Vector6d& _dP_dSigma,
+                     double& _softening);
 
  private:
   //! Elastic stiffness matrix
--- a/include/material/mohr_coulomb.h
+++ b/include/material/mohr_coulomb.h
@@ -1,0 +1,118 @@
+#ifndef MPM_MATERIAL_MOHR_COULOMB_H_
+#define MPM_MATERIAL_MOHR_COULOMB_H_
+
+#include <limits>
+
+#include "Eigen/Dense"
+
+#include "material.h"
+
+namespace mpm {
+
+//! MohrCoulomb class
+//! \brief Linear Elastic material model
+//! \details MohrCoulomb class stresses and strains
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class MohrCoulomb : public Material<Tdim> {
+ public:
+  //! Define a vector of 6 dof
+  using Vector6d = Eigen::Matrix<double, 6, 1>;
+  //! Define a Matrix of 6 x 6
+  using Matrix6x6 = Eigen::Matrix<double, 6, 6>;
+
+  //! Constructor with id and material properties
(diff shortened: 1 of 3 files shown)


include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,118 @@
...
+#include <limits>
+
+#include "Eigen/Dense"
+
+#include "material.h"
+
+namespace mpm {
+
+//! MohrCoulomb class
+//! \brief Linear Elastic material model
```
kks32, line 13: MohrCoulomb

include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,118 @@
...
+  //! Material properties
+  using Material<Tdim>::properties_;
+  //! Logger
+  using Material<Tdim>::console_;
+
+ private:
+  //! Compute elastic tensor
+  bool compute_elastic_tensor();
+  //! Compute j2, j3, rho, theta
+  bool compute_rho_theta(const Vector6d stress, double& _j2, double& _j3,
```
kks32: Could we have `j2`, `j3`, `rho`, `theta` and `softening` as private variables in the class, instead of passing around in the functions?

include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,118 @@
...
+  //! Porosity
+  double porosity_{std::numeric_limits<double>::max()};
+  //! Permeability
+  double permeability_{std::numeric_limits<double>::max()};
+  //! Friction angle phi
+  double phi_{std::numeric_limits<double>::max()};
+  //! Dilation angle psi
+  double psi_{std::numeric_limits<double>::max()};
+  //! Cohesion
+  double c_{std::numeric_limits<double>::max()};
```
kks32, line 110: Is this cohesion `c_` different from `cohesion_`?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,350 @@
...
+  // clang-format on
+  return true;
+}
+
+//! Return j2, j3, rho and theta
+template <unsigned Tdim>
+bool mpm::MohrCoulomb<Tdim>::compute_rho_theta(const Vector6d stress,
+                                               double& _j2, double& _j3,
+                                               double& _rho, double& _theta) {
+  const double ONETHIRDPI = 1.047197551;
```
kks32, line 64: Needs more digits for double

include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,127 @@
...
+  //! Material properties
+  using Material<Tdim>::properties_;
+  //! Logger
+  using Material<Tdim>::console_;
+
+ private:
+  //! Compute elastic tensor
+  bool compute_elastic_tensor();
+  //! Compute j2, j3, rho, theta
+  void compute_rho_theta(const Vector6d& stress);
```
kks32, line 64: We need to add `\param` parameters

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,370 @@
...
+
+  rho_ = sqrt(2. * j2_);
+}
+
+//! Compute dF/dSigma and dP/dSigma
+template <unsigned Tdim>
+void mpm::MohrCoulomb<Tdim>::compute_df_dp(const Vector6d& stress,
+                                           Vector6d& df_dsigma_,
+                                           Vector6d& dp_dsigma_) {
+  const double ONETHIRDPI = 1.047197551;
```
kks32, line 111: needs more digits

kks32 (changes requested):

include/material/mohr_coulomb.h
srhgk2: Yes, let's do.

include/material/mohr_coulomb.h
srhgk2: We need three values for each plastic parameters, (psi, phi and c). Two constant values for peak and residual parameters. And another value which is the current value depending on the hardening model.

srhgk2: So, yes `cohesion_` is different from `c_`.

include/material/mohr_coulomb.tcc
srhgk2: got it.

include/material/mohr_coulomb.h
kks32: I have made that change.

kks32:
Newer version #363

PR #364, opened 2019-02-12
:bug: Plastic deviatoric strain should not be updated in a tensile failure

tianchiTJ:
Plastic deviatoric strain should not be updated in a tensile failure


kks32:
Updated in `mc_softening` branch

PR #386, opened 2019-03-18
Add function to read entity sets JSON file in IO

tianchiTJ:


include/io.h
```diff
@@ -11,6 +11,8 @@
 //! Alias for JSON
 #include "json.hpp"
 using Json = nlohmann::json;
+//! MPM Index
```
kks32: Please define it inside the namespace `mpm`.

include/io.h
```diff
@@ -51,6 +53,13 @@ class IO {
   //! Return json object
   Json json_object(const std::string& name) const { return json_[name]; }
 
+  //! Return the entity sets from the input set JSON file
+  //! \param[in] filename File name
+  //! \param[in] sets_type type of sets
+  //! \return map of entity sets
```
kks32:
```suggestion
  //! \retval entity_sets map of entity sets
```

src/io.cc
```diff
@@ -161,3 +161,43 @@ std::string mpm::IO::output_folder() const {
...
+  sfs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+  // Map of the entity sets
+  std::map<Index, std::vector<Index>> entity_sets;
+
+  try {
+    sfs.open(filename);
+    // Entity sets JSON object
+    Json json_sets_ = Json::parse(sfs);
+    // type of sets (ex: node sets or particle sets)
+    Json esets = json_sets_["sets_type"];
```
kks32: Could we rename `esets` to `sets`

kks32:
```suggestion
    Json sets = json_sets_[sets_type];
```

kks32: `sets_type` variable can't be inside a string ""

src/io.cc
```diff
@@ -161,3 +161,43 @@ std::string mpm::IO::output_folder() const {
   }
   return path;
 }
+
+//! Return map of entity sets from the JSON file
+std::map<Index, std::vector<Index>> mpm::IO::entity_sets(
+    const std::string& filename, const std::string& sets_type) {
+
+  // Input file stream for sets JSON file
+  std::ifstream sfs;
```
kks32: Could you rename `sfs` to `sets_file`

src/io.cc
```diff
@@ -161,3 +161,43 @@ std::string mpm::IO::output_folder() const {
...
+  sfs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+  // Map of the entity sets
+  std::map<Index, std::vector<Index>> entity_sets;
+
+  try {
+    sfs.open(filename);
+    // Entity sets JSON object
+    Json json_sets_ = Json::parse(sfs);
+    // type of sets (ex: node sets or particle sets)
```
kks32:
```suggestion
    // type of sets (e.g: node_sets or particle_sets)
```

src/io.cc
```diff
@@ -161,3 +161,43 @@ std::string mpm::IO::output_folder() const {
...
+    Json esets = json_sets_["sets_type"];
+
+    if (esets.size() > 0) {
+      for (Json::iterator itr = esets.begin(); itr != esets.end(); ++itr) {
+        // Get the entity set ids
+        Index id = (*itr)["id"].template get<Index>();
+        // Get the entity ids
+        std::vector<Index> eids;
```
kks32:
```suggestion
        std::vector<Index> entity_ids;
```

kks32: Rename to `entity_ids` or `entities`

src/io.cc
```diff
@@ -161,3 +161,43 @@ std::string mpm::IO::output_folder() const {
...
+        // Get the size of the set vector
+        eids.resize((*itr).at("set").size());
+
+        for (unsigned i = 0; i < (*itr).at("set").size(); ++i)
+          eids.emplace_back((*itr).at("set").at(i).template get<Index>());
+
+        entity_sets[id] = eids;
```
kks32: I thin map has an insert function, which returns a boolean, could we try to use that?

kks32 (changes requested):

kks32 (commented):
Excellent work! Just a few minor tweaks

src/io.cc
tianchiTJ: Sure

src/io.cc
kks32: http://www.cplusplus.com/reference/map/map/insert/

src/io.cc
tianchiTJ: Ok

src/io.cc
tianchiTJ: Ok, I will

src/io.cc
```diff
@@ -161,3 +161,46 @@ std::string mpm::IO::output_folder() const {
...
+        for (unsigned i = 0; i < (*itr).at("set").size(); ++i)
+          entity_ids.emplace_back(
+              (*itr).at("set").at(i).template get<mpm::Index>());
+        // Add the entity set to the list
+        entity_sets.insert(
+            std::pair<mpm::Index, std::vector<mpm::Index>>(id, entity_ids));
+      }
+    }
+    sets_file.close();
+  } catch (const std::exception& exception) {
```
kks32, line 197:
```suggestion
const std::out_of_range& range_error) {
        console_->warn("{} {} reading {}: {}", __FILE__, __LINE__, sets_type, filename,
                    range_error.what())
 }
catch (const std::exception& exception) {
  console_->error("{} #{}: {}\n", __FILE__, __LINE__, exception.what());
}
```

kks32 (changes requested):

src/io.cc
```diff
@@ -161,3 +161,46 @@ std::string mpm::IO::output_folder() const {
...
+    Json json_sets_ = Json::parse(sets_file);
+    // type of sets (e.g: node_sets or particle_sets)
+    Json sets = json_sets_[sets_type];
+
+    if (sets.size() > 0) {
+      for (Json::iterator itr = sets.begin(); itr != sets.end(); ++itr) {
+        // Get the entity set ids
+        mpm::Index id = (*itr)["id"].template get<mpm::Index>();
+        // Get the entity ids
+        std::vector<mpm::Index> entity_ids;
```
kks32:
```suggestion
 std::vector<mpm::Index> entity_ids = (*itr).at("set");
```

kks32: Lines  187 - 193 is not needed.

kks32 (approved):

PR #396, opened 2019-03-28
:computer: :dart: Pressure smoothing with MPI

kks32:
Add Ezra's implementation of pressure smoothing fixed MPI #324

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -1009,6 +1009,17 @@ void mpm::Cell<Tdim>::map_mass_momentum_to_nodes(
   }
 }
 
+//! Map particle pressure to nodes for a given phase
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::map_pressure_to_nodes(const Eigen::VectorXd& shapefn,
+                                            unsigned phase, double pmass,
+                                            double ppressure) {
+
+  for (unsigned i = 0; i < this->nfunctions(); ++i)
+    nodes_[i]->update_pressure(true, phase,
+                               shapefn(i) * pmass * shapefn(i) * ppressure);
+}
+
 //! Compute nodal momentum from particle mass and velocity for a given phase
 template <unsigned Tdim>
 void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
@@ -1136,6 +1147,17 @@ Eigen::Matrix<double, Tdim, 1> mpm::Cell<Tdim>::interpolate_nodal_acceleration(
     acceleration += shapefn(i) * nodes_[i]->acceleration(phase);
 
   return acceleration;
+}
+
+//! Return pressure at a point by interpolating from nodes
+template <unsigned Tdim>
+double mpm::Cell<Tdim>::interpolate_nodal_pressure(
+    const Eigen::VectorXd& shapefn, unsigned phase) {
+  double pressure = 0;
+  for (unsigned i = 0; i < this->nfunctions(); ++i)
+    pressure += shapefn(i) * nodes_[i]->pressure(phase);
+
+  return pressure;
 }
 
 //! Assign velocity constraint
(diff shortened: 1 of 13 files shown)


include/cell.tcc
```diff
@@ -1009,6 +1009,17 @@ void mpm::Cell<Tdim>::map_mass_momentum_to_nodes(
...
 
+//! Map particle pressure to nodes for a given phase
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::map_pressure_to_nodes(const Eigen::VectorXd& shapefn,
+                                            unsigned phase, double pmass,
+                                            double ppressure) {
+
+  for (unsigned i = 0; i < this->nfunctions(); ++i)
+    nodes_[i]->update_pressure(true, phase,
+                               shapefn(i) * pmass * shapefn(i) * ppressure);
```
srhgk2: Is `shapefn(i) * pmass * shapefn(i) * ppressure` correct? Why do we have two `shapefn(i)`?

include/particle.tcc
```diff
@@ -593,3 +593,48 @@ bool mpm::Particle<Tdim, Tnphases>::update_pressure(unsigned phase,
...
+bool mpm::Particle<Tdim, Tnphases>::map_pressure_to_nodes(unsigned phase) {
+  bool status = true;
+  try {
+    // Check if particle mass is set
+    if (mass_(phase) != std::numeric_limits<double>::max()) {
+      // Map particle mass and momentum to nodes
+      this->cell_->map_pressure_to_nodes(this->shapefn_, phase, mass_(phase),
+                                         pressure_(phase));
+    } else {
+      throw std::runtime_error("Particle mass has not be computed");
```
srhgk2: may be "has not been computed"?

include/particle.tcc
```diff
@@ -593,3 +593,48 @@ bool mpm::Particle<Tdim, Tnphases>::update_pressure(unsigned phase,
...
+
+// Compute pressure smoothing of the particle based on nodal pressure
+template <unsigned Tdim, unsigned Tnphases>
+bool mpm::Particle<Tdim, Tnphases>::compute_pressure_smoothing(unsigned phase) {
+  bool status = true;
+  try {
+    // Check if particle has a valid cell ptr
+    if (cell_ != nullptr) {
+      // Get interpolated nodal pressure
+      double pressure =
```
srhgk2: may be we don't need this copy.

tests/cell_test.cc
```diff
@@ -617,6 +619,11 @@ TEST_CASE("Cell is checked for 2D case", "[cell][2D]") {
         for (unsigned i = 0; i < pvelocity.size(); ++i)
           REQUIRE(node->momentum(phase)(i) == Approx(2.0).epsilon(Tolerance));
       }
+
+      // Map particle pressure to nodes
+      cell->map_pressure_to_nodes(shapefns_xi, phase, pmass, ppressure);
+      for (const auto& node : nodes)
+        REQUIRE(node->pressure(phase) == Approx(1.5).epsilon(Tolerance));
```
srhgk2: `Approx(1.5)` may change depending on how we map pressure to nodes

tests/cell_test.cc
```diff
@@ -1835,6 +1897,11 @@ TEST_CASE("Cell is checked for 3D case", "[cell][3D]") {
         for (unsigned i = 0; i < pvelocity.size(); ++i)
           REQUIRE(node->momentum(phase)(i) == Approx(1.0).epsilon(Tolerance));
       }
+
+      // Map particle pressure to nodes
+      cell->map_pressure_to_nodes(shapefns_xi, phase, pmass, ppressure);
+      for (const auto& node : nodes)
+        REQUIRE(node->pressure(phase) == Approx(0.75).epsilon(Tolerance));
```
srhgk2: same as 2d

include/cell.tcc
kks32: I was modifying Ezra's equations to do MPI. So I originally had `shapefn(i) * pmass * ppressure`, but that failed the test case. I think what this is trying to do is map particle mass and pressure separately to the nodes, then update the pressure. I need to check the original equation

kks32: Having looked at the equation, I think it should be `shapefn(i) * pmass * ppressure`

srhgk2 (approved):

PR #400, opened 2019-04-06
:computer: Fixes #399 point outside mesh

### LINKED ISSUE #399: Unable to locate particle within a cell
**Description**
There are cases when a particle should be within a cell, but the code does not capture it.

**Example**
The inclined landslide problem is forced to stop when a particle is deemed to be outside the mesh.
image

However, by visual inspection, it is determined that the particle should be within the mesh still.
image

A [unit test](https://github.com/ezrayst/mpm/blob/de93f540f46e58a433df2ac763e55b72dcb8d993/tests/cell_test.cc#L209) is done to prove it.

kks32:
Use a tolerance when determining local coordinates xi


kks32:
#418 supersedes this fix, this approach isn't best, as it uses tolerance

PR #403, opened 2019-04-11
:computer: Fixes #398 nodal mass bug on resume

### LINKED ISSUE #398: A nodel mass bug in resume computation
**Describe the bug**

**To Reproduce**

**Expected behavior**

kks32:
Fixes the nodal mass being zero error by resetting the ids of the cell for each particle. 

@z149181511 Could you please run your problem with this branch and see if you still get the zero nodal mass error?

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1,3 +1,4 @@
+
 // Constructor with id
 template <unsigned Tdim>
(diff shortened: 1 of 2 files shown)


kks32:
This does not fix the issue!

PR #410, opened 2019-04-24
Remove particles at a specific step to simulate excavation 

tianchiTJ:
input file change:
add "remove" to "particle_sets"
```
"particle": {
        "material_id": 0,
        "particle_type": "P2D",
        "particle_sets": [
        {
          "set_id": 0,
          "material_id": 0,
          "remove": true,
          "rstep":10
        },
        {
          "set_id":1,
          "material_id": 1,
          "remove": true,
          "rstep":10
        },
        {
          "set_id":2,
          "material_id": 1,
          "remove": true,
          "rstep":20
        }
        ]
    },
```


include/mesh.h
```diff
@@ -333,6 +344,8 @@ class Mesh {
   Container<ParticleBase<Tdim>> particles_;
   //! Container of particle sets
   tsl::robin_map<unsigned, Container<ParticleBase<Tdim>>> particle_sets_;
+  //! Container of remove steps
+  tsl::robin_map<mpm::Index, std::vector<unsigned>> remove_steps_;
```
srhgk2: Let's remove this map and use the existing `particle_sets_`

include/mesh.tcc
```diff
@@ -1135,3 +1135,33 @@ bool mpm::Mesh<Tdim>::create_particle_sets(
   }
   return status;
 }
+
+//! Create map of remove steps
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::create_remove_step(const mpm::Index rstep,
```
srhgk2, line 1143: Instead of this function, Let's read all features specific to particle sets.

include/mpm_explicit.tcc
```diff
@@ -682,6 +682,20 @@ bool mpm::MPMExplicit<Tdim>::solve() {
   // Check point resume
   if (resume) this->checkpoint_resume();
 
+  // Create remove steps (excavation)
+  if (!particle_props["remove"].empty()) {
```
srhgk2: We will add `remove` as a bool and unsigned `remove_step` inside the existing particle sets.

srhgk2 (changes requested):

include/mesh.h
kks32: I think the mpm::Index is related to the time step, so we should have this in the explicit algorithm rather than the mesh.

kks32: mesh doesn't have a sense of time

kks32 (changes requested):

include/mesh.h
tianchiTJ: Yes, the key is the number of remove step, and the vector is the ids of particle sets which need to be removed. Do you mean store this map in the mpm_explicit straightly?

tianchiTJ:
If remove steps are defined, although the computation can run, there will be error message come out :" Nodal mass is zero or below threshold". I think it's similar with the problem in resume.

tianchiTJ:
I have update the "remove_particles" branch to the new format.
By the way, I also add a part to remove the particle from the cell, which fixes the bug that the error message of  "Nodal mass is zero or below threshold " will come out using the original code.

include/mpm_base.tcc
```diff
@@ -0,0 +1,628 @@
...
+    nodal_tractions_ = false;
+  }
+  return status;
+}
+
+//! Apply properties to particles sets (e.g: material, remove step)
+template <unsigned Tdim>
+bool mpm::MPMBase<Tdim>::apply_properties_to_particles_sets() {
+  bool status = false;
+  // Assign material to particle sets
```
srhgk2: Remove `// Assign material to particle sets` comment from here.

include/mpm_base.tcc
```diff
@@ -0,0 +1,628 @@
...
+    auto particle_props = io_->json_object("particle");
+    // Get particle sets properties
+    auto particle_sets = particle_props["particle_sets"];
+    // Iterate over each particle sets
+    for (const auto& psets : particle_sets) {
+      // Assign material to particle sets
+      // Get set material from list of materials
+      auto set_material = materials_.at(psets["material_id"]);
+      // Get sets ids
+      std::vector<unsigned> sids = psets["set_id"];
```
srhgk2: Let's have each particle set defined separately in the input.json file so it does not complicate when more properties are defined.
i.e. `set_id` is one  value, not a vector.

include/mpm_base.tcc
```diff
@@ -0,0 +1,628 @@
...
+                            std::placeholders::_1, set_material));
+      }
+
+      // Assign remove steps to particle sets
+      // Check if remove step is needed
+      if (psets["remove"] == true) {
+        // Get the number of remove step
+        mpm::Index rstep = psets["rstep"];
+        // Get ids of sets need to be removed
+        std::vector<unsigned> sids = psets["set_id"];
```
srhgk2: Remove get ids from here as it is already implemented above. For each `particle_set`, read properties. If `remove` is set to true, call `mesh_->create_remove_step(rstep, sid);`

include/mesh.tcc
```diff
@@ -1135,3 +1135,42 @@ bool mpm::Mesh<Tdim>::create_particle_sets(
   }
   return status;
 }
+
+//! Create map of container of remove steps
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::create_remove_step(const mpm::Index rstep,
+                                         const std::vector<unsigned> set_ids) {
+  // Get or create the remove step existing at "rstep"
+  std::vector<unsigned> sids = remove_steps_[rstep];
```
srhgk2, line 1146: Now, the `sids` is not a vector, but a single value. Change this.

srhgk2 (changes requested):

include/mesh.tcc
```diff
@@ -1135,3 +1135,42 @@ bool mpm::Mesh<Tdim>::create_particle_sets(
...
+bool mpm::Mesh<Tdim>::apply_remove_step(const mpm::Index rstep) {
+  bool status = false;
+  if (remove_steps_.find(rstep) != remove_steps_.end()) {
+    try {
+      // Iterate over each particle sets in the remove step
+      for (auto sid : remove_steps_.at(rstep)) {
+        // Iterate over each  particles in the set
+        for (auto particle = particle_sets_.at(sid).cbegin();
+             particle != particle_sets_.at(sid).cend(); particle++) {
+          // Remove particle from the cell
```
kks32: This is no longer needed.

kks32:
Will create a separate issue and I will work on this. I'm working on changing the particle structure, I'll add `staged_construction` to analysis type. Thanks for this!

Issue #411, opened 2019-04-25
Unable to add specified vtk attribute output from input file


ezrayst:
*Description*
Adding specified vtk attribute outputs (ie. "epds") does not work. The code makes sure to only write the output attribute files for the defaults: stresses, strains and velocities (or less).

*Source of bug*
This line makes sure only the defaults could be outputted
https://github.com/cb-geo/mpm/blob/73d816235e1d309d0458817d6d126c3d21ecc353/include/mpm_explicit.tcc#L77

*Proposed solution*
`emplace_back` whether the additional attribute is found in the default
This is ok because of the check we have in `mesh.tcc`
https://github.com/cb-geo/mpm/blob/73d816235e1d309d0458817d6d126c3d21ecc353/include/mesh.tcc#L455

kks32:
This part of printing `epds` is not part of the mpm code, so can't be treated as a bug at this time.

PR #412, opened 2019-04-25
Add vtk attribute accordingly #411

ezrayst:
Delete the condition that specified attributes have to be found within the defaults


kks32:
This is not the best way to fix this

PR #413, opened 2019-04-28
Modify the MC model code

tianchiTJ:
Eliminate some unnecessary computation when it's in elastic
Set a tolerance for the final check


include/material/mohr_coulomb.tcc
```diff
@@ -471,49 +469,49 @@ Eigen::Matrix<double, 6, 1> mpm::MohrCoulomb<Tdim>::compute_stress(
...
 
-  // compute the correction stress
+  // Compute stress invariants based on stress input
+  this->compute_stress_invariants(stress, state_vars);
+  // Compute yield function based on stress input
+  Eigen::Matrix<double, 2, 1> yield_function;
+  auto yield_type = this->compute_yield_state(&yield_function, state_vars);
+  // Compute plastic multiplier based on stress input (Lambda)
+  double softening = 0.;
+  Vector6d df_dsigma = Vector6d::Zero();
```
kks32, line 507: Is this Eigen::Vector6d? Could you please use just `auto` as you are using `Vector6d::Zero()`

include/material/mohr_coulomb.tcc
```diff
@@ -546,18 +544,23 @@ Eigen::Matrix<double, 6, 1> mpm::MohrCoulomb<Tdim>::compute_stress(
   // Check if the current stress state is outside the yield surface
   this->compute_stress_invariants(updated_stress, state_vars);
   yield_type = this->compute_yield_state(&yield_function, state_vars);
-  if (epds > epds_peak_ && yield_type != FailureState::Elastic) {
+
+  // Yield function tolerance
+  double Tolerance = 1.E-10;
```
kks32: const

kks32 (changes requested):

include/material/mohr_coulomb.tcc
z149181511: If I use auto, it will be defined as  Eigen::Matrix<double, 6, 1> but not Vector6D. Although  Vector6D is just  Eigen::Matrix<double, 6, 1>, there will be treated as different types in the arguments of functions. If you want the "auto", I need to change all the arguments types of functions from Vector6D to Eigen::Matrix<double, 6, 1>.

tianchiTJ: There will be an error happen when using auto, because it will set the value as "Eigen::Matrix<double, 6, 1>" but not "Vector6d". Although they are the same one, they are treated as two different types arguments in functions

kks32 (approved):

include/material/mohr_coulomb.tcc
kks32: Could you please use `Eigen::Vector6d df_dsigma = Eigen::Vector6d::Zero()`

kks32 (approved):

PR #429, opened 2019-05-30
Fix #398 nodal mass bug  

### LINKED ISSUE #398: A nodel mass bug in resume computation

ezrayst:
The problem arises from the double initialization of the `particles`. Even though the `particles` have the right `cell_id`, the `cell` does not have the right `particles`.

Proposed solution: in resume, remove each particle from the cell it belongs to.


kks32:
#430 supersedes this

PR #430, opened 2019-05-30
:hammer: Fix #398  nodal mass error on resume

### LINKED ISSUE #398: A nodel mass bug in resume computation

kks32:
Clear all particle ids in the cell on resume


ezrayst:
This is more general and it does solve the problem as well!

ezrayst (approved):

cbgeo (approved):

PR #446, opened 2019-07-02
Bbar after Hughes (1980)

ezrayst:
Implementation of Bbar to compute strain rate from velocity field after
[Hughes (1980)](https://doi.org/10.1002/nme.1620150914)


include/cell.h
```diff
@@ -242,11 +242,18 @@ class Cell {
...
+  //! \param[in] bmatrix Bmatrix corresponding to local coordinates of particle
+  //! \param[in] phase Phase associate to the particle
+  std::vector<Eigen::MatrixXd> compute_bbar(
+      const std::vector<Eigen::MatrixXd>& bmatrix, unsigned phase);
+
   //! Compute strain rate
   //! \param[in] bmatrix Bmatrix corresponding to local coordinates of particle
   //! \param[in] phase Phase associate to the particle
+  //! \param[in] bbar Option to compute strain using bbar
   Eigen::VectorXd compute_strain_rate(
```
kks32, line 254: I would do a separate `compute_strain_rate_bbar`

include/cell.tcc
```diff
@@ -1236,10 +1236,58 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
 
+//! Compute Bbar matrix
+template <unsigned Tdim>
+std::vector<Eigen::MatrixXd> mpm::Cell<Tdim>::compute_bbar(
+    const std::vector<Eigen::MatrixXd>& bmatrix, unsigned phase) {
+
+  // Compute volumetric Bmatrix
+  std::vector<Eigen::MatrixXd> bmatrix_vol;
+  for (unsigned i = 0; i < this->nnodes(); ++i) {
+    Eigen::Matrix<double, 3 * (Tdim - 1), Tdim> bvoli;
```
kks32: Replace this with a computed final `bbar` matrix, that is just iterated over nodes and appended. This would make the calculation quicker.

Use template specialization to split between 2D and 3D version

tests/cell_test.cc
```diff
@@ -1,3 +1,4 @@
+#include <iostream>
```
kks32: Remove this

kks32 (changes requested):
Could you replace the longer calculation of BBar with the final value of BBar matrix using `bmatrix_` and `bmatrix_centroid_` directly? So instead of calculating, you just assign the BBar matrix from the bmatrix and the centroid

include/cell.tcc
```diff
@@ -1236,6 +1236,27 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
+//! Compute Bbar matrix
+template <unsigned Tdim>
+std::vector<Eigen::MatrixXd> mpm::Cell<Tdim>::compute_bbar(
+    const std::vector<Eigen::MatrixXd>& bmatrix, unsigned phase) {
+
+  // Compute BBar
+  std::vector<Eigen::MatrixXd> bbar_matrix;
+  for (unsigned i = 0; i < this->nnodes(); ++i) {
+    Eigen::Matrix<double, 3 * (Tdim - 1), Tdim> bbari;
+    for (unsigned j = 0; j < Tdim; ++j) {
```
kks32, line 1250: The `bbari` has a size of 6x3 for 3D, and 3x2 for 2D but you are only iterating to maximum of Tdim, which isn't correct. Could you please fix that?

include/cell.tcc
```diff
@@ -1236,6 +1236,27 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
+      for (unsigned k = 0; k < Tdim; ++k)
+        bbari(j, k) =
+            bmatrix.at(i)(j, k) +
+            1. / 3. * (bmatrix_centroid_.at(i)(k, k) - bmatrix.at(i)(k, k));
```
kks32: Define `OneThird` as a `const double` outside the loop to be `1.0/3.0`

include/particle.h
```diff
@@ -145,7 +145,8 @@ class Particle : public ParticleBase<Tdim> {
   //! Compute strain
   //! \param[in] phase Index corresponding to the phase
   //! \param[in] dt Analysis time step
-  void compute_strain(unsigned phase, double dt) override;
+  //! \param[in] bbar Bmatrix (true: bbar, false: normal B matrix)
+  void compute_strain(unsigned phase, double dt, bool bbar) override;
```
kks32: I'd create a separate `compute_strain_bbar`

kks32 (changes requested):
Please check your BBar matrix, it is not right

include/particle.tcc
```diff
@@ -389,9 +389,12 @@ bool mpm::Particle<Tdim, Tnphases>::map_mass_momentum_to_nodes(unsigned phase) {
 
 // Compute strain of the particle
 template <unsigned Tdim, unsigned Tnphases>
-void mpm::Particle<Tdim, Tnphases>::compute_strain(unsigned phase, double dt) {
+void mpm::Particle<Tdim, Tnphases>::compute_strain(unsigned phase, double dt,
+                                                   bool bbar) {
   // Strain rate
-  const auto strain_rate = cell_->compute_strain_rate(bmatrix_, phase);
+  auto strain_rate = cell_->compute_strain_rate(bmatrix_, phase);
+  if (bbar) strain_rate = cell_->compute_strain_rate_bbar(bmatrix_, phase);
```
ezrayst: I cannot declare just within if-statements. This is my current solution - any feedback is appreciated

include/cell.tcc
ezrayst: I think you misread it. The size is 6x3 for 3D `(3 * (Tdim - 1), Tdim)`, and 3x2 for 2D

kks32: It's not an issue of size, that was just a typo in my comment, the issue is the bbar matrix is not filled correctly

include/cell.tcc
ezrayst: Understood

include/cell.tcc
ezrayst: I see. I fixed it. Meant to fill it with bmatrix first, then change only the Tdim part because bbar does not interfere with the shear part.

ezrayst:
Pending work:
- [ ] write test cases for `compute_strain_bbar` in `particle`

include/cell.tcc
```diff
@@ -1236,6 +1236,29 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
+template <unsigned Tdim>
+std::vector<Eigen::MatrixXd> mpm::Cell<Tdim>::compute_bbar(
+    const std::vector<Eigen::MatrixXd>& bmatrix, unsigned phase) {
+
+  const double OneThird = 1.0 / 3.0;
+
+  // Compute BBar
+  std::vector<Eigen::MatrixXd> bbar_matrix;
+  for (unsigned i = 0; i < this->nnodes(); ++i) {
+    Eigen::Matrix<double, 3 * (Tdim - 1), Tdim> bbari = bmatrix.at(i);
```
kks32: use `auto`

include/cell.tcc
```diff
@@ -1236,6 +1236,29 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
+    for (unsigned j = 0; j < Tdim; ++j) {
+      for (unsigned k = 0; k < Tdim; ++k)
+        bbari(j, k) =
+            bmatrix.at(i)(j, k) +
```
kks32: you can do `bbari(j, k) += OneThird * (bmatrix_centroid_.at(i)(k, k) - bmatrix.at(i)(k, k));`

include/cell.tcc
```diff
@@ -1236,6 +1236,29 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
+            OneThird * (bmatrix_centroid_.at(i)(k, k) - bmatrix.at(i)(k, k));
+    }
+    bbar_matrix.push_back(bbari);
```
kks32: use `emplace_back` instead

include/cell.tcc
```diff
@@ -1236,6 +1236,29 @@ void mpm::Cell<Tdim>::compute_nodal_momentum(const Eigen::VectorXd& shapefn,
...
 
+//! Compute Bbar matrix
+template <unsigned Tdim>
+std::vector<Eigen::MatrixXd> mpm::Cell<Tdim>::compute_bbar(
+    const std::vector<Eigen::MatrixXd>& bmatrix, unsigned phase) {
+
+  const double OneThird = 1.0 / 3.0;
+
+  // Compute BBar
+  std::vector<Eigen::MatrixXd> bbar_matrix;
```
kks32: rename to `bbar`

include/particle.tcc
```diff
@@ -392,6 +392,65 @@ template <unsigned Tdim, unsigned Tnphases>
 void mpm::Particle<Tdim, Tnphases>::compute_strain(unsigned phase, double dt) {
```
kks32: OK sorry to go back on this, there is quite a lot of repetition of code here, so let's pass `bbar` as an argument and switch between the `strain_rate`

kks32 (changes requested):

include/particle.tcc
ezrayst: I agree!

include/particle.tcc
ezrayst: @kks32, what do you think about this?

kks32: Declare the type and initalize based on which one is being used

kks32 (changes requested):

include/particle.h
kks32: could you default this to `false`, so if we don't pass, it's false

kks32: so you don't have to change the old test cases

kks32:
We no longer store bmatrix in particle or cell, so this would have to be rethought through. Will think about pressure smoothing and how best to incorporate this.

PR #451, opened 2019-07-18
Graph partitioning with ParMETIS

ryanwentaoxu:
The following are added by Wentao
file added:
include/graph.tcc
include/graph.h

I modified a little in CMakefile.txt


kks32 (changes requested):
Could you update it so that ParMETIS is run if and only if both MPI is found and also when the number of cells is > 1?

ryanwentaoxu:
> Could you update it so that ParMETIS is run if and only if both MPI is found and also when the number of cells is > 1?

OK! Sorry I just saw this message.

PR #454, opened 2019-08-05
Initial condition for newton-raphson

ezrayst:
- [x] Affine transformation might give initial condition of newton-raphson outside of [-1, 1], thus creating problems for particles near the edge of an element which might result in failure of finding the natural coordinates of the particle. Thus, initial condition is kept [-1, 1].
- [x] Stress computation from material is checked for NaN (not a number) which usually happens in case of zero divided by zero.

### DIFF
--- a/include/cell.h
+++ b/include/cell.h
@@ -8,6 +8,8 @@
 #include <set>
 #include <vector>
 
+#include <math.h>
+
 #include "Eigen/Dense"
 #include "Eigen/LU"
 
--- a/include/cell.h
+++ b/include/cell.h
@@ -8,8 +8,6 @@
 #include <set>
 #include <vector>
 
-#include <math.h>
-
 #include "Eigen/Dense"
 #include "Eigen/LU"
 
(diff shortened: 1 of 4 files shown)


include/particle.tcc
```diff
@@ -452,8 +452,23 @@ bool mpm::Particle<Tdim, Tnphases>::compute_stress(unsigned phase) {
     if (material_.at(phase) != nullptr) {
       Eigen::Matrix<double, 6, 1> dstrain = this->dstrain_.col(phase);
       // Calculate stress
-      this->stress_.col(phase) = material_.at(phase)->compute_stress(
+      auto stress = material_.at(phase)->compute_stress(
           this->stress_.col(phase), dstrain, this, &state_variables_);
+
```
kks32: This is an expensive check to do every time, why is this happening?

include/cell.h
```diff
@@ -8,6 +8,8 @@
 #include <set>
 #include <vector>
 
+#include <math.h>
```
kks32:
```suggestion
#include <cmath>
```

kks32 (changes requested):

include/particle.tcc
ezrayst: Materials can give NaN stresses. I think I can omit it for this PR, but in MC model, we can have zero divided by zero case, which will give NaN stress, thus NaN acceleration, velocity and location.

kks32:
On new branch `nr`

PR #456, opened 2019-08-15
Triangle element class and unit tests

thiagordonho:
Implement the `TriangleElement` class that includes:

- 3-noded triangular element.
- 6-noded triangular element.

And the quadrature rules for such elements (`TriangleQuadrature `class):

- 1 quadrature point.
- 3 quadrature points.

The unit tests for these classes are also implemented.

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -243,12 +243,13 @@ inline void mpm::Cell<2>::compute_volume() {
       auto node1 = nodes_[indices(1)]->coordinates();
       auto node2 = nodes_[indices(2)]->coordinates();
 
-      // 2 * Area = (x1 * y2 - x2 * y1) - (x0 * y2 - x2 * y0) + (x0 * y1 - x1 *
-      // y0)
+      // Area = 0.5 * [ (x1 * y2 - x2 * y1)
+      //              - (x0 * y2 - x2 * y0)
+      //              + (x0 * y1 - x1 * y0) ]
       volume_ = std::fabs(((node1(0) * node2(1)) - (node2(0) * node1(1))) -
                           ((node0(0) * node2(1)) - (node2(0) * node0(1))) +
-                          ((node0(0) * node1(1)) - (node1(0) * node0(1)))) /
-                2.0;
+                          ((node0(0) * node1(1)) - (node1(0) * node0(1)))) *
+                0.5;
       // Quadrilateral
     } else if (indices.size() == 4) {
 
@@ -552,11 +553,11 @@ inline Eigen::Matrix<double, 2, 1> mpm::Cell<2>::local_coordinates_point(
                            (node2(0) - node0(0)) * (node1(1) - node0(1))) /
                           2.0;
 
-      xi(0) = 1 / (2 * area) *
+      xi(0) = 1. / (2. * area) *
               ((point(0) - node0(0)) * (node2(1) - node0(1)) -
                (node2(0) - node0(0)) * (point(1) - node0(1)));
 
-      xi(1) = -1 / (2 * area) *
+      xi(1) = -1. / (2. * area) *
               ((point(0) - node0(0)) * (node1(1) - node0(1)) -
                (node1(0) - node0(0)) * (point(1) - node0(1)));
       // Quadrilateral
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -472,13 +472,12 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
   // Check if the transformed coordinate is within the unit cell:
   // between 0 and 1-xi(1-i) if the element is a triangle, and between
   // -1 and 1 if otherwise
-  switch (this->element_->corner_indices().size()) {
-    case 3 : for (unsigned i = 0; i < (*xi).size(); ++i)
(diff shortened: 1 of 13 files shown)


include/cell.tcc
```diff
@@ -450,9 +469,17 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
   else
     (*xi) = this->transform_real_to_unit_cell(point);
 
-  // Check if the transformed coordinate is within the unit cell (-1, 1)
-  for (unsigned i = 0; i < (*xi).size(); ++i)
-    if ((*xi)(i) < -1. || (*xi)(i) > 1. || std::isnan((*xi)(i))) status = false;
+  // Check if the transformed coordinate is within the unit cell:
+  // between 0 and 1-xi(1-i) if the element is a triangle, and between
+  // -1 and 1 if otherwise
+  switch (this->element_->corner_indices().size()) {
```
kks32: Why can't you have an `if else`?

kks32 (changes requested):

include/cell.tcc
thiagordonho: I thought about it. But if in the future someone wants to implement another type of element than this structure would be more helpful. Should I still change it to if-else?

include/cell.tcc
```diff
@@ -223,13 +223,34 @@ inline void mpm::Cell<1>::compute_volume() {
...
+      //     |        `\
+      //   0 0----------0 1
+      //
+      auto node0 = nodes_[indices(0)]->coordinates();
+      auto node1 = nodes_[indices(1)]->coordinates();
+      auto node2 = nodes_[indices(2)]->coordinates();
+
+      // 2 * Area = (x1 * y2 - x2 * y1) - (x0 * y2 - x2 * y0) + (x0 * y1 - x1 *
+      // y0)
+      volume_ = std::fabs(((node1(0) * node2(1)) - (node2(0) * node1(1))) -
```
kks32, line 249: This is confusing, if we are returning the `volume`, we should return the `area` and not `2 * area`

include/cell.tcc
```diff
@@ -450,9 +471,18 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
   else
     (*xi) = this->transform_real_to_unit_cell(point);
 
-  // Check if the transformed coordinate is within the unit cell (-1, 1)
-  for (unsigned i = 0; i < (*xi).size(); ++i)
-    if ((*xi)(i) < -1. || (*xi)(i) > 1. || std::isnan((*xi)(i))) status = false;
+  // Check if the transformed coordinate is within the unit cell:
+  // between 0 and 1-xi(1-i) if the element is a triangle, and between
+  // -1 and 1 if otherwise
+  if (this->element_->corner_indices().size() == 3) {
```
kks32, line 478:
```
double min = -1., max = 1.;
if (this->element_->corner_indices().size() == 3) {
  min = 0;
  max = 1.;
}
for (unsigned i = 0; i < (*xi).size(); ++i)
  if ((*xi)(i) < min || (*xi)(i) > max || std::isnan((*xi)(i)))
    status = false;
```

kks32: I know this doesn't test the inclined side, but we need a better approach than too many if else conditions

include/cell.tcc
```diff
@@ -502,8 +532,35 @@ inline Eigen::Matrix<double, 2, 1> mpm::Cell<2>::local_coordinates_point(
...
+
+      auto node0 = nodes_[indices(0)]->coordinates();
+      auto node1 = nodes_[indices(1)]->coordinates();
+      auto node2 = nodes_[indices(2)]->coordinates();
+
+      const double area = ((node1(0) - node0(0)) * (node2(1) - node0(1)) -
+                           (node2(0) - node0(0)) * (node1(1) - node0(1))) /
+                          2.0;
+
+      xi(0) = 1 / (2 * area) *
```
kks32: Use `1./ (2. * area)` when using floating points

include/triangle_element.tcc
```diff
@@ -0,0 +1,379 @@
...
+    const Eigen::Matrix<double, 2, 1>& particle_size,
+    const Eigen::Matrix<double, 2, 1>& deformation_gradient) const {
+  Eigen::Matrix<double, 3, 1> shapefn;
+  shapefn(0) = 1 - (xi(0) + xi(1));
+  shapefn(1) = xi(0);
+  shapefn(2) = xi(1);
+  return shapefn;
+}
+
+//! Return gradient of shape functions of a 4-node Triangle Element at a
```
kks32:
```suggestion
//! Return gradient of shape functions of a 3-node Triangle Element at a
```

include/cell.tcc
kks32: Alternatively, we call a function in `element` class to see if a point is in it's unit cell, that would be a lot nicer. I'd go with that. If you wish that could be a separate PR improving this. So leave this section as is, and then let's do a separate PR improving this.

include/cell.tcc
kks32: Where are you using this?

kks32 (changes requested):

kks32 (approved):

PR #471, opened 2019-10-10
Unit test for local coordinates calculation of triangular element

thiagordonho:
An additional test implemented to the point_in_cell_test.cc file testing the derivation of local coordinates for triangular elements using an analytical solution. The analytical solution is based on barycentric coordinates in a triangle.

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -510,6 +510,7 @@ inline Eigen::Matrix<double, 3, 1> mpm::Cell<3>::local_coordinates_point(
   return xi;
 }
 
+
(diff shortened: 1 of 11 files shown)


include/cell.tcc
```diff
@@ -510,6 +510,7 @@ inline Eigen::Matrix<double, 3, 1> mpm::Cell<3>::local_coordinates_point(
   return xi;
 }
 
+
```
kks32: Remove this extra line

kks32 (changes requested):
Fix clang-format errors

kks32 (approved):

PR #472, opened 2019-10-14
Displacement vtp output and fix typo

ezrayst:
- [x] Add a variable `original_coordinates_` in order to record displacement in reference to the original location of particles
- [x] Add `displacement` output 
- [ ] Test cases

PR #476, opened 2019-10-21
:computer: :art: Output vector data of particles for viz

kks32:
This should define displacements as a new particle vector field, overrides #472

### DIFF
--- a/include/particle.tcc
+++ b/include/particle.tcc
@@ -82,12 +82,13 @@ bool mpm::Particle<Tdim, Tnphases>::initialise_particle(
 // Initialise particle properties
 template <unsigned Tdim, unsigned Tnphases>
 void mpm::Particle<Tdim, Tnphases>::initialise() {
+  displacement_.setZero();
   dstrain_.setZero();
   mass_.setZero();
+  natural_size_.setZero();
   pressure_.setZero();
   set_traction_ = false;
   size_.setZero();
-  natural_size_.setZero();
   strain_rate_.setZero();
   strain_.setZero();
   stress_.setZero();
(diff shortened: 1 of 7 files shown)


include/particle.tcc
```diff
@@ -575,6 +576,8 @@ bool mpm::Particle<Tdim, Tnphases>::compute_updated_position(unsigned phase,
 
       // New position  current position + velocity * dt
       this->coordinates_ += nodal_velocity * dt;
+      // Update displacement (displacement is initialized from zero)
+      this->displacement_ += nodal_velocity * dt;
```
ezrayst, line 590: I like this!

include/particle.tcc
```diff
@@ -719,3 +724,20 @@ void mpm::Particle<Tdim, Tnphases>::apply_particle_velocity_constraints() {
     this->velocity_(direction, phase) = constraint.second;
   }
 }
+
+//! Apply particle velocity constraints
```
ezrayst: This comment does not fit the function

include/particle.tcc
```diff
@@ -719,3 +724,20 @@ void mpm::Particle<Tdim, Tnphases>::apply_particle_velocity_constraints() {
...
+template <unsigned Tdim, unsigned Tnphases>
+Eigen::VectorXd mpm::Particle<Tdim, Tnphases>::vector_data(
+    unsigned phase, const std::string& property) {
+  //! Map of properties
+  std::map<std::string, std::function<Eigen::VectorXd(unsigned)>> properties;
```
ezrayst: I wonder whether we should make `map` every time we call `vector_data`. My thoughts would be to have the map as private variables.

ezrayst (commented):
Like the general structure!
When one day we output `state variables`, how should we expand this?

include/particle.tcc
kks32: the problem is we can't pass phase otherwise? I'm not quite sure about this.

ezrayst: I see. How about a map that has two keys (attribute and phase), like a tuple? We can define the phase up to 3 only right, not infinite.

I still feel making map for every particle and every time step we output is a bit inefficient.

ezrayst (approved):

PR #481, opened 2019-10-28
:hammer: Fixes #480 material property type

### LINKED ISSUE #480: Material property test for bool
Function `property` in `material.h` is only outputting `double` material properties.
It does not work when we have a `bool` material property such as `softening` in Mohr-Coulomb and what I am developing with bonded NorSand.

[Code Line](https://github.com/cb-geo/mpm/blob/164a6665a2b791f893e0e5492c44625acda36de7/include/material/material.h#L59)

Suggestion:
Overloading the function with `bool`?

kks32:
call property as `material->template property<type>(key)`

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -511,14 +511,20 @@ inline Eigen::Matrix<double, 3, 1> mpm::Cell<3>::local_coordinates_point(
 }
 
 //! Return the local coordinates of a point in a cell
+template <>
+inline Eigen::Matrix<double, 1, 1> mpm::Cell<1>::transform_real_to_unit_cell(
+    const Eigen::Matrix<double, 1, 1>& point) {
+  return this->local_coordinates_point(point);
+}
+
+//! Return the local coordinates of a point in a cell
 template <unsigned Tdim>
 inline Eigen::Matrix<double, Tdim, 1>
     mpm::Cell<Tdim>::transform_real_to_unit_cell(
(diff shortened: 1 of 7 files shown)


include/cell.tcc
```diff
@@ -510,15 +510,21 @@ inline Eigen::Matrix<double, 3, 1> mpm::Cell<3>::local_coordinates_point(
   return xi;
 }
 
+//! Return the local coordinates of a point in a 1D cell
+template <>
+inline Eigen::Matrix<double, 1, 1> mpm::Cell<1>::transform_real_to_unit_cell(
```
ezrayst, line 515: Why do we have this function here? is it from #479 that has not been accounted for?

tests/material/bingham_test.cc
```diff
@@ -69,11 +69,11 @@ TEST_CASE("Bingham is checked in 2D", "[material][bingham][2D]") {
     REQUIRE(material->id() == 0);
 
     // Get material properties
-    REQUIRE(material->property("density") ==
+    REQUIRE(material->template property<double>("density") ==
```
ezrayst, line 72: Great!

include/cell.tcc
kks32: clang failed, at least locally on my machine

ezrayst: Understood, it's fine then!

ezrayst (approved):

ezrayst:
Codecov decreases though :(

PR #487, opened 2019-10-31
Fix minor errors in cmake after metis

bodhinandach:
@kks32 this solves my compilation problem with METIS, feel free to modify if you find a better solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,7 +93,7 @@ endif()
 # METIS
 find_package(METIS)
 if (METIS_FOUND)
-    include_directories(${METIS_INCLUDE_DIR})
+    include_directories(SYSTEM ${METIS_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -91,9 +91,9 @@ if (VTK_FOUND)
 endif()
 
 # METIS
-find_package(METIS REQUIRED)
+find_package(METIS)
 if (METIS_FOUND)
-    include_directories(SYSTEM ${METIS_INCLUDE_DIR})
```
kks32, line 96: Does it work without removing SYSTEM?

kks32 (changes requested):

CMakeLists.txt
bodhinandach: It works, I'll revert.

kks32 (approved):

PR #492, opened 2019-11-07
Change iterating through particles to iterating through cells

ryanwentaoxu:
remove cells by iterating through cells instead of iterating through particles

### DIFF
--- a/include/particle.tcc
+++ b/include/particle.tcc
@@ -208,7 +208,13 @@ bool mpm::Particle<Tdim, Tnphases>::assign_cell_id(mpm::Index id) {
 template <unsigned Tdim, unsigned Tnphases>
 void mpm::Particle<Tdim, Tnphases>::remove_cell() {
   // if a cell is not nullptr
-  // if (cell_ != nullptr) cell_->remove_particle_id(this->id_);
+  if (cell_ != nullptr) cell_->remove_particle_id(this->id_);
+  cell_id_ = std::numeric_limits<Index>::max();
+}
+
+// set cell id after remove cell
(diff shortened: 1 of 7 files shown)


include/particle.tcc
```diff
@@ -208,7 +208,7 @@ bool mpm::Particle<Tdim, Tnphases>::assign_cell_id(mpm::Index id) {
 template <unsigned Tdim, unsigned Tnphases>
 void mpm::Particle<Tdim, Tnphases>::remove_cell() {
   // if a cell is not nullptr
-  if (cell_ != nullptr) cell_->remove_particle_id(this->id_);
+  // if (cell_ != nullptr) cell_->remove_particle_id(this->id_);
```
kks32: why is this commented out?

CassiniXu: When iterate through cells, one needs to iterate through local paticles_ids in that cells and then delete them, but when calling this function once, the particles_ids would be deleted and the iteration process would be stopped. So I tried to comment it out.

CassiniXu: We can write another similar function which comment this out and the original one can remain the same.

ryanwentaoxu: > why is this commented out?

When iterate through cells, one needs to iterate through local paticles_ids in that cells and then delete them, but when calling this function once, the particles_ids would be deleted and the iteration process would be stopped. So I tried to comment it out.

kks32: but this is just removing id

ryanwentaoxu: > why is this commented out?

We can write another similar function which comment this out and the original one can remain the same.

PR #494, opened 2019-11-08
Norsand

ezrayst:
- [x] Traditional Norsand
- [x] Norsand bonded model
- [x] Test cases for 3D, including yielded bonded model


CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ if(MPM_BUILD_TESTING)
     ${mpm_SOURCE_DIR}/tests/material/bingham_test.cc
     ${mpm_SOURCE_DIR}/tests/material/linear_elastic_test.cc
     ${mpm_SOURCE_DIR}/tests/material/newtonian_test.cc
+    ${mpm_SOURCE_DIR}/tests/material/nor_sand_test.cc
```
kks32: Could you rename it to `norsand_test` and the rest of it to `norsand`

ezrayst: Will do!

include/material/norsand.tcc
```diff
@@ -0,0 +1,589 @@
...
+        material_properties["p_image_initial"].template get<double>();
+    // Flag for bonded model
+    bond_model_ = material_properties["bond_model"].template get<bool>();
+    // Initial p_cohesion
+    p_cohesion_initial_ =
+        material_properties["p_cohesion_initial"].template get<double>();
+    // Degradation parameter m upon shearing
+    m_ = material_properties["m"].template get<double>();
+
+    double sin_friction_cs = sin(friction_cs_);
```
bodhinandach: const?

include/material/norsand.tcc
```diff
@@ -0,0 +1,589 @@
...
+  de_(4,0)= 0;    de_(4,1)= 0;    de_(4,2)= 0;    de_(4,3)=0;    de_(4,4)=G;    de_(4,5)=0;
+  de_(5,0)= 0;    de_(5,1)= 0;    de_(5,2)= 0;    de_(5,3)=0;    de_(5,4)=0;    de_(5,5)=G;
+  // clang-format on
+
+  return true;
+}
+
+//! Compute stress invariants
+template <unsigned Tdim>
+bool mpm::NorSand<Tdim>::compute_stress_invariants(const Vector6d& stress,
```
bodhinandach: As I said during our meeting, such functions can be generalized into a namespace/class with static member functions so that it can be reused by any other constitutive models.

include/material/norsand.tcc
```diff
@@ -0,0 +1,589 @@
...
+                                                   mpm::dense_map* state_vars) {
+
+  // Note that in this subroutine, stress is compression positive
+
+  // Compute mean stress p
+  double mean_p = (stress(0) + stress(1) + stress(2)) / 3.;
+  if (mean_p < 1.0E-15) mean_p = 1.0E-15;
```
bodhinandach: I prefer: std::numeric_limits<double>::epsilon() instead of 1.0E-15

include/material/norsand.tcc
```diff
@@ -0,0 +1,589 @@
...
+  double j3 = (dev_stress(0) * dev_stress(1) * dev_stress(2)) -
+              (dev_stress(2) * pow(dev_stress(3), 2));
+  if (Tdim == 3)
+    j3 += ((2 * dev_stress(3) * dev_stress(4) * dev_stress(5)) -
+           (dev_stress(0) * pow(dev_stress(4), 2)) -
+           (dev_stress(1) * pow(dev_stress(5), 2)));
+  (*state_vars).at("j3") = j3;
+
+  // Compute q
+  double deviatoric_q = sqrt(3 * j2);
```
bodhinandach: It's always better to use std::sqrt() instead of sqrt().
See: https://stackoverflow.com/questions/26547095/why-sqrt-in-global-scope-is-much-slower-than-stdsqrt-in-mingw

include/material/norsand.tcc
```diff
@@ -0,0 +1,589 @@
...
+//! Compute yield function and yield state
+template <unsigned Tdim>
+typename mpm::NorSand<Tdim>::FailureState
+    mpm::NorSand<Tdim>::compute_yield_state(double* yield_function,
+                                            const Vector6d& stress,
+                                            const mpm::dense_map* state_vars) {
+
+  // Get stress invariants
+  // Compute mean stress p
+  double mean_p = (stress(0) + stress(1) + stress(2)) / 3.;
```
bodhinandach: I saw you code duplicate the computation of p and q, I think it should be made to at least a function.

bodhinandach (approved):
Thanks for the PR @ezrayst. I got some minor comments, but overall, looks good to me.

include/material/norsand.tcc
bodhinandach: @kks32 @ezrayst Perhaps, we can generalize this once the Mohr-Coulomb and MCC are merged from @tianchiTJ

ezrayst: Let's get this to work first, we can always refactor later.

include/material/norsand.tcc
ezrayst: I will use `std::` for everything now. Not sure why I didn't use it.

include/material/norsand.tcc
ezrayst: Agree, but there are two parts here: (1) computing, (2) storing as state-variables. I want to use the `compute_stress_invariant` here but I don't want to update state-variables. 

Another way is to have a function to get p and q in which `compute_stress_invariant` and `compute_yield_state` call. But I am not sure what's the efficiency of doing it over a few lines of simple computations where we have to pass vectors and maps to do simple computation. @kks32, any comment?

I think if we want to make a function, we should make a state-variables update function too. It will make things cleaner.

include/material/norsand.tcc
ezrayst: I purposefully used this because we know it's double, and I want to get it to zero even before it reaches the machine epsilon. Let's just say, this could be a parameter we put in.

include/material/norsand.tcc
ezrayst: I will use it for all :) I need to use it for `pow` too!

include/material/norsand.tcc
bodhinandach: @ezrayst it won't increase in efficiency, just code maintenance. The more you have code duplication, the more your code prone to bugs. Anyway, we can refactor later as discussed.

kks32 (commented):
Any particular reason why this PR is from your repo rather than the cb-geo/mpm repo?

include/material/norsand.h
```diff
@@ -0,0 +1,165 @@
+#ifndef MPM_MATERIAL_NOR_SAND_H_
+#define MPM_MATERIAL_NOR_SAND_H_
+
+#include <iostream>
+#include <limits>
+
+#include <cmath>
```
kks32, line 4: This should go above c++ includes

include/material/norsand.h
```diff
@@ -0,0 +1,165 @@
+#ifndef MPM_MATERIAL_NOR_SAND_H_
```
kks32:
```suggestion
#ifndef MPM_MATERIAL_NORSAND_H_
```

include/material/norsand.h
```diff
@@ -0,0 +1,165 @@
...
+#define MPM_MATERIAL_NOR_SAND_H_
+
+#include <iostream>
```
kks32: We don't need `iostream`

kks32 (changes requested):

ezrayst:
No reason in particular, except it's easier to push from my branch. No one else is working on it anyway

include/material/norsand.h
ezrayst: Yes, sorry, had it for debugging

include/material/norsand.h
ezrayst: Clang does not do this? I will make them on top then.

srhgk2 (approved):

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+  dev_stress(1) = stress(1) - mean_p;
+  dev_stress(2) = stress(2) - mean_p;
+  dev_stress(3) = stress(3);
+  if (Tdim == 3) {
+    dev_stress(4) = stress(4);
+    dev_stress(5) = stress(5);
+  }
+
+  // Compute J2
+  double j2 = (std::pow((stress(0) - stress(1)), 2) +
```
kks32, line 126: How does this change for 2d?

include/material/norsand.h
```diff
@@ -0,0 +1,168 @@
...
+  double p_cohesion_initial_{std::numeric_limits<double>::max()};
+  //! Initial p_dilation
+  double p_dilation_initial_{std::numeric_limits<double>::max()};
+  //! Cohesion degradation parameter m upon shearing
+  double m_cohesion_{std::numeric_limits<double>::max()};
+  //! Dilation degradation parameter m upon shearing
+  double m_dilation_{std::numeric_limits<double>::max()};
+  //! Parameter for shear modulus
+  double m_shear_{std::numeric_limits<double>::max()};
+
```
tianchiTJ, line 166: Can we unify all these name which may be involved in multi models?

ezrayst: I agree, but to what point?

I did do that for the general ones like `poisson_ratio_` but I thought the rest is pretty specific to NorSand. 

Also, in MCC, it is called p_cd and p_cc because it relates to the p_c, a parameter in MCC. But in NorSand, we do not have p_c, thus I named them `p_cohesion` and `p_dilation` to make it clear. 

Any suggestion?

tianchiTJ (approved):

include/material/norsand.h
```diff
@@ -0,0 +1,168 @@
...
+template <unsigned Tdim>
+class NorSand : public Material<Tdim> {
+ public:
+  //! Define a vector of 6 dof
+  using Vector6d = Eigen::Matrix<double, 6, 1>;
+  //! Define a Matrix of 6 x 6
+  using Matrix6x6 = Eigen::Matrix<double, 6, 6>;
+
+  //! Failure state
+  enum FailureState { Elastic = 0, Yield = 1 };
```
kks32:
```suggestion
  enum class FailureState { Elastic, Plastic};
```

include/material/norsand.h
```diff
@@ -0,0 +1,168 @@
...
+
+  //! Constructor with id and material properties
+  //! \param[in] material_properties Material properties
+  NorSand(unsigned id, const Json& material_properties);
+
+  //! Destructor
+  ~NorSand() override{};
```
kks32:
```suggestion
  ~NorSand() override = default;
```

include/material/norsand.h
```diff
@@ -0,0 +1,168 @@
...
+
+  //! Initialise history variables
+  //! \retval state_vars State variables with history
+  mpm::dense_map initialise_state_variables() override;
+
+  //! Thermodynamic pressure
+  //! \param[in] volumetric_strain dVolumetric_strain
+  //! \retval pressure Pressure for volumetric strain
+  double thermodynamic_pressure(double volumetric_strain) const override {
+    return 0;
```
kks32:
```suggestion
    std::numeric_limits<double>::quiet_NaN();
```

include/material/norsand.h
```diff
@@ -0,0 +1,168 @@
...
+  //! \param[in] stress Stress
+  //! \param[in] dstrain Strain
+  //! \param[in] particle Constant point to particle base
+  //! \param[in] state_vars History-dependent state variables
+  //! \retval updated_stress Updated value of stress
+  Vector6d compute_stress(const Vector6d& stress, const Vector6d& dstrain,
+                          const ParticleBase<Tdim>* ptr,
+                          mpm::dense_map* state_vars) override;
+
+  //! Compute stress invariants (p, q, etc)
```
kks32: Can these be private?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+      // J3
+      {"j3", 0.},
+      // M_theta
+      {"M_theta", Mtc_},
+      // Current void ratio
+      {"void_ratio", void_ratio_initial_},
+      // Void ratio image
+      {"e_image",
+       gamma_ - lambda_ * log(p_image_initial_ / reference_pressure_)},
+      // {"e_image",
```
kks32: This comment should be for above?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+      // Plastic strain components
+      {"plastic_strain0", 0.},
+      {"plastic_strain1", 0.},
+      {"plastic_strain2", 0.},
+      {"plastic_strain3", 0.},
+      {"plastic_strain4", 0.},
+      {"plastic_strain5", 0.}};
+
+  // Check for bond model e_image
+  double e_image;
```
kks32: This `e_image` would be undefined value, could you explain why is it defined outside the if condition?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+  if (bond_model_) {
+    e_image =
+        e_max_ -
+        (e_max_ - e_min_) /
+            log(crushing_pressure_ /
+                (p_image_initial_ + p_cohesion_initial_ + p_dilation_initial_));
+    if (e_image < 1.0E-15) e_image = 1.0E-15;
```
kks32: Do we need `e_image` to be such a low value?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+    state_vars.at("e_image") = e_image;
```
kks32: so if it is not `bond_model_` then the state variable `e_image` is not set at all?

include/material/norsand.tcc
kks32: then why not set `mean_p` to be 0? does that cause a `NAN`?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+  // Get state variables
+  const double mean_p = (*state_vars).at("p");
+  const double deviatoric_q = (*state_vars).at("q");
+  const double M_theta = (*state_vars).at("M_theta");
+  const double p_cohesion = (*state_vars).at("p_cohesion");
+  const double p_dilation = (*state_vars).at("p_dilation");
+  double p_image;
+  double e_image;
+
+  if (yield_type == FailureState::Elastic) {
```
kks32:
```suggestion
  if (yield_type == mpm::NorSand::FailureState::Elastic) {
```

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+          std::pow(((1 - N_ / M_theta * deviatoric_q / mean_p) / (1 - N_)),
+                   ((N_ - 1) / N_));
+    }
+    (*state_vars).at("p_image") = p_image;
+
+    // Compute and update void ratio image
+    // e_image = e_max_ - (e_max_ - e_min_) / log(crushing_pressure_ / p_image);
+    e_image = gamma_ - lambda_ * log(p_image / reference_pressure_);
+
+    if (e_image < 1.0E-15) e_image = 1.0E-15;
```
kks32, line 208: can we not set `e_image` to zero?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+
+    (*state_vars).at("e_image") = e_image;
+  }
+
+  // Update void ratio
+  // Note that dstrain is in tension positive - depsv = de / (1 + e_initial)
+  double dvolumetric_strain = dstrain(0) + dstrain(1) + dstrain(2);
+  double void_ratio = (*state_vars).at("void_ratio") -
```
kks32, line 220: You could inline a function that does a ternary operator check and return your value.

```suggestion
  double void_ratio = check_low((*state_vars).at("void_ratio") -(1 + void_ratio_initial_) * dvolumetric_strain);
  
  inline  double check_low(double val) {
    return (val > 1.0e-15 ? val : 0);
  }
```

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+}
+
+//! Compute elastic tensor
+template <unsigned Tdim>
+bool mpm::NorSand<Tdim>::compute_p_bond(mpm::dense_map* state_vars) {
+
+  // Compute current zeta cohesion
+  double zeta_cohesion = exp(-m_cohesion_ * (*state_vars).at("epds"));
+
+  if (zeta_cohesion > 1.)
```
kks32: Write a simple inline function that does this check

include/material/norsand.tcc
kks32: @ezrayst I agree with @bodhinandach have functions that compute `mean_p`, `j2`, `j3` etc so we don't have so many lines of code doing the same operations

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+  double deviatoric_q = std::sqrt(3 * j2);
+  if (deviatoric_q < 1.0E-15) deviatoric_q = 1.0E-15;
+
+  // Get image pressure and M_theta
+  const double p_image = (*state_vars).at("p_image");
+  const double M_theta = (*state_vars).at("M_theta");
+  const double p_cohesion = (*state_vars).at("p_cohesion");
+  const double p_dilation = (*state_vars).at("p_dilation");
+
+  // Initialise yield status (0: elastic, 1: yield)
```
kks32:
```suggestion
  // Initialise yield status (Elastic, Plastic)
```

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+  if (bond_model_) {
+    (*yield_function) =
+        deviatoric_q / (mean_p + p_cohesion) -
+        M_theta / N_ *
+            (1 + (N_ - 1) * std::pow(((mean_p + p_cohesion) /
+                                      (p_image + p_cohesion + p_dilation)),
+                                     (N_ / (1 - N_))));
+  } else {
+    (*yield_function) =
+        deviatoric_q / mean_p -
```
kks32: if we have a `non-bond` model could we just `p_cohesion` and `p_dilation` to zero, we then don't have to go through so many if conditions

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+    shear_modulus_ = 3. * bulk_modulus_ * (1. - 2. * poisson_ratio_) /
+                         (2.0 * (1. + poisson_ratio_)) +
+                     m_shear_ * ((*state_vars).at("p_cohesion") +
+                                 (*state_vars).at("p_dilation"));
+  } else {
+    shear_modulus_ = 3. * bulk_modulus_ * (1. - 2. * poisson_ratio_) /
+                     (2.0 * (1. + poisson_ratio_));
+  }
+
+  // Using Young's Modulus
```
kks32: Could you please remove these comments, if you are not using those?

include/material/norsand.tcc
```diff
@@ -0,0 +1,693 @@
...
+
+  // Trial stress - elastic
+  Vector6d trial_stress = stress_neg + (this->de_ * dstrain_neg);
+
+  // Initialise value for yield function
+  double yield_function;
+  auto yield_type =
+      this->compute_yield_state(&yield_function, trial_stress, state_vars);
+
+  // --------------------------------------------------------------------------------------
```
kks32: Could you please remove these lines, they are confusing

kks32 (changes requested):
Could we try reducing the repetitions in the code and make some of the repeated calls functions?

include/material/norsand.tcc
ezrayst: Yes, it could cause `nan` because there are many instances where we divide by `mean_p`

include/material/norsand.tcc
ezrayst: Will refactor just for `mean_p`, `j2`, `j3`, and `q` for now.

ezrayst: The idea is to have less repetition

include/material/norsand.tcc
ezrayst: OK. I initially wanted different definition of those. But it's ok, let's stick with one.

include/material/norsand.tcc
ezrayst: OK

include/material/norsand.tcc
ezrayst: This is just to check that `e_image` is not negative

include/material/norsand.tcc
ezrayst: Alright. I can do it!

include/material/norsand.tcc
ezrayst: I can remove it. It was another way of doing void ratio computation

ezrayst:
I will slowly refactor this. I wanted to make another test case for non-bond so to check that everything is still ok!

include/material/norsand.h
kks32: Add this to material.h

include/material/norsand.tcc
```diff
@@ -0,0 +1,666 @@
+//! Constructor with id and material properties
+template <unsigned Tdim>
+mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
+    : Material<Tdim>(id, material_properties) {
+  try {
+    // Density
+    density_ = material_properties["density"].template get<double>();
```
kks32: all these should be `.at()`

include/material/norsand.tcc
```diff
@@ -0,0 +1,666 @@
...
+  // Compute mean stress p
+  double mean_p = (stress(0) + stress(1) + stress(2)) / 3.;
+  if (mean_p < 1.0E-15) mean_p = 1.0E-15;
+  (*state_vars).at("p") = mean_p;
+
+  // Compute the deviatoric stress
+  Vector6d dev_stress = Vector6d::Zero();
+  dev_stress(0) = stress(0) - mean_p;
+  dev_stress(1) = stress(1) - mean_p;
+  dev_stress(2) = stress(2) - mean_p;
```
kks32: This is not correct for 2D

kks32 (changes requested):

include/material/norsand.tcc
ezrayst: I think the formulation of the model should always be in 3D, with 2D (plane strain or plane stress) as special cases of 3D. Thus the deviatoric stress should still be defined as 3D. What do you think?

include/material/norsand.tcc
ezrayst: Done so, except for a few derivatives needing if statements

include/material/norsand.tcc
ezrayst: Done so!

tests/material/norsand_test.cc
```diff
@@ -0,0 +1,602 @@
+#include <fstream>
```
kks32: Remove these file streams and `cmath`

include/material/norsand.tcc
```diff
@@ -0,0 +1,627 @@
...
+  else if (zeta_dilation < 1.0E-15)
+    (*state_vars).at("zeta_dilation") = 0.;
+  else
+    (*state_vars).at("zeta_dilation") = zeta_dilation;
+
+  // Update p_dilation
+  double p_dilation = p_dilation_initial_ * (*state_vars).at("zeta_dilation");
+  (*state_vars).at("p_dilation") = p_dilation;
+
+  return true;
```
kks32: Could you explain where does this fail to return true? Why is this function defined as a bool?

include/material/norsand.tcc
```diff
@@ -0,0 +1,627 @@
...
+  // Compute dF / dq
+  const double dF_dq = 1.;
+
+  // Compute the deviatoric stress
+  Vector6d dev_stress = Vector6d::Zero();
+  dev_stress(0) = stress(0) - mean_p;
+  dev_stress(1) = stress(1) - mean_p;
+  dev_stress(2) = stress(2) - mean_p;
+  dev_stress(3) = stress(3);
+  if (Tdim == 3) {
```
kks32: Do we need this check? Those will be zeros anyway?

include/material/norsand.tcc
```diff
@@ -0,0 +1,627 @@
...
+    dj3_dsigma(4) = dev2.dot(dev3);
+    dj3_dsigma(5) = dev1.dot(dev3);
+  }
+
+  // Compute dtheta / dsigma
+  Vector6d dtheta_dsigma = std::sqrt(3.) / 2. / cos(3 * lode_angle) /
+                           std::pow(j2, 1.5) *
+                           (dj3_dsigma - 3. / 2. * j3 / j2 * dj2_dsigma);
+  if (Tdim == 2) {
```
kks32, line 390: Either handle 2D as special cases or 3D, there seems to be both?

kks32 (changes requested):
Could you please make the previous request changes too?

include/material/norsand.tcc
ezrayst: I will change this to void then

include/material/norsand.tcc
ezrayst: In the current formulation, I think it's OK to have zero even though it's not physical. Previously, since I used logarithmic function, I tried to prevent zero.

include/material/norsand.tcc
ezrayst: No need, I have changed it.

include/material/norsand.tcc
ezrayst: I think I would handle 2D as special cases. I have changed the previous lines

tests/material/norsand_test.cc
ezrayst: `cmath` is required to compute M from friction angle, even at the test level. `stream` has been removed.

include/material/norsand.h
ezrayst: Moving this to private!

include/material/norsand.tcc
ezrayst: I have refactored it.

include/material/norsand.h
ezrayst: Can we do Elastic and Yield? Because depending on the model, we could have plastic strains before it yields (like subloading)

include/material/norsand.tcc
ezrayst: Nope, I have made them all general regardless dimensions

include/material/norsand.h
ezrayst: Changed.

include/material/norsand.h
ezrayst: Not relevant anymore

include/material/norsand.tcc
ezrayst: It is set still. `e_image` will not change in elastic range

include/material/norsand.tcc
ezrayst: Fixed enum!

include/material/norsand.tcc
ezrayst: Added for `one` and `zero` but not for the other ones that are used only once: negative one, and limits on lode angle

kks32 (approved):

PR #495, opened 2019-11-13
[HDF5] Modify write particles hdf5

bodhinandach:
As discussed, I moved the function to `particle`.

### DIFF
--- a/include/particle.h
+++ b/include/particle.h
@@ -55,10 +55,10 @@ class Particle : public ParticleBase<Tdim> {
   //! \retval status Status of reading HDF5 particle
   bool initialise_particle(const HDF5Particle& particle) override;
 
-  //! Output particle as HDF5 data
-  //! \param[in] particle HDF5 data of particle
-  //! \retval status Status of writing HDF5 particle
-  bool get_particle_data(unsigned phase, HDF5Particle& particle_data) override;
+  //! Retrun particle data as HDF5
+  //! \param[in] phase Properties of a given phase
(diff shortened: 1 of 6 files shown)


include/particle.h
```diff
@@ -55,6 +55,11 @@ class Particle : public ParticleBase<Tdim> {
   //! \retval status Status of reading HDF5 particle
   bool initialise_particle(const HDF5Particle& particle) override;
 
+  //! Output particle as HDF5 data
+  //! \param[in] particle HDF5 data of particle
+  //! \retval status Status of writing HDF5 particle
+  bool get_particle_data(unsigned phase, HDF5Particle& particle_data) override;
```
kks32:
```suggestion
  HDF5Particle particle_data(unsigned phase) const override;
```

kks32: Return a HDF5Particle object instead of pass by reference. We don't modify objects passed by reference. To modify we normally use a pointer. In this case returning a value and emplace into a vector is better

kks32 (changes requested):

kks32:
Please run through clang-format using `sh ../clang-tools/format.sh` before pushing from the build directory: https://circleci.com/gh/cb-geo/mpm/2106

include/particle.h
kks32: copy elision would mean the compiler will generate the same assembly code, so its better to say nothing gets modified in the function, we are likely to get a better performance

cbgeo (approved):

kks32 (approved):

PR #497, opened 2019-11-18
Update mass density when updating volume

bodhinandach:
The title says everything: basically to update the mass_density_ while volume_ is updated.

Suggestion for better for memory optimization such that we don't have to store the three of them: 
```
double mass_density(unsigned phase){
      return mass_(phase)/volume_(phase);
}
```

### DIFF
--- a/include/particle.tcc
+++ b/include/particle.tcc
@@ -43,6 +43,8 @@ bool mpm::Particle<Tdim, Tnphases>::initialise_particle(
   this->mass_(phase) = particle.mass;
   // Volume
   this->assign_volume(phase, particle.volume);
+  // Mass Density
+  this->mass_density_(phase) = particle.mass / particle.volume;
   // Set local size of particle
   Eigen::Vector3d psize;
   psize << particle.nsize_x, particle.nsize_y, particle.nsize_z;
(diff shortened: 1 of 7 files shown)


ezrayst:
Please make a test case too for this function.

include/particle.tcc
```diff
@@ -439,6 +439,7 @@ bool mpm::Particle<Tdim, Tnphases>::update_volume_strainrate(unsigned phase,
       Eigen::VectorXd strain_rate_centroid =
           cell_->compute_strain_rate_centroid(phase);
       this->volume_(phase) *= (1. + dt * strain_rate_centroid.head(Tdim).sum());
+      this->mass_density_(phase) = this->mass_(phase) / this->volume(phase);
```
srhgk2: For single point two phase formulation, I am not sure if this would work. Because, the water phase mass is not guaranteed accurate, so we cannot update density using mass_(phase).

srhgk2: @bodhinandach Please include Tianchi and Weijiang to this discussion before making this change.

bodhinandach: @srhgk2 I will discuss with them. Thanks for pointing it out.

srhgk2: Use strain rate (deformation gradient) to update density, if you want to update it.

bodhinandach: My motivation now is whether I should interpolate the mass_density_ too or not, cause I am not doing that through the hdf5

bodhinandach: @srhgk2 I modified the update as we discussed

kks32 (approved):

kks32:
@srhgk2 are we good to merge?

srhgk2 (approved):

PR #499, opened 2019-11-20
[WIP] Working in Progress UVP formulation

bodhinandach:
This is not yet to merge, just that it's easier for discussion for 2 phase 1 point MPM.


tests/particle_test.cc
```diff
@@ -17,11 +17,20 @@ TEST_CASE("Particle is checked for 1D case", "[particle][1D]") {
   const unsigned Dim = 1;
```
kks32, line 17: I'd recommend moving this to a separate two phase particle test file

include/particle.h
```diff
@@ -179,9 +198,24 @@ class Particle : public ParticleBase<Tdim> {
     this->stress_.col(phase) = stress;
   }
 
+  //! Initial pore pressure
+  //! \param[in] pore_fluid Index corresponding to fluid phase
+  //! \param[in] pore pressure Initial pore pressure
+  void initial_pore_pressure(const unsigned pore_fluid,
```
kks32, line 204: No need to `const` variables that are size `double` or below because it makes it less efficient. Pass those by value.

bodhinandach:
Things to modify (feel free to add/modify):

- [ ]  Derive Particle 2 phase from the Particle class and all the necessary functions
- [ ]  Create a register
- [ ]  Derive/modify hdf5 struct including more variables specific to 2 phase particle
- [ ]  Fix 2 phase explicit strategy accordingly (see the new USL/USF flag). 
- [ ]  Fix and add testing

include/mpm_explicit_twophase.h
```diff
@@ -0,0 +1,65 @@
...
+#include "mpm_base.h"
+
+namespace mpm {
+
+//! MPMExplicit class
+//! \brief A class that implements the fully explicit one phase mpm
+//! \details A single-phase explicit MPM
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class MPMExplicitTwoPhase : public MPMBase<Tdim> {
```
bodhinandach, line 13: Hmmm, I guess 2 phase explicit should be derived from mpm_explicit, not from mpm_base to reduce code duplication. What do you think @srhgk2 @kks32?

include/node.tcc
```diff
@@ -128,6 +130,49 @@ bool mpm::Node<Tdim, Tdof, Tnphases>::update_internal_force(
   return status;
 }
 
+//! Update mixture internal force (body force / traction force)
```
bodhinandach, line 133: Isn't it better to include all the specific 2 phase nodal functions to a separate .tcc file?

bodhinandach (commented):
just some thoughts

bodhinandach:
Moving this to private repo.

PR #502, opened 2019-11-22
:computer: Refactor phases in MPM core code

kks32:
Refactor and remove template argument phases in the MPM code to facilitate deriving particles of different phases and this helps with the MPI process as well.

Additionally, we have refactored `MPMExplicitUSF2D/3D` and `MPMExplicitUSL2D/3D` to `MPMExplicit2D/3D`

```
"analysis" : {
    "stress_update": "usf",
    "type": "MPMExplicit3D"
}
```

By default, it will perform `usf`.


kks32:
The node class needs phases template so #503 supersedes this request.

PR #503, opened 2019-11-23
:computer: Refactor and remove phases in particle class

kks32:
Refactor and remove template argument phases in the MPM Particle class to facilitate deriving particles of different phases and this helps with the MPI process as well.

Additionally, we have refactored MPMExplicitUSF2D/3D and MPMExplicitUSL2D/3D to MPMExplicit2D/3D
```
"analysis" : {
    "stress_update": "usf",
    "type": "MPMExplicit3D"
}
```
By default, it will perform usf.

### DIFF
--- a/include/material/newtonian.tcc
+++ b/include/material/newtonian.tcc
@@ -31,7 +31,7 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<2>::compute_stress(
 
   const unsigned phase = 0;
 
-  const double pressure = ptr->pressure(phase);
+  const double pressure = ptr->pressure();
 
   const double volumetric_strain = dstrain(0) + dstrain(1);
 
@@ -57,7 +57,7 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<3>::compute_stress(
 
   const unsigned phase = 0;
 
-  const double pressure = ptr->pressure(phase);
+  const double pressure = ptr->pressure();
 
   const double volumetric_strain = dstrain(0) + dstrain(1) + dstrain(2);
 
--- a/include/material/newtonian.tcc
+++ b/include/material/newtonian.tcc
@@ -29,8 +29,6 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<2>::compute_stress(
     const Vector6d& stress, const Vector6d& dstrain, const ParticleBase<2>* ptr,
     mpm::dense_map* state_vars) {
 
-  const unsigned phase = 0;
-
   const double pressure = ptr->pressure();
 
   const double volumetric_strain = dstrain(0) + dstrain(1);
@@ -54,8 +52,6 @@ template <>
 Eigen::Matrix<double, 6, 1> mpm::Newtonian<3>::compute_stress(
     const Vector6d& stress, const Vector6d& dstrain, const ParticleBase<3>* ptr,
     mpm::dense_map* state_vars) {
-
-  const unsigned phase = 0;
 
   const double pressure = ptr->pressure();
 
--- a/include/particle.h
+++ b/include/particle.h
@@ -267,8 +267,6 @@ class Particle : public ParticleBase<Tdim> {
   double mass_;
   //! Volume
   double volume_;
-  //! Phase
-  unsigned phase_{0};
   //! Size of particle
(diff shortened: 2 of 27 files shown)


include/material/newtonian.tcc
```diff
@@ -31,7 +31,7 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<2>::compute_stress(
 
   const unsigned phase = 0;
```
bodhinandach: @kks32 did you forget to delete this one?

include/material/newtonian.tcc
```diff
@@ -57,7 +57,7 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<3>::compute_stress(
...
```
bodhinandach: This one too

include/particle.h
```diff
@@ -310,45 +262,47 @@ class Particle : public ParticleBase<Tdim> {
...
-  Eigen::Matrix<double, 1, Tnphases> mass_density_;
+  double mass_density_;
   //! Mass
-  Eigen::Matrix<double, 1, Tnphases> mass_;
+  double mass_;
   //! Volume
-  Eigen::Matrix<double, 1, Tnphases> volume_;
+  double volume_;
+  //! Phase
+  unsigned phase_{0};
```
bodhinandach: I am not sure whether this variable is needed... Could you elaborate on what we can use this for? 
If this is needed to initiate the size of certain variables in nodes or to iterate over the phase inside the particle, should it be better to make it 1 by default for single-phase?
If this is used to call a certain variable, why we need to save it?

bodhinandach (commented):
Some concern. But overall looks okay. Thanks @kks32

include/material/newtonian.tcc
kks32: Yep!

include/particle.h
kks32: This is used to set the variable index for mass/momentum of a particular phase in the nodes. So we use a phase index of 0 for single phase. It doesn't have to be defined as a member variable, we can define it inside individual functions.

bodhinandach: If you have two phase or three, are you going to safe n number of unsigned such as solid_phase, water_phase, or air_phase? I rather define it in each function to call.

kks32: I have a new definition of phases using an enum, so the `phase_` is removed.

bodhinandach: I like this one! Thanks @kks32. Though, we might need to rethink about the naming for hydro-particle (if it is used alone), cause it will use phase=0, which is defined as solid.

kks32: Let's keep it this way for now.

bodhinandach (approved):

bodhinandach:
I am good with this. Let's hear from @srhgk2 @tianchiTJ and Weijian before merging.

srhgk2 (approved):

kks32:
@tianchiTJ could you please let us know if this is good?

include/particle.tcc
```diff
@@ -456,18 +438,16 @@ bool mpm::Particle<Tdim, Tnphases>::update_volume_strainrate(unsigned phase,
...
-        material_.at(phase) != nullptr) {
+    if (volume_ != std::numeric_limits<double>::max() && material_ != nullptr) {
       // Mass = volume of particle * mass_density
-      this->mass_density_(phase) =
-          material_.at(phase)->template property<double>(
-              std::string("density"));
-      this->mass_(phase) = volume_(phase) * mass_density_(phase);
+      this->mass_density_ =
+          material_->template property<double>(std::string("density"));
+      this->mass_ = volume_ * mass_density_;
```
tianchiTJ, line 450: If we want to treat the "particle" as the solid phase used for all conditions(1p, 2p, 3p), should we consider to add a "volume fraction" and change the computation of mass as " this->mass_ = volume_ * mass_density_* volume_fraction_solid;"? And default it as 1 for "single phase".

tianchiTJ:
> @tianchiTJ could you please let us know if this is good?

I think other parts are good for me, and only need to think carefully about the mass.

include/particle.tcc
kks32: We can do that when we add `uvp` branch.

PR #507, opened 2019-11-28
:computer: State variables and Material ID in HDF5

### LINKED ISSUE #374: Add particle state_variable to HDF5
**Describe the feature**
To present more field variables other than "stress, strain and velocity", which are stored in the particle class, some something should be added to the code. 

**Describe alternatives**
Add new part to the function "particles_vector_data" in "mesh.tcc".

kks32:
This PR includes the following:

- addition of state variables to HDF5 fixes #374. Allows the use of resume function when using materials with state variables. I tested it with Linear Elastic, giving it an arbitrary state variable and value. It works. [Obviously, I removed it from the code, so you won't find it in the PR]

- HDF5 now supports material id, so this allows resuming models with particles having different material ids. [Previous code also worked, but it may have been prone to errors]. This HDF5 with material id is needed for packing/sending particles across MPI tasks.


kks32:
@ezrayst Could you please run some tests with your MC model to see if you could resume and the state variables are exactly the same as before?

cbgeo (approved):

ezrayst:
I will check it @kks32!

include/particle.tcc
```diff
@@ -161,6 +192,18 @@ mpm::HDF5Particle mpm::Particle<Tdim>::hdf5() const {
 
   particle_data.cell_id = this->cell_id();
 
+  particle_data.material_id = this->material_id();
+
+  // Write state variables
+  if (material_ != nullptr) {
+    particle_data.nstate_vars = state_variables_.size();
+    unsigned i = 0;
```
bodhinandach: I think it's good to throw an error in case `state_variables_.size()` return more than 20, cause your particle_data cant handle that.

include/hdf5_particle.h
```diff
@@ -39,8 +38,33 @@ typedef struct HDF5Particle {
   mpm::Index cell_id;
   // Status
   bool status;
+  // Material id
+  unsigned material_id;
+  // Number of state variables
+  unsigned nstate_vars;
+  // State variables (init to zero)
+  double svars[20] = {0};
```
bodhinandach: Will the state variables equal to 0, i.e. uninitialized, be printed too in the write_particle_hdf5?

include/particle.tcc
```diff
@@ -161,6 +192,18 @@ mpm::HDF5Particle mpm::Particle<Tdim>::hdf5() const {
 
   particle_data.cell_id = this->cell_id();
 
+  particle_data.material_id = this->material_id();
+
+  // Write state variables
+  if (material_ != nullptr) {
+    particle_data.nstate_vars = state_variables_.size();
```
bodhinandach, line 199: I think it's good to throw an error if the `state_variables_.size()>20`. Otherwise, it will cause a segmentation fault while accessing `particle_data.svars[i]`, if `i>19`.

bodhinandach (approved):
Okay, some minor questions.

include/hdf5_particle.h
bodhinandach, line 46: Will you print all the svars value, even if it's not initialized, i.e. equal to 0, to the hdf5?

kks32: There is a restriction on size, we have to define a size with this design: have a fixed number of svars, otherwise, we may be able to store it as a vector

include/particle.tcc
kks32: Sure!

kks32:
https://www.sharcnet.ca/help/images/f/fa/MPI_Datatypes_with_Opaque_Types_seminar_2014.pdf

include/hdf5_particle.h
bodhinandach: It's fine to store them as a vector with a fix size. I mean when you print, are we printing all the vector values even it's 0? Or is there any check that we have to adaptively print depends on `nstate_vars`?

include/hdf5_particle.h
```diff
@@ -39,8 +38,33 @@ typedef struct HDF5Particle {
...
 } HDF5Particle;
 
+namespace hdf5::particle {
+const hsize_t NFIELDS = 53;
```
bodhinandach, line 50: Also, this NFIELDS. Are you fixing the total number of fields printed? This means that for 2 or 3 phase particles, with more variables, we need to derive this `HDF5Particle` struct right? Isn't it easier and better to make the struct to be implemented inside the `particle` class then?

bodhinandach (changes requested):
Some further thoughts.

include/hdf5_particle.h
kks32: We can't have the `struct` defined inside the `Particle` because we are serializing the `Particle` class. Yes, we need to derive `HDF5Particle` for 2 and 3 phase

bodhinandach (approved):

include/hdf5_particle.h
bodhinandach: Okay. I am sorry. You are right. I realized that it's also harder to access from `mesh` if we do that.

include/hdf5_particle.h
kks32: You mean, when we write to HDF5, yes, it's a table with a fixed size. HDF5 table is our restriction to have a fixed size, not the struct itself.

ezrayst:
@kks32, two things from me:

(1) Material ID seems to work very well now. You can run the following [entity ](https://www.dropbox.com/sh/05bp8yhtwzoud0d/AADzPMog1lGZSAPRkK97E0JPa?dl=0)test for it. You can change `resume` to `false` first and then change it back to `true` so we can test this part of the code too. I have tested it and it's good!

(2) State variables, however, are not working quite well. My test case is Iverson, but the coarsest mesh to make it quick. You can download them in this [link](https://www.dropbox.com/sh/23b4k0bosim6b0i/AABUFjZxBM-RlIcWuSRbByUWa?dl=0) too. Again, the thing about `resume` is the same, you can turn the `false` and `true` to check.

I was just merely checking `epds` and clearly they reset. You can see the comparison of the snapshots on the bottom too: (a) original, (b) with resume between step 99000 and 100000

image

kks32:
@ezrayst Could you go through your HDF5 data and check `nstate_variables`, and `svars`?

ezrayst:
@kks32, my bad, yeah, the vtk is not a good representation.

Running the same files with the same way got me the following `epds` for the last 20 particles:

 If run continuously, at step 100,000
```
 -3.67610013e-01 -6.68027112e-01 -1.05067425e+00 -1.28358691e+00
 -1.20412946e+00 -9.66024570e-01 -5.57930052e-01 -3.68667940e-01
 -2.58018258e-01 -2.53727933e-01 -1.82975029e-01 -1.58330111e-01
 -1.45942255e-01 -1.36686552e-01 -1.15436906e-01 -1.03968943e-01
 -1.10408405e-01 -1.42175285e-01 -1.41924452e-01 -1.07232920e-01]
```

If run resumed from 99,000, at step 100,000
```
 -3.67589636e-01 -6.67998976e-01 -1.05061033e+00 -1.28345797e+00
 -1.20409843e+00 -9.66004651e-01 -5.57899350e-01 -3.68656319e-01
 -2.58004120e-01 -2.53753562e-01 -1.82986546e-01 -1.58339465e-01
 -1.45938086e-01 -1.36718175e-01 -1.15470507e-01 -1.03989507e-01
 -1.10408497e-01 -1.42175925e-01 -1.41932015e-01 -1.07240019e-01
```

There is slight difference. I am still checking right now what is happening, but any thoughts on this?

kks32:
@ezrayst I don't think it's a state variable issue. We may have to look carefully to see if other variables have some differences or not.

PR #508, opened 2019-11-28
:computer: Mohr Coulomb softening

tianchiTJ:
I have made the test been through, and delete some unnecessary part. 
The use of the MC model is the same with the original one.


include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,136 @@
...
+template <unsigned Tdim>
+class MohrCoulomb : public Material<Tdim> {
+ public:
+  //! Define a vector of 6 dof
+  using Vector6d = Eigen::Matrix<double, 6, 1>;
+  //! Define a Matrix of 6 x 6
+  using Matrix6x6 = Eigen::Matrix<double, 6, 6>;
+
+  //! Failure state
+  enum FailureState { Elastic = 0, Tensile = 1, Shear = 2 };
```
kks32:
```suggestion
  enum class FailureState { Elastic, Tensile, Shear};
```

include/material/mohr_coulomb.h
```diff
@@ -0,0 +1,136 @@
+#ifndef MPM_MATERIAL_MOHR_COULOMB_H_
+#define MPM_MATERIAL_MOHR_COULOMB_H_
+
+#include <limits>
```
kks32: Swap the order of `<cmath>` and `<limits>`

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,586 @@
...
+  dev_stress(0) = stress(0) - mean_p;
+  dev_stress(1) = stress(1) - mean_p;
+  dev_stress(2) = stress(2) - mean_p;
+  dev_stress(3) = stress(3);
+  if (Tdim == 3) {
+    dev_stress(4) = stress(4);
+    dev_stress(5) = stress(5);
+  }
+  // Compute J2
+  double j2 = (std::pow((stress(0) - stress(1)), 2) +
```
kks32: Do we need to handle 2D specifically?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,586 @@
...
+
+  return true;
+}
+
+//! Compute yield function and yield state
+template <unsigned Tdim>
+typename mpm::MohrCoulomb<Tdim>::FailureState
+    mpm::MohrCoulomb<Tdim>::compute_yield_state(
+        Eigen::Matrix<double, 2, 1>* yield_function,
+        const mpm::dense_map* state_vars) {
```
kks32: Pass this by reference, if you are not modifying `state_vars`

include/material/mohr_coulomb.h
kks32: Alternatively, could define the `FailureState` outside the `MohrCoulomb` class in a namespace `MohrCoulomb`

kks32 (changes requested):

include/material/mohr_coulomb.tcc
tianchiTJ: I have delete the specified parts for 2D

include/material/mohr_coulomb.h
tianchiTJ: Is it necessary to create a new namespace, or just put it in the "mpm"

kks32: Yes, namespace is necessary.

tianchiTJ: I have create mohrcoulomb namespace.

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+  Eigen::Matrix<double, 2, 1> yield_function_trial;
+  auto yield_type_trial =
+      this->compute_yield_state(&yield_function_trial, (*state_vars));
+  // Return the updated stress in elastic state
+  if (yield_type_trial == mpm::mohrcoulomb::FailureState::Elastic)
+    return trial_stress;
+  //-------------------------------------------------------------------------
+  // Plastic-corrector stage: correct the stress back to the yield surface
+  // Define tolerance of yield function
+  double Tolerance = 1E-1;
```
ezrayst: Is it a typo? Is it 1E-15?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+}
+
+//! Compute yield function and yield state
+template <unsigned Tdim>
+typename mpm::mohrcoulomb::FailureState
+    mpm::MohrCoulomb<Tdim>::compute_yield_state(
+        Eigen::Matrix<double, 2, 1>* yield_function,
+        const mpm::dense_map& state_vars) {
+  // Tolerance for yield function
+  const double Tolerance = -1E-1;
```
bodhinandach, line 150: Following @ezrayst comment, this one too. Isn't it too large?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+      trial_stress - (p_multiplier * this->de_ * dp_dsigma_final);
+
+  // Check the update stress
+  // Compute stress invariants based on updated stress
+  this->compute_stress_invariants(updated_stress, state_vars);
+  // Compute yield function based on updated stress
+  yield_type_trial =
+      this->compute_yield_state(&yield_function_trial, (*state_vars));
+  // Define the maximum iteration step
+  int itr_max = 100;
```
bodhinandach: `const`
Further notes: @tianchiTJ Can you check for all variables that you don't change and make them to const?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+  (*state_vars).at("pstrain0") += dpstrain(0);
+  (*state_vars).at("pstrain1") += dpstrain(1);
+  (*state_vars).at("pstrain2") += dpstrain(2);
+  (*state_vars).at("pstrain3") += dpstrain(3);
+  (*state_vars).at("pstrain4") += dpstrain(4);
+  (*state_vars).at("pstrain5") += dpstrain(5);
+  // Update equivalent plastic deviatoric strain
+  (*state_vars).at("epds") = std::sqrt(
+      2. / 9. *
+          (std::pow(
```
bodhinandach: Could you please replace the 2. to 2? The performance of std::pow(x, int) is way better than std::pow(x, double).

include/material/mohr_coulomb.tcc
bodhinandach: I think this at least has to be E-6 for iteration.

include/material/mohr_coulomb.tcc
bodhinandach: @tianchiTJ this part is not used for iteration, so we can make it more accurate.

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+               ((2. * e_val - 1) * std::sqrt(sqpart));
+    if (fabs(m) < 1.E-22) m = 1.E-5;
+    double l = (4. * (1. - e_val * e_val) * std::pow(cos(theta), 2)) +
+               std::pow((2. * e_val - 1.), 2);
+    double r_mw = (l / m) * r_mc;
+    // Initialise meridional eccentricity
+    double xi = 0.1;
+    double omega = std::pow((xi * cohesion_peak_ * tan(psi)), 2) +
+                   std::pow((r_mw * std::sqrt(1.5) * rho), 2);
+    if (omega < 1.E-22) omega = 1.E-5;
```
bodhinandach: E-22 will mean 0 in double. instead, I think it's better to use `std::numeric_limits<double>::epsilon()`

bodhinandach (dismissed):

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,571 @@
...
+  this->compute_stress_invariants(updated_stress, state_vars);
+  // Compute yield function based on updated stress
+  yield_type_trial =
+      this->compute_yield_state(&yield_function_trial, (*state_vars));
+  // Define the maximum iteration step
+  int itr_max = 100;
+  // Initialise counter of iteration step
+  int itr = 0;
+  // Correct the stress again
+  while ((yield_function_trial(0) > Tolerance ||
```
kks32: Could you write it as a for loop instead?

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,558 @@
...
+//! Compute yield function and yield state
+template <unsigned Tdim>
+typename mpm::mohrcoulomb::FailureState
+    mpm::MohrCoulomb<Tdim>::compute_yield_state(
+        Eigen::Matrix<double, 2, 1>* yield_function,
+        const mpm::dense_map& state_vars) {
+  // Tolerance for yield function
+  const double Tolerance = -1E-1;
+  // Get stress invariants
+  const double& epsilon = state_vars.at("epsilon");
```
kks32: Could you please explain why these are `const double&` instead of `const double`

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,558 @@
...
+      std::sqrt(1.5) * rho *
+          ((sin(theta + M_PI / 3.) / (std::sqrt(3.) * cos(phi))) +
+           (cos(theta + M_PI / 3.) * tan(phi) / 3.)) +
+      (epsilon / std::sqrt(3.)) * tan(phi) - cohesion;
+  // Initialise yield status (0: elastic, 1: tension failure, 2: shear failure)
+  auto yield_type = mpm::mohrcoulomb::FailureState::Elastic;
+  // Check for tension and shear
+  if ((*yield_function)(0) > Tolerance && (*yield_function)(1) > Tolerance) {
+    // Compute tension and shear edge parameters
+    double n_phi = (1. + sin(phi)) / (1. - sin(phi));
```
kks32: Make these const

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,558 @@
...
+}
+
+//! Compute dF/dSigma and dP/dSigma
+template <unsigned Tdim>
+void mpm::MohrCoulomb<Tdim>::compute_df_dp(
+    mpm::mohrcoulomb::FailureState yield_type, const mpm::dense_map* state_vars,
+    const Vector6d& stress, Vector6d* df_dsigma, Vector6d* dp_dsigma,
+    double* dp_dq, double* softening) {
+  // Get stress invariants
+  const double& j3 = (*state_vars).at("j3");
```
kks32: Could we have them as `const double`

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,558 @@
...
+//! Compute stress
+template <unsigned Tdim>
+Eigen::Matrix<double, 6, 1> mpm::MohrCoulomb<Tdim>::compute_stress(
+    const Vector6d& stress, const Vector6d& dstrain,
+    const ParticleBase<Tdim>* ptr, mpm::dense_map* state_vars) {
+  // Get equivalent plastic deviatoric strain
+  const double& pdstrain = (*state_vars).at("pdstrain");
+  // Update MC parameters using a linear softening rule
+  if (softening_ && pdstrain > pdstrain_peak_) {
+    switch (pdstrain < pdstrain_residual_) {
```
kks32: Could we make this an `if`

include/material/mohr_coulomb.tcc
```diff
@@ -0,0 +1,558 @@
...
+                      &dp_dq, &softening);
+  double lambda =
+      ((df_dsigma.transpose() * this->de_).dot(dstrain)) /
+      (((df_dsigma.transpose() * this->de_).dot(dp_dsigma)) + softening);
+  // Initialise updated stress
+  Vector6d updated_stress = trial_stress;
+  // Initialise incremental of plastic deviatoric strain
+  double dpdstrain = 0.;
+  // Correction stress based on stress
+  switch (fabs(yield) < Tolerance) {
```
kks32: Could we make this an if

kks32 (changes requested):

include/material/mohr_coulomb.tcc
tianchiTJ: I have store these values in the "state_variable", so I just use the reference but not transferring their values.

kks32: As `state_vars.at()` will return a value, we won't be able to get the reference. Please use `const double`

tianchiTJ: OK, I will do that

kks32 (approved):

PR #509, opened 2019-12-03
Feature surrounding cell neighbour

bodhinandach:
Modifying compute cell neighbors to include the surrounding cells. Old tests are passing, but need further tests.


include/cell.h
```diff
@@ -79,6 +79,14 @@ class Cell {
   //! Number of nodes
   unsigned nnodes() const { return nodes_.size(); }
 
+  //! Return nodes id in a cell
+  std::vector<mpm::Index> nodes_id() {
```
kks32: Could we return a `const std::set` instead of a vector?

include/cell.h
```diff
@@ -79,6 +79,14 @@ class Cell {
...
+    std::vector<mpm::Index> nodes_id_lists;
+    for (unsigned i = 0; i < nodes_.size(); ++i)
```
kks32: Please use a range based for loop to iterate over nodes

include/mesh.h
```diff
@@ -298,6 +298,9 @@ class Mesh {
   //! Find cell neighbours
   void compute_cell_neighbours();
```
kks32: Could we move the construction of `node_cell_maps_` to `compute_cell_neighbours` instead of a separate function.

include/mesh.h
```diff
@@ -379,6 +387,8 @@ class Mesh {
   std::map<unsigned, std::shared_ptr<mpm::Material<Tdim>>> materials_;
   //! Logger
   std::unique_ptr<spdlog::logger> console_;
+  //! Neighbouring node-cell ids map
+  std::map<mpm::Index, std::set<mpm::Index>> node_cell_maps_;
```
kks32: Let's not store this, as we don't use it for anything other than finding neighbours. Could you please rename to `map_node_cells_`

include/mesh.tcc
```diff
@@ -232,30 +232,38 @@ void mpm::Mesh<Tdim>::iterate_over_cells(Toper oper) {
...
 template <unsigned Tdim>
 void mpm::Mesh<Tdim>::compute_cell_neighbours() {
-  for (auto citr = cells_.cbegin(); citr != cells_.cend(); ++citr) {
-    const auto faces = (*citr)->sorted_face_node_ids();
-    for (const auto& face : faces) {
-      faces_cells_.insert(
-          std::pair<std::vector<mpm::Index>, mpm::Index>(face, (*citr)->id()));
-    }
+  // Check if node_cell_maps is empty
+  if (node_cell_maps_.empty()) {
```
kks32: Cell neighbours are not going to change, so let's construct it here directly

include/mesh.tcc
```diff
@@ -232,30 +232,38 @@ void mpm::Mesh<Tdim>::iterate_over_cells(Toper oper) {
...
 
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_node_cell_maps() {
+  // Loop over the cells
+  for (auto citr = cells_.cbegin(); citr != cells_.cend(); ++citr) {
+    // Get cell id and nodes id
+    const auto cell_id = (*citr)->id();
+    const auto nodes_id_list = (*citr)->nodes_id();
+    // Populate node_cell_maps_ with the node_id and multiple cell_id
+    for (const auto& id : nodes_id_list) node_cell_maps_[id].insert(cell_id);
```
```suggestion
    auto cell_id = (*citr)->id();
    for (const auto& id : (*citr)->nodes_id()) node_cell_maps_[id].insert(cell_id);
```

kks32 (changes requested):
Could we not store the `map_node_cells_` in `mesh`, but rather just use it once and throw away?

include/mesh.h
```diff
@@ -309,6 +312,11 @@ class Mesh {
   //! Return the number of neighbouring meshes
   unsigned nneighbours() const { return neighbour_meshes_.size(); }
 
+  //! Node cell maps
+  std::map<mpm::Index, std::set<mpm::Index>> node_cell_maps() const {
```
kks32: Could you try using the `mpm::dense_map` instead?

include/mesh.h
bodhinandach: @kks32 I am okay not to store this. Initially I stored it because the neighbours set in the cell has been used for the sharing-edge neighbour. I will refactor.

kks32 (approved):

PR #514, opened 2019-12-09
Change state vars to 25

ezrayst:
Since `norsand` has 21 state variables, I think we need to increase this number to 25.

### DIFF
--- a/include/hdf5_particle.h
+++ b/include/hdf5_particle.h
@@ -43,11 +43,11 @@ typedef struct HDF5Particle {
   // Number of state variables
   unsigned nstate_vars;
   // State variables (init to zero)
-  double svars[20] = {0};
+  double svars[25] = {0};
 } HDF5Particle;
 
 namespace hdf5::particle {
-const hsize_t NFIELDS = 53;
+const hsize_t NFIELDS = 58;
 
 const size_t dst_size = sizeof(HDF5Particle);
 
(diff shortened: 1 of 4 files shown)


kks32:
Are storing state variables like `p` and `q` necessary, could it not be easily computed from the stresses? Could you please explain the reasoning behind storing these state variables?

kks32 (changes requested):
Is this change necessary if you can drop `p` and `q` from state variables in NorSand as they can be computed easily from the stresses?

ezrayst:
If the choice is to save them as state variables or to compute them, I rather do the former because we will end up passing or computing them a lot (we have different functions that use them). I would make a subroutine to reduce repetition in `norsand` for sure. 

Any thoughts why 25 is not favorable compared to 20?

kks32:
You need to compute mean p and q already at each step so I don't understand why you need to store it. We can avoid increasing the number of state variables, which is a far bigger impact than computing and doing an addition operation. 20 is already a lot and with millions of particles this will be an unnecessary communication overhead. If we can't avoid it, it makes sense, but we can avoid it, so why try to increase it unnecessarily?

ezrayst:
I will try to refactor the constitutive model then. I should be able to pass them in different functions. Maybe that's a better solution than to increase the number of state variables.

kks32:
Closing this PR as NorSand can be refactored to not need an increase in the number of state variables.

PR #547, opened 2020-01-15
:rocket: Adding noexcept to improve performance

kks32:
The code shouldn't throw, unless the exception can be handled. If not, this should be checked by assert and not a throw. The standard library is changing how to do logic and runtime errors, and this improves performance and aligns us to use Contracts when it's available in c++20.


cbgeo (approved):

cbgeo:
Improves performance by 13%

PR #548, opened 2020-01-16
:art: Photo-realistic rendering with Disney partio

kks32:
Support photo-realistic rendering with Houdini


cbgeo (approved):
Add partio to circleci

PR #551, opened 2020-01-24
:hammer: Revert improve node-node halo region MPI communication

kks32:
Reverts cb-geo/mpm#550


cbgeo (approved):
Reverting because of performance loss

PR #565, opened 2020-02-15
Clone particle material history 

### LINKED ISSUE #564: Clone particle material history from a neighbour
**Describe the feature**
When adding a new particle, the material history of the new particle is cloned from a given neighbour particle.

- [x]  Add a function in particle class to clone the material state variables from a given particle hdf5 data

srhgk2:
Cloning material history of a particle from a given neighbour particle.
resolves #564

### DIFF
--- a/include/particle.tcc
+++ b/include/particle.tcc
@@ -248,17 +248,14 @@ bool mpm::Particle<Tdim>::assign_material_state_vars(
     const mpm::dense_map& state_vars,
     const std::shared_ptr<mpm::Material<Tdim>>& material) {
   bool status = false;
-  if (material != nullptr) {
-    if (this->material_id_ == material->id() ||
-        this->material_id_ == std::numeric_limits<unsigned>::max()) {
-      // Assign material
-      // material_ = material;
-      // Clone state variables
-      auto mat_state_vars = material_->initialise_state_variables();
-      if (mat_state_vars.size() == state_vars.size()) {
-        this->state_variables_ = state_vars;
-        status = true;
-      }
+  if (material != nullptr && this->material_ != nullptr &&
+      this->material_id_ == material->id()) {
+    // Clone state variables
+    auto mat_state_vars = material_->initialise_state_variables();
+    if (this->state_variables_.size() == state_vars.size() &&
+        mat_state_vars.size() == state_vars.size()) {
+      this->state_variables_ = state_vars;
+      status = true;
     }
   }
   return status;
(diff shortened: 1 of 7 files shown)


include/particle.tcc
```diff
@@ -242,6 +242,38 @@ void mpm::Particle<Tdim>::initialise() {
   this->properties_["displacements"] = [&]() { return displacement(); };
 }
 
+//! Clone material state variables from neighbour particle
+template <unsigned Tdim>
+bool mpm::Particle<Tdim>::clone_neighbour_material_state_vars(
+    const mpm::dense_map& state_vars,
+    const std::shared_ptr<mpm::Material<Tdim>>& material) {
+  bool status = true;
```
kks32:
```suggestion
  bool status = false;
  if (material != nullptr) {
    if (this->material_id_ == material->id() ||
        this->material_id_ == std::numeric_limits<unsigned>::max()) {
      material_ = material;
      // Clone state variables
      auto mat_state_vars = material_->initialise_state_variables();
      if (mat_state_vars.size() == state_vars.size()) {
        this->state_variables_ = state_vars;
        status = true;
      } 
    }
  return status;
```

kks32: I'm trying to avoid having throws, as C++ standard is changing to use `throw` only when you can recover from an error. Here, if we have any of the conditions failing we can't recover from it, so no reason to `throw`.

srhgk2: okay

include/particle.tcc
```diff
@@ -242,6 +242,28 @@ void mpm::Particle<Tdim>::initialise() {
...
+template <unsigned Tdim>
+bool mpm::Particle<Tdim>::assign_material_state_vars(
+    const mpm::dense_map& state_vars,
+    const std::shared_ptr<mpm::Material<Tdim>>& material) {
+  bool status = false;
+  if (material != nullptr) {
+    if (this->material_id_ == material->id() ||
+        this->material_id_ == std::numeric_limits<unsigned>::max()) {
+      // Assign material
+      // material_ = material;
```
kks32: I wrote a test function, and I think this should be commented out. The reason is if we don't do this, we are essentially overwriting the existing material, in which case, the condition `if (mat_state_vars.size() == state_vars.size())` will always be true. I tested this with two materials with same material ids (not likely to happen, but a test case, where it will pass the first if, and it will overwrite the existing material with the new one. Since we are only interpolating new material properties, there is no need to overwrite the material pointer in the particle, what do you think? @srhgk2

kks32 (changes requested):
@srhgk2 let me know what you think

kks32 (approved):

PR #568, opened 2020-02-18
Interface class implementation

thiagordonho:
Implementation:

- Define Interface class with friction parameters.
- Define struct for nodal (multimaterial) properties.

### DIFF
--- a/include/interface.h
+++ b/include/interface.h
@@ -50,6 +50,4 @@ class Interface {
 };  // Interface class
 }  // namespace mpm
 
-#include "interface.tcc"
-
 #endif  // MPM_INTERFACE_INTERFACE_H_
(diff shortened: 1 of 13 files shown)


include/multimaterial_properties.h
```diff
@@ -0,0 +1,25 @@
+#ifndef MPM_MULTIMATERIAL_H_
```
kks32, line 1: remove this for this PR

include/interface.h
```diff
@@ -0,0 +1,53 @@
+#ifndef MPM_INTERFACE_INTERFACE_H_
```
kks32, line 1: Why do we need a class? can this not be saved directly if there is only one parameter?

kks32 (changes requested):

kks32:
Closing this PR due to code restructuring

PR #569, opened 2020-02-19
Neighbour Particles

bodhinandach:
I am adding some functions to add neighbor particles in a similar fashion to cells. There are three functions of compute_particle_neighbours overloaded:
1. global: if needed for all particles
2. for particles inside a specific cell: suppose free surface cell, or cell with refinement 

I will add testing in the following commits. Let me know if you have any comments and suggestions to optimize it.


include/cell.h
```diff
@@ -117,6 +117,9 @@ class Cell {
   //! \retval insertion_status Return the successful addition of a node
   bool add_neighbour(mpm::Index neighbour_id);
 
+  //! Has neighbours
+  bool has_neighbours() const { return neighbours_.size(); }
```
kks32: What difference does this function make compared to checking `nneighbours`?

include/mesh.tcc
```diff
@@ -69,27 +69,26 @@ bool mpm::Mesh<Tdim>::remove_node(
...
 template <typename Toper>
 void mpm::Mesh<Tdim>::iterate_over_nodes(Toper oper) {
-  tbb::parallel_for(
-      tbb::blocked_range<int>(size_t(0), size_t(nodes_.size()),
-                              tbb_grain_size_),
-      [&](const tbb::blocked_range<int>& range) {
-        for (int i = range.begin(); i != range.end(); ++i) oper(nodes_[i]);
-      },
-      tbb::simple_partitioner());
+  tbb::parallel_for(tbb::blocked_range<int>(size_t(0), size_t(nodes_.size()),
```
kks32: Please run through clang-format

kks32 (changes requested):
We need MPI as particles are likely to be at different MPI tasks

include/particle.tcc
```diff
@@ -788,3 +788,41 @@ void mpm::Particle<Tdim>::append_material_id_to_nodes() const {
...
+}
+
+//! Assign neighbour particles
+template <unsigned Tdim>
+bool mpm::Particle<Tdim>::assign_neighbours(
+    const std::vector<mpm::Index>& neighbours_set) {
+  bool status = true;
+  try {
+    // Remove the existing neighbours' id
+    neighbours_.clear();
```
kks32: neighbours.erase(std::remove(neighbours.begin(), neighbours.end(), neighbour_id), neighbours.end());

include/mesh.tcc
```diff
@@ -299,22 +298,104 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+                              cells_[i]->add_neighbour(neighbour_id);
+                        }
+                      }
+                    },
+                    tbb::simple_partitioner());
+}
+
+//! Find particle neighbours for all particle
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_particle_neighbours() {
```
kks32: These following sections need MPI Calls

include/mesh.tcc
```diff
@@ -299,22 +298,104 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+          map_particles_[p_id]->assign_neighbours(neighbouring_particle_sets);
+      if (!status)
+        throw std::runtime_error("Cannot assign valid particle neighbours");
+    }
+  }
+}
+
+//! Find particle neighbours for all particle in a given cell
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_particle_neighbours(const Cell<Tdim>& cell) {
```
kks32: use a default function argument as a nullptr, so if nothing is passed it would be the same function as above.

include/particle.h
```diff
@@ -262,6 +262,26 @@ class Particle : public ParticleBase<Tdim> {
...
+  //! Has neighbour particles
+  bool has_neighbours() const override { return neighbours_.size(); };
+
+  //! Return the number of neighbour particles
+  unsigned nneighbours() const override { return neighbours_.size(); };
+
+  //! Add a neighbour particle
+  //! \param[in] neighbour_id id of the neighbouring particle
+  //! \retval insertion_status Return the successful addition of a node
+  bool add_neighbour(mpm::Index neighbour_id) override;
```
kks32: I'm unable to find where the `add_neighbour` function is used, could we remove it, if it is not used?

include/mesh.tcc
```diff
@@ -299,22 +298,104 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+//! Find particle neighbours for all particle
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_particle_neighbours() {
+  // Check whether cells have been assigned with neighbours
+  if (!(*cells_.cbegin())->has_neighbours()) this->compute_cell_neighbours();
+
+  // Loop over cells
+  for (auto citr = cells_.cbegin(); citr != cells_.cend(); ++citr) {
+    // Loop over the neighbouring cell particles
+    std::vector<mpm::Index> neighbouring_particle_sets = (*citr)->particles();
```
kks32: avoid using type names. This is a vector and not a set. Please call it `neighbour_particles`

include/particle.h
bodhinandach: If you create a new particle, instead of searching the neighbours again, it is better to just add the created particle id into the existing particles. That's true that it's not used, but I think it is useful.

include/mesh.tcc
bodhinandach: without default `nullptr`, wouldn't it work too?

include/cell.h
srhgk2: @kks32 Yes, I have asked this before and i think we can use `neighbours()` as both are same.

include/particle.tcc
bodhinandach: I think this is to remove the current particle id from `neighbours_` right? But, what I want here is to empty `neighbours_` as it has been assigned with new vector of neighbours

include/cell.h
bodhinandach: Resolved

include/mesh.tcc
bodhinandach: I made default `nullptr` in the header. Resolved

include/mesh.tcc
bodhinandach: It should be done already... always do it while compiling. Not sure why it changes...

include/mesh.tcc
kks32: I was thinnking of merging this function and the one above, they are very similar

include/particle.h
kks32: Shall we add it when we start using it?

include/particle.tcc
kks32: Yes, but then are checking if the id is present or not, instead let's just use the erase on find idiom

include/mesh.tcc
kks32:
```
mpm::Container<mpm::Cell> cell_ctr;
cell_ctr.add(cell);
auto cells = (cell == nullptr) ? cells_ : cell_ctr;
for (auto c : cells) {

```

include/particle.h
```diff
@@ -330,6 +346,8 @@ class Particle : public ParticleBase<Tdim> {
   std::unique_ptr<spdlog::logger> console_;
   //! Map of vector properties
   std::map<std::string, std::function<Eigen::VectorXd()>> properties_;
+  //! Container of particle neighbour ids
+  std::vector<mpm::Index> neighbours_;
```
srhgk2: Don't you think it's good to have this in `particle base`? what do you think @kks32 ?

bodhinandach: I am fine with it to be moved to base if necessary.

kks32: Sure

bodhinandach: Done!

include/mesh.tcc
bodhinandach: @kks32 I think all comments are sorted. If you can help me with the MPI Calls, that will be helpful. I will create the tests afterward.

kks32: @bodhinandach Would it be possible for you to create unit tests (without MPI) so I can test the MPI call implementation and update the tests? Otherwise it would be hard to know if MPI is working or not! Thanks for your help

include/mesh.tcc
kks32: Not sure how you run `clang-format`. One option is to `cd clang-tools` and run `sh ./format.sh`. We use one `.clang-format` file in the root directory, instead of having additional ones in each folder we create.

include/particle.h
kks32: We may have to recompute neighbours of all the particles at each step anyway, so I don't think this function helps much. If we see an issue in performance, we can think again. For now, let's remove this function.

kks32: The `compute_particle_neighbours` with a particular `cell` should improve this performance, without any additional function at the particle level.

include/particle.tcc
```diff
@@ -788,3 +788,36 @@ void mpm::Particle<Tdim>::append_material_id_to_nodes() const {
...
+  }
+  return status;
+}
+
+//! Assign neighbour particles
+template <unsigned Tdim>
+bool mpm::Particle<Tdim>::assign_neighbours(
+    const std::vector<mpm::Index>& neighbours) {
+  bool status = true;
+  try {
```
kks32: I don't see a possible exception here, could we make this function `void` and remove the `try` block completely?

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+                        }
+                      }
+                    },
+                    tbb::simple_partitioner());
+}
+
+//! Find particle neighbours for all particle
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_particle_neighbours(const Cell<Tdim>& cell) {
+  // Check argument passed
```
kks32: Please clarify what's being done here.

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+  Container<Cell<Tdim>> cell_ctr;
```
```suggestion
  Container<Cell<Tdim>> cell_container;
```

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+  cell_ctr.add(cell);
+  auto& cells = (cell_ctr == nullptr) ? cells_ : cell_ctr;
+
+  // Check whether cells have been assigned with neighbours
+  if ((*cells_.cbegin())->nneighbours() == 0)
+    throw std::runtime_error(
```
kks32: There is no reason to throw, as we don't deal with this exception.

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+        "No neighbours have been assigned to cell, cannot "
+        "compute_particle_neighbours for single cell");
+
+  // Loop over cells
+  for (auto citr = cells.cbegin(); citr != cells.cend(); ++citr) {
```
kks32: Could you make this loop `tbb::parallel_for`, a simple version without partitioners

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+    // Loop over the neighbouring cell particles
```
kks32: This comment is not very clear on what the following line or the iteration is doing

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+  cell_ctr.add(cell);
+  auto& cells = (cell_ctr == nullptr) ? cells_ : cell_ctr;
+
+  // Check whether cells have been assigned with neighbours
+  if ((*cells_.cbegin())->nneighbours() == 0)
+    throw std::runtime_error(
+        "No neighbours have been assigned to cell, cannot "
+        "compute_particle_neighbours for single cell");
+
+  // Loop over cells
```
kks32: Please remove this redundant comment

include/mesh.tcc
```suggestion
    // Loop over all neighboring cells, and append particle ids from each cell
```

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+  for (auto citr = cells.cbegin(); citr != cells.cend(); ++citr) {
+    // Loop over the neighbouring cell particles
+    std::vector<mpm::Index> neighbour_particles = (*citr)->particles();
+    for (const auto& neighbour_cell_id : (*citr)->neighbours())
+      neighbour_particles.insert(
+          neighbour_particles.end(),
+          map_cells_[neighbour_cell_id]->particles().begin(),
```
kks32: Can we get the `particle_ids` and then add it to the existing vector, here aren't we asking for the value twice?

include/mesh.tcc
```diff
@@ -299,22 +298,84 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+          map_cells_[neighbour_cell_id]->particles().end());
+
+    // Loop over the particles in the current cells and assign particle
+    // neighbours
+    for (const auto& p_id : (*citr)->particles()) {
+      bool status =
+          map_particles_[p_id]->assign_neighbours(neighbour_particles);
+      if (!status)
```
kks32: No throw is required here

kks32 (changes requested):

kks32 (commented):
@bodhinandach I have added MPI calls, could you please add the unit tests, so I can check the MPI implementation

bodhinandach:
@kks32 Sure! Thanks a lot, also for the modification. I will make the test sometime in the upcoming week.

bodhinandach:
@kks32 as for the `add_neighbour` function in particle, I meant to have this function for adding a single particle to the neighbourhood, but not necessarily do the search for the cell. Think about adaptive particle refinement or cloning. But, I guess it can be removed at the moment as it is not currently used.

include/mesh.tcc
```diff
@@ -298,84 +299,47 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
...
+      auto& particle_ids = map_cells_[neighbour_cell_id]->particles();
+      neighbour_particles.insert(neighbour_particles.end(),
+                                 particle_ids.begin(), particle_ids.end());
     }
-  }
-}
 
-//! Find particle neighbours for a given particle
-template <unsigned Tdim>
-void mpm::Mesh<Tdim>::compute_particle_neighbours(
```
bodhinandach: Why this is removed? This is important for the boundary particles. Could we please revert?

include/mesh.tcc
```diff
@@ -317,6 +317,76 @@ void mpm::Mesh<Tdim>::compute_cell_neighbours() {
       tbb::simple_partitioner());
 }
 
+//! Find particle neighbours for all particle
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::find_particle_neighbours(const Cell<Tdim>& cell) {
```
bodhinandach: I intended to make the naming the same as the one for cell, isn't it better to have similar function naming?

kks32: I'm going to change the name in cell

include/mesh.tcc
kks32: we can accomplish something similar with passing the cell to `find_particle_neighbours`, also we may have to recompute neighbors at each step, so this specialized function won't have a big effect.

bodhinandach: I got what you mean. I do have some points to say though. I will talk to you privately.

include/particle.h
```diff
@@ -262,6 +262,17 @@ class Particle : public ParticleBase<Tdim> {
   //! Assign material id of this particle to nodes
   void append_material_id_to_nodes() const override;
 
+  //! Return the number of neighbour particles
+  unsigned nneighbours() const override { return neighbours_.size(); };
```
kks32: What is the intended use of `nneighbours`, could we not use `neighbours` and get the size?

include/mesh.tcc
bodhinandach: I did that

bodhinandach:
@kks32 @srhgk2 can you check this once more? This is ready to merge I guess.

kks32:
@bodhinandach It's not done yet, we need to still add tests to check if the particle neighbours are working fine when cells are distributed across different MPI tasks. I'll update your test cases with MPI calls

bodhinandach:
@kks32 Thanks for adding the MPI test. Just curiosity, with the current test format, can you run in both non-mpi and mpi?

kks32 (approved):

kks32:
@bodhinandach It will run only for MPI

PR #571, opened 2020-02-20
[Material] Fix newtonian

### LINKED ISSUE #572: Mistakes in Newtonian implementation
**Describe the bug**
There is a slight mistake in the formulation of newtonian material. Strain rate should be used instead of dstrain. Fixed by https://github.com/cb-geo/mpm/pull/571

bodhinandach:
I am fixing Newtonian fluid as we discussed @kks32. The formulation used is taken from the fluid mechanics equation to obtain shear stress:
image
I made a flag if the fluid is incompressible, the divergence of velocity is zero, and thus the second term should go. I will make some other adjustment in the Navier Stokes branch to generalize the pressure term with Chorin's projection.

If this looks good, I will work on the tests, they are not working now.
Fix #572.

### DIFF
--- a/include/material/newtonian.h
+++ b/include/material/newtonian.h
@@ -69,8 +69,8 @@ class Newtonian : public Material<Tdim> {
   double bulk_modulus_{std::numeric_limits<double>::max()};
   //! Fluid Dynamic viscosity
   double dynamic_viscosity_{std::numeric_limits<double>::max()};
-  //! Fluid Incompressibility
-  bool incompressible_{false};
+  //! Compressibility multiplier
+  double compressibility_multiplier_{1.0};
 
 };  // Newtonian class
 }  // namespace mpm
--- a/include/material/newtonian.h
+++ b/include/material/newtonian.h
@@ -61,12 +61,12 @@ class Newtonian : public Material<Tdim> {
   //! Thermodynamic pressure
   //! \param[in] volumetric_strain dVolumetric_strain
   //! \retval pressure Pressure for volumetric strain
-  double thermodynamic_pressure(double volumetric_strain_rate) const;
+  double thermodynamic_pressure(double volumetric_strain) const;
 
   //! Density
   double density_{std::numeric_limits<double>::max()};
-  //! Fluid Bulk viscosity
-  double bulk_viscosity_{std::numeric_limits<double>::max()};
+  //! Fluid Bulk modulus
+  double bulk_modulus_{std::numeric_limits<double>::max()};
   //! Fluid Dynamic viscosity
   double dynamic_viscosity_{std::numeric_limits<double>::max()};
   //! Fluid Incompressibility
--- a/include/material/newtonian.h
+++ b/include/material/newtonian.h
@@ -61,14 +61,16 @@ class Newtonian : public Material<Tdim> {
   //! Thermodynamic pressure
   //! \param[in] volumetric_strain dVolumetric_strain
   //! \retval pressure Pressure for volumetric strain
-  double thermodynamic_pressure(double volumetric_strain) const;
+  double thermodynamic_pressure(double volumetric_strain_rate) const;
 
   //! Density
   double density_{std::numeric_limits<double>::max()};
-  //! Bulk modulus
-  double bulk_modulus_{std::numeric_limits<double>::max()};
-  //! mu - constant plastic viscosity [N s / m^2 or kg / m / s]
-  double mu_{std::numeric_limits<double>::max()};
+  //! Fluid Bulk viscosity
+  double bulk_viscosity_{std::numeric_limits<double>::max()};
+  //! Fluid Dynamic viscosity
+  double dynamic_viscosity_{std::numeric_limits<double>::max()};
+  //! Fluid Incompressibility
+  bool incompressible_{false};
 
 };  // Newtonian class
 }  // namespace mpm
--- a/include/material/newtonian.tcc
+++ b/include/material/newtonian.tcc
@@ -10,9 +10,11 @@ mpm::Newtonian<Tdim>::Newtonian(unsigned id, const Json& material_properties)
         material_properties.at("dynamic_viscosity").template get<double>();
 
     // Special material properties
-    if (material_properties.contains("incompressible"))
-      incompressible_ =
+    if (material_properties.contains("incompressible")) {
+      bool incompressible =
           material_properties.at("incompressible").template get<bool>();
+      if (incompressible) compressibility_multiplier_ = 0.0;
+    }
 
     properties_ = material_properties;
   } catch (Json::exception& except) {
@@ -45,7 +47,7 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<2>::compute_stress(
   // Get strain rate
   const auto& strain_rate = ptr->strain_rate();
 
-  // Update volumetric and deviatoric stress
+  // Deviatoric stress component
   Eigen::Matrix<double, 6, 1> pstress;
   pstress(0) = 2. * dynamic_viscosity_ * strain_rate(0);
   pstress(1) = 2. * dynamic_viscosity_ * strain_rate(1);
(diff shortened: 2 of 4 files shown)


include/material/newtonian.tcc
```diff
@@ -25,9 +27,9 @@ mpm::dense_map mpm::Newtonian<Tdim>::initialise_state_variables() {
 //! Compute pressure
 template <unsigned Tdim>
 double mpm::Newtonian<Tdim>::thermodynamic_pressure(
-    double volumetric_strain) const {
+    double volumetric_strain_rate) const {
   // Bulk modulus
-  return (-bulk_modulus_ * volumetric_strain);
+  return (-bulk_viscosity_ * volumetric_strain_rate);
```
srhgk2: you return dp/dt? Is that what you want? I don't think so.

srhgk2 (changes requested):

srhgk2:
Can you give us any reference to the equations you are using?

include/material/newtonian.tcc
bodhinandach: nope. I am returning to stress. See the screenshot above.

bodhinandach:
Sure, here it is. Wikipedia is my friend: https://en.wikipedia.org/wiki/Viscosity#General_definition

include/material/newtonian.tcc
srhgk2: Sorry, I don't understand. Can you explain a bit more clearly what is `-bulk_density * volumetric_strain_rate`. For me it is dp/dt.

srhgk2:
Also, standard explicit MPM always assume weak compressibility, so its never fully incompressible. And for projection method, you can add compressibility/incompressibility by using bulk modulus. Consider these when you generalize things.

bodhinandach:
@srhgk2 you are right and I agree with you. I want to discuss it first here. I am still not sure that the pressure that I wrote here is correct, particularly if pressure is not updated inside the material. Take a look at this complete NS equation
image
- Grad(P) is a pressure caused by the flow potential and external stresses
- The Grad(\zeta * div(v)) term is the bulk viscosity term.

The thing is, we often combine them with strain through bulk_modulus inspired by solid mechanics, which is actually fine and correct, but generalizing them is tricky. 

I am still thinking of a way to generalize it with the fluid code I am developing, particularly the pressure update. So if that's okay, I want to spend more time thinking about it and, probably make the modification when the fluid code is ready. I made this PR to make you aware that there is a formulation error in the prior Newtonian.

But I am aware of your concern. Thanks.

include/material/newtonian.tcc
kks32: You need to return `dp` not `dp/dt`, as we compute `dp = K * (dv / v0) ` so this should be dvolumetric_strain, not `dvolumetric_strain_rate`, Shyamini is saying exactly this, I think

include/material/newtonian.tcc
```diff
@@ -36,26 +38,36 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<2>::compute_stress(
...
 
-  // const double pressure = ptr->pressure();
-  // Update pressure
-  (*state_vars).at("pressure") +=
-      this->thermodynamic_pressure(ptr->dvolumetric_strain());
-  const double pressure = (*state_vars).at("pressure");
-
-  const double volumetric_strain = dstrain(0) + dstrain(1);
+  // Get strain rate
+  const auto& strain_rate = ptr->strain_rate();
```
kks32, line 48: This is the only change, we use `strain_rate` instead of `dstrain` in the stress calculations

kks32:
I have updated the mpm-doc on Newtonian: https://mpm.cb-geo.com/#/theory/material/newtonian please check

include/material/newtonian.tcc
bodhinandach: Okay. I think you are right. I will reformulate. Thank you :smiley:

include/material/newtonian.tcc
```diff
@@ -6,7 +6,10 @@ mpm::Newtonian<Tdim>::Newtonian(unsigned id, const Json& material_properties)
     density_ = material_properties.at("density").template get<double>();
     bulk_modulus_ =
         material_properties.at("bulk_modulus").template get<double>();
-    mu_ = material_properties.at("mu").template get<double>();
+    dynamic_viscosity_ =
+        material_properties.at("dynamic_viscosity").template get<double>();
+    incompressible_ =
```
srhgk2: Is this a necessary material property then? I think it would be better to pass this from the projection method solver. Because if the user accidentally gives it as incompressible, the standard MPM solver cannot handle it. What do you think?

bodhinandach: I think passing a variable to a material requires a special function, which is public and needs to be defined in the base material class. I guess that's unnecessary. 

To compensate, I make the default value of `incompressible_` to be `false` and make a check at the Newtonian constructor to look whether the user defines incompressible bool or not in the input.

Here, by default, the code is the same with the old MPM and can run in explicit too without additional parameter.

What do you think?

bodhinandach: See:
https://github.com/cb-geo/mpm/blob/073c703eaf8271df31993c9860a17e499244d03f/include/material/newtonian.tcc#L12-L15

include/material/newtonian.h
```diff
@@ -65,10 +65,12 @@ class Newtonian : public Material<Tdim> {
 
   //! Density
   double density_{std::numeric_limits<double>::max()};
-  //! Bulk modulus
+  //! Fluid Bulk modulus
   double bulk_modulus_{std::numeric_limits<double>::max()};
-  //! mu - constant plastic viscosity [N s / m^2 or kg / m / s]
-  double mu_{std::numeric_limits<double>::max()};
+  //! Fluid Dynamic viscosity
+  double dynamic_viscosity_{std::numeric_limits<double>::max()};
```
kks32, line 71: What is the reasoning behind this breaking change?, why not just leave it as `mu_`

include/material/newtonian.h
```diff
@@ -65,10 +65,12 @@ class Newtonian : public Material<Tdim> {
...
+  //! Fluid Incompressibility
+  bool incompressible_{false};
```
kks32: Since we are almost always dealing with weakly compressible, could we remove this and create a new class `NewtonianFluid` that can deal with returning shear stresses for your projection method without the pressure term.

kks32 (changes requested):
Could you please make this PR just modifying `dstrain` to `strain_rate` without adding new features. We can discuss those later when you have the projection method.

include/material/newtonian.h
bodhinandach: I can revert, but I guess the naming is pretty confusing. We should avoid naming variables with greek letter and be more explicit. Different material models might denote different variables with `mu_` or any other greek letters.

include/material/newtonian.h
bodhinandach: hmmm... That was my initial thought, but after we clarify that Newtonian also takes strain_rate, not dstrain. I dont see differentiating the materials is necessary. I guess, making NewtonianFluid class will cause even more confusion and unnecessary code duplication than merely adding a boolean as material property.

I am at the moment developing the projection method, and I am finding problems with stresses, that's why I came up with this. Perhaps, we can talk on Monday? @srhgk2 @kks32

srhgk2: All you need is shear stress, i.e. the deviatoric part of your stress, because the volumetric part is handled by the projection method. It would be better to see your implementation to see why you are having problems.

bodhinandach: I am aware of that. I am sure that this is not the only bug. If you add pressure in the `compute_stress` function to the returned `pstress`, that causes an error already right?

srhgk2: So, create a new material model as @kks32 suggested.

bodhinandach: My argument is the same as I wrote, but if you think that's the best way, I will do it. Thanks

srhgk2: I can understand that its not good to duplicate every fluid model. But until we come up with a better solution, I would suggest let's go with the most straightforward way.

bodhinandach: I think this is a good time to rethink about this. Let's discuss it soon.

include/material/newtonian.tcc
```diff
@@ -64,24 +84,37 @@ Eigen::Matrix<double, 6, 1> mpm::Newtonian<3>::compute_stress(
     const Vector6d& stress, const Vector6d& dstrain, const ParticleBase<3>* ptr,
     mpm::dense_map* state_vars) {
 
+  // Get strain rate
+  const auto& strain_rate = ptr->strain_rate();
+
+  // Deviatoric stress component
+  Eigen::Matrix<double, 6, 1> pstress;
+  pstress(0) = 2. * dynamic_viscosity_ * strain_rate(0);
```
kks32: Could we combine this line of `pstress` with the later line in line# 108, so we can do it once

bodhinandach: Good idea!

include/material/newtonian.h
bodhinandach: Resolved as agreed

kks32 (approved):

srhgk2 (approved):

PR #573, opened 2020-02-26
Particle injection

kks32:
Inject particles in specific cell during specific duration


cbgeo (approved):

bodhinandach (approved):
It seems okay. As far as I see, velocity, material, volume, and mass are added. Any way that we can specify a certain input of material like stresses, strains, and the state variables?

kks32:
I think when we do copy properties from neighbours, we can do that

PR #574, opened 2020-02-26
Property pool struct

thiagordonho:
Define property pool struct for handling multimaterial properties in the nodes belonging to interface between materials.


include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <Eigen/Dense>
+#include <map>
+
+namespace mpm {
+// \brief Multimaterial parameters on each node
+struct NodalProperties {
+
+  //! Constructor
+  NodalProperties();
```
cbgeo: Remove this

include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+  //! \param[in] columns Number of materials
+  bool create_property(const std::string& property, unsigned rows,
+                       unsigned columns);
+
+  // Fetch data in the nodal properties map
+  // \param[in] property Property name
+  // \param[in] id_node Id of the node within the property data
+  // \param[in] id_mat Id of the material within the property data
+  // \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  Eigen::MatrixXd property_data(const std::string& property, unsigned id_node,
```
cbgeo: rename to just `property`

cbgeo: This should be const

include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+
+  //! Function to create new property with given name and size (rows x cols)
+  //! \param[in] property Property name
+  //! \param[in] rows Number of nodes times the number of the dimension of the
+  //! property (1 if scalar, Tdim if vector)
+  //! \param[in] columns Number of materials
+  bool create_property(const std::string& property, unsigned rows,
+                       unsigned columns);
+
+  // Fetch data in the nodal properties map
```
cbgeo:
```suggestion
  // Return data in the nodal properties map at a specific index
```

include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+  // \param[in] property Property name
+  // \param[in] id_node Id of the node within the property data
```
cbgeo: `node_id` and `mat_id`

include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+
+  // Assign property value to a pair of node and material
+  // \param[in] property_value Property value to be assigned
+  // \param[in] property Property name
+  // \param[in] id_node Id of the node within the property data
+  // \param[in] id_mat Id of the material within the property data
+  // \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  bool assign_property(const std::string& property,
+                       Eigen::MatrixXd property_value, unsigned id_node,
+                       unsigned id_mat, unsigned nprops = 1);
```
cbgeo: why is `nprops` needed?

include/nodal_properties.h
```diff
@@ -0,0 +1,44 @@
...
+  Eigen::MatrixXd property_data(const std::string& property, unsigned id_node,
+                                unsigned id_mat, unsigned nprops = 1);
+
+  // Assign property value to a pair of node and material
+  // \param[in] property_value Property value to be assigned
+  // \param[in] property Property name
+  // \param[in] id_node Id of the node within the property data
+  // \param[in] id_mat Id of the material within the property data
+  // \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  bool assign_property(const std::string& property,
```
cbgeo: move the `value` to the end of the function argument

src/nodal_properties.cc
```diff
@@ -0,0 +1,36 @@
+#include "nodal_properties.h"
+
+// Constructor
+mpm::NodalProperties::NodalProperties() {
```
cbgeo: Delete this

src/nodal_properties.cc
```diff
@@ -0,0 +1,36 @@
...
+      std::pair<std::string, Eigen::MatrixXd>(property, property_data));
+  return true;
+}
+
+// Fetch data in the nodal properties map
+Eigen::MatrixXd mpm::NodalProperties::property_data(const std::string& property,
+                                                    unsigned id_node,
+                                                    unsigned id_mat,
+                                                    unsigned nprops = 1) {
+  return properties_.at(property).block(id_node * nprops, id_mat, nprops, 1);
```
cbgeo: You may need to use a const map, my guess is `block` will be very slow.

cbgeo (approved):

kks32 (changes requested):
Could you try map of Eigen matrix, block is likely to be slower

kks32 (commented):
Please run `clang-format` from build `sh ../clang-tools/format.sh`

include/nodal_properties.h
thiagordonho: To differentiate properties that are scalar (like mass) of those that are vectors (like velocity or momentum). However, since I won't be using block operations anymore, this might not be needed in this member function

src/nodal_properties.cc
```diff
@@ -0,0 +1,30 @@
...
+  properties_.insert(
+      std::pair<std::string, Eigen::MatrixXd>(property, property_data));
+}
+
+// Return data in the nodal properties map at a specific index
+Eigen::MatrixXd mpm::NodalProperties::property(const std::string& property,
+                                               unsigned node_id,
+                                               unsigned mat_id,
+                                               unsigned nprops) const {
+  double* position = &properties_.at(property)(node_id * nprops, mat_id);
```
kks32:
```suggestion
  const double* position = &properties_.at(property)(node_id * nprops, mat_id);
```

kks32: The return type here is `const double *` and not `double *`, hence the failure in clang

src/nodal_properties.cc
```diff
@@ -0,0 +1,30 @@
...
+  mpm::MapProperty property_map(position, nprops);
+  return property_map;
+}
+
+// Assign property value to a pair of node and material
+void mpm::NodalProperties::assign_property(const std::string& property,
+                                           unsigned node_id, unsigned mat_id,
+                                           Eigen::MatrixXd property_value,
+                                           unsigned nprops) {
+  properties_.at(property).block(node_id * nprops, mat_id, nprops, 1) =
```
kks32, line 33: What does the `1` at the end denote?

kks32 (changes requested):

src/nodal_properties.cc
thiagordonho: The two last arguments are an indication of the size of the block. nprops is the number of rows and 1 is the number of columns. Since only one material is returned, 1 is always the argument passed in this case.

kks32: Please add that in the comment

kks32 (approved):

src/nodal_properties.cc
```diff
@@ -0,0 +1,33 @@
+#include "nodal_properties.h"
+
+// Function to create new property with given name and size (rows x cols)
+void mpm::NodalProperties::create_property(const std::string& property,
+                                           unsigned rows, unsigned columns) {
+  // Initialize a matrix with size of rows times columns and insert it to the
+  // property database map
+  Eigen::MatrixXd property_data = Eigen::MatrixXd::Zero(rows, columns);
+  properties_.insert(
```
kks32: Return the result of insertion status

kks32 (changes requested):

include/nodal_properties.h
```diff
@@ -0,0 +1,46 @@
...
+
+// \brief Multimaterial parameters on each node
+struct NodalProperties {
+
+  //! Function to create new property with given name and size (rows x cols)
+  //! \param[in] property Property name
+  //! \param[in] rows Number of nodes times the number of the dimension of the
+  //! property (1 if scalar, Tdim if vector)
+  //! \param[in] columns Number of materials
+  void create_property(const std::string& property, unsigned rows,
```
bodhinandach: I would probably put a default value of the number of materials to 1.

include/nodal_properties.h
```diff
@@ -0,0 +1,46 @@
...
+#include <map>
+
+namespace mpm {
+
+// Define Eigen map type
+typedef Eigen::Matrix<double, Eigen::Dynamic, 1> MatrixProperty;
+typedef Eigen::Map<const MatrixProperty> MapProperty;
+
+// \brief Multimaterial parameters on each node
+struct NodalProperties {
```
bodhinandach, line 14: Perhaps, CellProperties too? That will be helpful for elemental properties, like local matrices, etc.

tests/nodal_properties_test.cc
```diff
@@ -0,0 +1,166 @@
...
+      for (int j = 0; j < nmaterials; ++j) {
+        REQUIRE_NOTHROW(nodal_properties.assign_property(
+            property1, i, j, data1.block<dim, 1>(i * dim, j), dim));
+        REQUIRE_NOTHROW(nodal_properties.assign_property(
+            property2, i, j, data2.block<dim, 1>(i * dim, j), dim));
+      }
+    }
+
+    // Check values in matrix of property data
+    for (int i = 0; i < nnodes; ++i) {
```
bodhinandach, line 77: What will be the default value of inactive nodes? What will happen if one node was active and become inactive as particle moves?

tests/nodal_properties_test.cc
```diff
@@ -0,0 +1,166 @@
...
+          REQUIRE(nodal_properties.property(property2, i, j, dim)(k, 0) ==
+                  Approx(0.0).epsilon(tolerance));
+        }
+      }
+    }
+
+    // clang-format off
+    Eigen::Matrix<double,6,2> data1;
+    data1 << 4.2, 3.1,
+             1.0, 1.7, // end of first node
```
bodhinandach, line 130: Just want to confirm, is the property arranged in this way for velocity at node 1: 
v_x_1, v_x_2
v_y_1, v_y_2
or the transpose of this?

tests/nodal_properties_test.cc
```diff
@@ -0,0 +1,166 @@
...
+  SECTION("Create vector properties") {
+    // Declare nodal properties
+    mpm::NodalProperties nodal_properties;
+
+    // Define properties to be created
+    std::string property1 = "velocities";
+    std::string property2 = "momenta";
+
+    // Define dimension (2D)
+    const unsigned dim = 2;
```
bodhinandach, line 99: If I have a matrix, should I arrange it in an array?

bodhinandach (commented):
Some comments. But overall great idea @thiagordonho @kks32

include/nodal_properties.h
kks32: Each row is a new material property, hmm setting a default would mean changing the order of the parameters of rows and cols. Eigen is column-major. We should probably reconsider how we arrange these materials.

tests/nodal_properties_test.cc
kks32: This is just for testing. We have `status` parameter to deal with those situations

tests/nodal_properties_test.cc
kks32: For cells, we'll have a vector of matrices

tests/nodal_properties_test.cc
bodhinandach: I see. So for each node the array size will just be dim * nmaterials for each property?

kks32: There will be one property pool, if you add your node to it, it means it's active and it will get a handle, if it is a non-active node, you won't get a handle or add to pool

bodhinandach: Does that mean the size of the pool change with the number of active nodes? It means it requires you either to construct the pool at every step (size nodal quantity is reset anyway) or to resize the pool.

kks32: Yes, the pool size will change dynamically, but only when load rebalancing happens, not at every step

bodhinandach: Do you consider special cases, such as when we have 2 materials at a shared node? I am thinking about contact or two-phase two-layers implementation

kks32: That's the whole idea behind this property pool to handle contact

bodhinandach: Awesome!

tests/nodal_properties_test.cc
thiagordonho: The one you indicated is the correct one. Columns represent materials and rows represent nodes **and** the dimensions if dealing with vector quantities.

kks32: so we store it in column-major format?

kks32 (approved):

PR #562, opened 2020-02-15
[Constitutive] Modified Cam-Clay

tianchiTJ:

PR #577, opened 2020-03-11
:art: Restructure container files and materials

kks32:
- Move containers to a separate folder
- Rename container as vector to be more specific
- Rename `material` folder to `materials` to be consistent with other folders


cbgeo (approved):

bodhinandach (approved):
I am okay, but why you change the name from Container->Vector? Any specific thoughts? Are you planning to generate a Matrix container too?

bodhinandach:
@kks32 shall we wait for #562 to merge first before merging this?

kks32:
@bodhinandach Map is technically a container, so I'm being more specific with renaming it to a Vector. We are likely to have other containers, and since we mirror std::vector functions it's easier to call it a Vector than a more generic container. 

I'll wait for #562

bodhinandach:
I am good. Thanks Krishna! We can merge now

PR #579, opened 2020-03-12
Output vtk for shear stresses and strains

ezrayst:
- [x] vtk output for shear stresses
- [x] vtk output for shear strains
- [x] always get all 3 components (especially for stresses and strains)


include/particles/particle.h
```diff
@@ -161,6 +161,11 @@ class Particle : public ParticleBase<Tdim> {
   //! Return strain of the particle
   Eigen::Matrix<double, 6, 1> strain() const override { return strain_; }
 
+  //! Return shear strain of the particle
+  Eigen::Matrix<double, 3, 1> shear_strain() const override {
```
kks32: Can we not just `tail` the output of `strain`?

include/mesh.tcc
```diff
@@ -69,27 +69,26 @@ bool mpm::Mesh<Tdim>::remove_node(
...
 template <typename Toper>
 void mpm::Mesh<Tdim>::iterate_over_nodes(Toper oper) {
-  tbb::parallel_for(
-      tbb::blocked_range<int>(size_t(0), size_t(nodes_.size()),
-                              tbb_grain_size_),
-      [&](const tbb::blocked_range<int>& range) {
-        for (int i = range.begin(); i != range.end(); ++i) oper(nodes_[i]);
-      },
-      tbb::simple_partitioner());
+  tbb::parallel_for(tbb::blocked_range<int>(size_t(0), size_t(nodes_.size()),
```
kks32: Please run clang-format using `clang-tools/format.sh`

include/particles/particle.tcc
```diff
@@ -237,7 +237,9 @@ void mpm::Particle<Tdim>::initialise() {
 
   // Initialize vector data properties
   this->properties_["stresses"] = [&]() { return stress(); };
+  this->properties_["shear_stresses"] = [&]() { return shear_stress(); };
```
kks32: Could you please remind me where this is called?

kks32 (changes requested):

include/mesh.tcc
bodhinandach: @kks32 I think we need to check this. The `clang-tools/format.sh` always return the correction like @ezrayst has. You saw it in my previous PR too...

include/particles/particle.tcc
ezrayst: First in `mpm-base.tcc` [here](https://github.com/cb-geo/mpm/blob/0675471fceb83d95afae78ca5cc4ffb7f7bcf2ad/include/solvers/mpm_base.tcc#L502), then in `mesh.tcc` [here](https://github.com/cb-geo/mpm/blob/0675471fceb83d95afae78ca5cc4ffb7f7bcf2ad/include/mesh.tcc#L934)

include/particles/particle.h
ezrayst: We can do that, so less functions. Shall I go ahead with this?

include/mesh.tcc
kks32: You can see it fail in `circleci` https://circleci.com/gh/cb-geo/mpm/3979?pipelines-ui-opt-out, so it's a local issue

Which version of clang-tools do you have?

clang-tools/format.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 find ../include/ -iname *.h -o -iname *.tcc | xargs clang-format -style=file -i
 find ../src/ -iname *.h -o -iname *.cc | xargs clang-format -style=file -i
-find ../tests/ -iname *.h -o -iname *.cc | xargs clang-format -style=file -i
```
kks32: Please don't modify this file

include/mesh.tcc
ezrayst:
```
ezrayst@ezrayst-acer:~/Documents/cb-geo/mpm/build$ clang-format --version
clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
```

clang-tools/format.sh
ezrayst: I did not, sorry let me revert back

include/mesh.tcc
bodhinandach: I am having the same version as ezra

include/solvers/mpm_base.tcc
```diff
@@ -81,8 +81,9 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
   }
 
   // Default VTK attributes
-  std::vector<std::string> vtk = {"velocities", "stresses", "strains",
-                                  "displacements"};
+  std::vector<std::string> vtk = {"velocities",     "stresses",
```
bodhinandach, line 84: In 3D will this output twice shear stresses and strains @ezrayst? Or the normal "stresses" and "strains" only output the normal components?

ezrayst: Normal only

include/mesh.tcc
kks32: Could you please upgrade to the later versions: `clang-format version 9.0.1`

include/solvers/mpm_base.tcc
kks32: @ezrayst `stress()` returns a 6-component vector, so Nanda is right

include/mesh.tcc
```diff
@@ -940,8 +940,11 @@ std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::particles_vector_data(
       Eigen::Vector3d data;
       data.setZero();
       auto pdata = (*pitr)->vector_data(attribute);
-      // Fill stresses to the size of dimensions
-      for (unsigned i = 0; i < Tdim; ++i) data(i) = pdata(i);
+      // Fill data attributes to size of 3
+      const unsigned attribute_size =
```
kks32, line 944: I'm sorry but I don't understand what's happening here with all these std::min and max?

Can't we just copy the data to the dimension size? I feel like we don't have to change this section.

kks32 (changes requested):

kks32:
@ezrayst the build also fails in debug mode in Clang, could you please fix those

include/solvers/mpm_base.tcc
ezrayst: @kks32 but `mesh` will only take the first 3 components right, so normal only.

include/mesh.tcc
ezrayst: `vtk` has to be a component of 3, so we cannot iterate more than 3.
In cases of `displacement` and `velocities` in 2D, we want to only output 2 components, not 3 since `pdata` is a vector of dimension 2.
But in cases of `stresses` in 2D, we want to output all 3 components, not only 2. So in this case `pdata` is a vector of dimension 6, so we take the first 3.

Therefore we need to do this. Maybe you have a better idea?

include/solvers/mpm_base.tcc
bodhinandach: These are the line of codes.
https://github.com/cb-geo/mpm/blob/38172eed4eae47c23503661f3189b3684dba9611/include/mesh.tcc#L940-L947
I think @ezrayst  is right @kks32. `pdata` is a 6-component vector, but it is copied to `data` which is only 3-component. I checked that vtk accepts a 6-component vector too. So we might want to think about whether we should expand stresses or strains-like vectors to 6 component vector if we like to.

include/mesh.tcc
bodhinandach: @ezrayst I think the prior one is okay as @kks32 mentioned. `TDim` will always be 2 or 3 right? So even if `pdata` is 6-component vector, it will only copy the first `TDim` values.

ezrayst: But I want to get stress in 3 directions even in 2D

kks32: The magnitude output of VTK will be incorrect if you copy uncomputed 3rd component of stress in 2D

kks32: This change affects not just stresses, but other dimensional vectors such as velocities, so I don't think it's a good idea to change this.

include/solvers/mpm_base.tcc
kks32: Yes, it's only when you copy you truncate the components

include/mesh.tcc
ezrayst: But the 3rd component of stress in 2D is computed!

ezrayst: Zero strain in that direction does not mean zero stress in that direction.

kks32:
@ezrayst @bodhinandach the best course of action would be to use `tensor` notation for both `stress` and `strain`. I'll create a new branch to do that. Shall we close this PR?

bodhinandach:
@kks32 I think that's better to distinguish stresses and strains as tensors.

bodhinandach:
But maybe array works better (?), cause we don't have to print all the 9 components due to symmetry. Try to reduce the output file. See:
https://www.kitware.com/products/books/VTKUsersGuide.pdf

ezrayst:
Understood, let's close this PR

kks32:
Issue #582 will address the output

PR #580, opened 2020-03-12
:computer: Particles set id when generating particles - fixes #578

### LINKED ISSUE #578: Implementation of particle 'set_id's within input.json file
Currently, the implementation of particle entity sets "set_id" is not working.

This would add the ability to define point entity sets that can have boundary conditions (load, velocity etc) at creation, and from multiple files.

The example is given below is how this implementation would look in the mpm.json file:

```
"particles": [
    {
      "generator": {
        "check_duplicates": false,
        "location": "3dparticles.txt",
        "io_type": "Ascii3D",
        "particle_type": "P3D",
        "material_id": 0,
        "type": "file"
      },
      "set_id": 0
    },
    {
      "generator": {
        "check_duplicates": false,
        "location": "cone_out.txt",
        "io_type": "Ascii3D",
        "particle_type": "P3D",
        "material_id": 1,
        "type": "file"
      },
      "set_id": 1
    }
  ],
 ```

### PR DESCRIPTION
**Pull request by kks32**
Automatically assign particles set id on generation

```
      "generator": {
        "check_duplicates": true,
        "location": "particles.txt",
        "io_type": "Ascii2D",
        "particle_type": "P2D",
        "material_id": 1,
        "pset_id": 0,
        "type": "file"
      }
```


cbgeo (approved):

bodhinandach (commented):
I am okay and this is essentially very useful.
But I got one concern, will we check whether `pset_id` has been defined in `entity_sets` for BC? I mean, what happens if I want to apply force, body-force like but not gravity, to all the generated material with `pset_id = 0`, should I define it again in the `entity_sets.json`? What is the appropriate `pset_id` written in the JSON file?

kks32:
@bodhinandach I'm not sure if I understood your question correctly. 

You don't have to add a `pset_id` to `entity_sets` again. This will automatically add you a new set id similar to what you have in `entity_sets` so you don't have to add it again to entity sets. This is an additional way to do so. Does this answer your question?

bodhinandach:
I meant, is the `pset_id` defined in both `generator` and `entity_set` unique to each other? Suppose that I have defined `pset_id=0` in `generator`, I should not define `pset_id=0` anymore in `entity_set`.

kks32:
Yes, that's correct, the onus is on the user, but if the user does that, when the code runs it will throw an error to say particle set creation failed. This error should be sufficient to tell the user that they have made an incorrect input.

bodhinandach (approved):

cw646 (approved):

PR #581, opened 2020-03-12
:computer: Continue running simulation when locate particles fn fails flag

kks32:
This PR adds a new optional flag to analysis, which allows one to continue the analysis when locating particles fail (when particles go out of the mesh). The default is set to true and will throw an error when particles are not found.

```
  "analysis": {
    "locate_particles": true,
   }
```


bodhinandach (approved):
Do you think that we need a test for this?

bodhinandach:
Also, will we delete the particles outside the mesh?

kks32:
I have a benchmark test which works as expected. I can add that to the benchmarks repo. Unit test for this is difficult as this is not a function, but an option in the solver.

If we can't find a particle in the mesh, it will be ignored, not deleted.

bodhinandach:
@kks32 I just concerned whether it will appear in the visualization, but thanks again.

kks32:
@bodhinandach when it fails to locate it is no longer in the mesh area, so it won't appear in visualization. I'll share a video.

kks32:
@bodhinandach Would you recommend deleting those particles when we can't locate? I see they appear in viz outside the domain.

bodhinandach:
@kks32 That's what I expect. Cause I think the vtk writer outputs for all particles, even outside the mesh. 

At the moment I think we can delete. I am thinking though that if we can reuse these particles somehow if inflow or injection is necessary. There, we don't have to delete and allocate memory (this might be costly and we need to broadcast this to all MPI nodes, I guess?). But as these features are not yet ready at the moment, I think deleting will be good. What do you think @kks32?

Thanks for checking btw!

cw646 (approved):
I gave this a trial run using velocities to push points out of the mesh, and it behaved as expected! Thanks!

PR #583, opened 2020-03-14
:computer: Write VTK stresses and strains as tensors

### LINKED ISSUE #582: Write stresses and strains as a tensor output
**Describe the feature**
Stresses and strains should include all components

**Describe alternatives**
VTK Array is a possibility, but I need more help in doing that as we can't have vector data with more than 3 components

**Additional context**
Stresses and strains only print normal components at the moment

kks32:
Fixes #582 

Removing `vtk_attributes`, we write all basic properties. Otherwise, it is adding a lot of code for not a lot of flexibility.

vtkArray doesn't work very well, so using `Tensor`.


bodhinandach:
@WeijianLIANG This might be interesting for you too

src/io/vtk_writer.cc
```diff
@@ -41,6 +41,51 @@ void VtkWriter::write_geometry(const std::string& filename) {
...
+  // Create an array to hold distance information
+  auto tensordata = vtkSmartPointer<vtkDoubleArray>::New();
+  tensordata->SetNumberOfComponents(9);
+  tensordata->SetNumberOfTuples(pdata->GetNumberOfPoints());
+  tensordata->SetName(data_field.c_str());
+
+  // Evaluate the signed distance function at all of the grid points
+  for (vtkIdType id = 0; id < pdata->GetNumberOfPoints(); ++id) {
+    const double* vdata = data.at(id).data();
+    tensordata->InsertTuple9(id, vdata[0], vdata[3], vdata[4], vdata[3],
```
bodhinandach: I think there is an error here. Here, you are putting vdata[4] in the x-z direction. However, in our code, vdata[5] should be in x-z direction as:
https://github.com/cb-geo/mpm/blob/72cd3842961f89962ac66b87b823c5674f9ce2f9/include/particles/particle.tcc#L546-L554

tests/io/vtk_writer_test.cc
```diff
@@ -4,16 +4,49 @@
...
 TEST_CASE("VTK Writer is checked", "[vtk][writer]") {
 
+  SECTION("Check scalar/vector/tensor file ") {
+    std::vector<Eigen::Matrix<double, 3, 1>> coordinates;
+    // VTK PolyData writer
+    auto vtk_writer = std::make_unique<VtkWriter>(coordinates);
+
+    SECTION("Write scalar data") {
+      const std::string vtk_file = "scalar_vtk.vtp";
+      const std::string attribute = "mass";
```
bodhinandach: Can you really output mass as the example? I see that `write_scalar_point_data` only allow output of variables inside the `state_vars` and mass or volume or material id, for instance, are not there.
https://github.com/cb-geo/mpm/blob/72cd3842961f89962ac66b87b823c5674f9ce2f9/include/solvers/mpm_base.tcc#L526-L527

bodhinandach (changes requested):
@kks32 I detected some possible bug. I will have another look in the afternoon.

src/io/vtk_writer.cc
kks32: Sorry my bad, I made a mistake when entering these values in the matrix

tests/io/vtk_writer_test.cc
kks32: Will do "epds"

bodhinandach: How about all the other parameters outside `state_vars`? Any thoughts? Cause I always hardcoded `particle.tcc` to output that, we might need to think about it...

kks32: This function is purely for `state_vars`, if we need others we'd have to do something similar like we have in `vector_data`. Probably a different PR

ezrayst:
I am OK with this implementation, generally!

include/solvers/mpm_base.tcc
```diff
@@ -497,13 +474,16 @@ void mpm::MPMBase<Tdim>::write_vtk(mpm::Index step, mpm::Index max_steps) {
   MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
 #endif
 
+  //! VTK vector variables
+  std::vector<std::string> vtk_vector_data = {"displacements", "velocities"};
```
ezrayst, line 478: Should we add `accelerations` as the default?

kks32: `accelerations` data is not stored at the particle level.

ezrayst: My bad, yes!

tests/io/vtk_writer_test.cc
bodhinandach: I see... That's okay. In that case, then all the particle private variables should be also included there too perhaps? Let's discuss this further.

bodhinandach (approved):

ezrayst (approved):

PR #585, opened 2020-03-17
[Constitutive Law] Minor Mohr-Coulomb

bodhinandach:
As you update the final stress through iteration, the state variables should also be updated.

### DIFF
--- a/include/materials/mohr_coulomb.tcc
+++ b/include/materials/mohr_coulomb.tcc
@@ -63,8 +63,6 @@ mpm::dense_map mpm::MohrCoulomb<Tdim>::initialise_state_variables() {
                                // Cohesion
                                {"cohesion", this->cohesion_peak_},
                                // Stress invariants
-                               // Equivalent (Von-Misses) stress
-                               {"evms", 0.},
                                // j3
                                {"j3", 0.},
                                // Epsilon
@@ -114,8 +112,6 @@ bool mpm::MohrCoulomb<Tdim>::compute_stress_invariants(
                         6.0 +
                     std::pow(stress(3), 2) + std::pow(stress(4), 2) +
                     std::pow(stress(5), 2);
-  // Compute equivalent (Von-Mises) stress
-  (*state_vars).at("evms") = std::sqrt(3 * j2);
   // Compute J3
   (*state_vars).at("j3") = dev_stress(0) * dev_stress(1) * dev_stress(2) -
                            dev_stress(2) * std::pow(dev_stress(3), 2) +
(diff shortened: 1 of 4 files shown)


include/materials/mohr_coulomb.tcc
```diff
@@ -112,6 +114,8 @@ bool mpm::MohrCoulomb<Tdim>::compute_stress_invariants(
                         6.0 +
                     std::pow(stress(3), 2) + std::pow(stress(4), 2) +
                     std::pow(stress(5), 2);
+  // Compute equivalent (Von-Mises) stress
+  (*state_vars).at("evms") = std::sqrt(3 * j2);
```
tianchiTJ: It's seems the same with "q", which may be used more often as the name

bodhinandach: I am open to the naming. But can we make it in the next PR where we clean up the entire existing laws?

tianchiTJ: I know, but I just use the same parameter "evms" as "q" in MCC.

tianchiTJ (approved):

include/materials/mohr_coulomb.tcc
```diff
@@ -409,22 +414,22 @@ Eigen::Matrix<double, 6, 1> mpm::MohrCoulomb<Tdim>::compute_stress(
...
-      (*state_vars)["cohesion"] =
+      (*state_vars).at("cohesion") =
           cohesion_residual_ + ((cohesion_peak_ - cohesion_residual_) *
                                 (pdstrain - pdstrain_residual_) /
                                 (pdstrain_peak_ - pdstrain_residual_));
     } else {
-      (*state_vars)["phi"] = phi_residual_;
-      (*state_vars)["psi"] = psi_residual_;
-      (*state_vars)["cohesion"] = cohesion_residual_;
+      (*state_vars).at("phi") = phi_residual_;
```
kks32, line 426: Do we need these `.at` checks? they would be checked and initialized at the start. Contracts in C++2x standard will remove this idea of throwing an error with `at`, it's considered an irrecoverable error at this point. I'd just update it using `[]` instead of `at`

include/materials/mohr_coulomb.tcc
ezrayst: I think the `state_vars` should only contain state variables that need to be passed to the next time step. `p` and `q` or von mises stresses can be computed, and thus should not be part of the `state_vars`

kks32: Agree with @ezrayst Since we don't need Von-Mises stresses to be calculated, I don't think it's good to have them stored in `state_vars`. If you want to plot, you can use the calculator function in Paraview or HDF5 to compute these.

include/materials/mohr_coulomb.tcc
bodhinandach: It's a clean up, solved as discussed privately

kks32 (approved):

PR #586, opened 2020-03-17
[Constitutive Law] Incompressible Bingham

bodhinandach:
Add incompressibility multiplier. By default, it is the same as the current version. You can combine this with the branch `solver/navier-stokes`

### DIFF
--- a/include/materials/bingham.tcc
+++ b/include/materials/bingham.tcc
@@ -15,6 +15,13 @@ mpm::Bingham<Tdim>::Bingham(unsigned id, const Json& material_properties)
     // Calculate bulk modulus
     bulk_modulus_ = youngs_modulus_ / (3.0 * (1. - 2. * poisson_ratio_));
 
+    // Special material properties
+    if (material_properties.contains("incompressible")) {
+      bool incompressible =
+          material_properties.at("incompressible").template get<bool>();
+      if (incompressible) compressibility_multiplier_ = 0.0;
+    }
+
     properties_ = material_properties;
   } catch (Json::exception& except) {
     console_->error("Material parameter not set: {} {}\n", except.what(),
@@ -83,14 +90,17 @@ Eigen::Matrix<double, 6, 1> mpm::Bingham<Tdim>::compute_stress(
 
   // Update pressure
   (*state_vars).at("pressure") +=
-      this->thermodynamic_pressure(ptr->dvolumetric_strain());
+      (compressibility_multiplier_ *
+       this->thermodynamic_pressure(ptr->dvolumetric_strain()));
 
   // Update volumetric and deviatoric stress
   // thermodynamic pressure is from material point
   // stress = -thermodynamic_pressure I + tau, where I is identity matrix or
   // direc_delta in Voigt notation
   const Eigen::Matrix<double, 6, 1> updated_stress =
-      -(*state_vars).at("pressure") * this->dirac_delta() + tau;
+      -(*state_vars).at("pressure") * this->dirac_delta() *
+          compressibility_multiplier_ +
+      tau;
 
   return updated_stress;
 }
--- a/include/materials/bingham.h
+++ b/include/materials/bingham.h
@@ -80,6 +80,8 @@ class Bingham : public Material<Tdim> {
   double mu_{std::numeric_limits<double>::max()};
   //! Critical yielding shear rate
   double critical_shear_rate_{std::numeric_limits<double>::max()};
+  //! Compressibility multiplier
+  double compressibility_multiplier_{1.0};
 
 };  // Bingham class
 }  // namespace mpm
--- a/tests/materials/bingham_test.cc
+++ b/tests/materials/bingham_test.cc
@@ -360,6 +360,105 @@ TEST_CASE("Bingham is checked in 2D", "[material][bingham][2D]") {
     REQUIRE(state_vars.at("pressure") ==
             Approx(-K * volumetric_strain).epsilon(Tolerance));
   }
+
+  SECTION("Bingham check stresses incompressible, with strain rate, yielded") {
+    unsigned id = 0;
+    // Initialise material
+    Json jmaterial;
+    jmaterial["density"] = 1000.;
+    jmaterial["youngs_modulus"] = 1.0E+7;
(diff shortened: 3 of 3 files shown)


kks32 (approved):

include/materials/bingham.tcc
```diff
@@ -15,6 +15,13 @@ mpm::Bingham<Tdim>::Bingham(unsigned id, const Json& material_properties)
     // Calculate bulk modulus
     bulk_modulus_ = youngs_modulus_ / (3.0 * (1. - 2. * poisson_ratio_));
 
+    // Special material properties
+    if (material_properties.contains("incompressible")) {
```
ezrayst, line 19: I think we can make a `incompressible` boolean to be a compulsory input parameter just like the rest. What do you think?

ezrayst (changes requested):

include/materials/bingham.tcc
```diff
@@ -83,14 +90,17 @@ Eigen::Matrix<double, 6, 1> mpm::Bingham<Tdim>::compute_stress(
...
+       this->thermodynamic_pressure(ptr->dvolumetric_strain()));
 
   // Update volumetric and deviatoric stress
   // thermodynamic pressure is from material point
   // stress = -thermodynamic_pressure I + tau, where I is identity matrix or
   // direc_delta in Voigt notation
   const Eigen::Matrix<double, 6, 1> updated_stress =
-      -(*state_vars).at("pressure") * this->dirac_delta() + tau;
+      -(*state_vars).at("pressure") * this->dirac_delta() *
+          compressibility_multiplier_ +
```
ezrayst, line 102: I don't think you need to multiply here again, given that the `pressure` will be zero.

bodhinandach: It's not zero. It's computed by the implicit solver for NS. So you need to have it here.

include/materials/bingham.tcc
kks32: It's only needed for NS solver, so I recommend leaving it as is

include/materials/bingham.tcc
kks32: I'd leave the multiplier in case there is some initial pressure.

ezrayst: I am OK with multiplying then.

include/materials/bingham.tcc
ezrayst: I am fine, but we have never done special input parameters for any `material` model. I just feel it's clearer if we have only compulsory input parameters. Less `if` statements too.

ezrayst (approved):

kks32:
@bodhinandach could you add unit tests?

cw646:
```
//! Compute stress
template <unsigned Tdim>
Eigen::Matrix<double, 6, 1> mpm::Bingham<Tdim>::compute_stress(
    const Vector6d& stress, const Vector6d& dstrain,
    const ParticleBase<Tdim>* ptr, mpm::dense_map* state_vars) {

  // Get strain rate
  auto strain_rate = ptr->strain_rate();

  // Convert strain rate to rate of deformation tensor
  strain_rate.tail(3) *= 0.5;

  // Set threshold for minimum critical shear rate
  const double shear_rate_threshold = 1.0E-15;
  if (critical_shear_rate_ < shear_rate_threshold)
    critical_shear_rate_ = shear_rate_threshold;

  // Rate of shear = sqrt(2 * D_ij * D_ij)
  // Since D (D_ij) is in Voigt notation (D_i), and the definition above is in
  // matrix, the last 3 components have to be doubled D_ij * D_ij = D_0^2 +
  // D_1^2 + D_2^2 + 2*D_3^2 + 2*D_4^2 + 2*D_5^2 Yielding is defined: rate of
  // shear > critical_shear_rate_^2 Checking yielding from strain rate vs
  // critical yielding shear rate
  double shear_rate =
      std::sqrt(2. * (strain_rate.dot(strain_rate) +
                      strain_rate.tail(3).dot(strain_rate.tail(3))));

  // Apparent_viscosity maps shear rate to shear stress

  double apparent_viscosity = 0.;

  Eigen::Matrix<double, 6, 1> tau;

  if (shear_rate * shear_rate > critical_shear_rate_ *critical_shear_rate_) {

    apparent_viscosity = 2. * ((tau0_ / shear_rate) + mu_);

    tau = apparent_viscosity * strain_rate;

  } else {

    apparent_viscosity = 5000.; // TODO should be viscosity*1000

    tau = (2 * apparent_viscosity) * strain_rate;
  }

  // Update pressure
  (*state_vars).at("pressure") +=
      this->thermodynamic_pressure(ptr->dvolumetric_strain());

  // Update volumetric and deviatoric stress
  // thermodynamic pressure is from material point
  // stress = -thermodynamic_pressure I + tau, where I is identity matrix or
  // direc_delta in Voigt notation
  const Eigen::Matrix<double, 6, 1> updated_stress =
      -(*state_vars).at("pressure") * this->dirac_delta() + tau;

  return updated_stress;
}
```

This is the simplified Bingham model - it's closer to the original paper the method was taken from - could you possibly adapt this with the incompressibility changes?

bodhinandach:
@cw646 sure, you can do similar things. Just multiply the compressibility_multiplier_ to the line:
```
(*state_vars).at("pressure") += compressibility_multiplier_ * this->thermodynamic_pressure(ptr->dvolumetric_strain());
```
and to:
```
const Eigen::Matrix<double, 6, 1> updated_stress = -(*state_vars).at("pressure") * this->dirac_delta() * compressibility_multiplier_ + tau;
```
Basically the modification does not change the constitutive model, just to separate the calculation of stresses from pressure which account for incompressibility constraints. If you want to try it in NS solver, we can have a chat.

cw646:
Sorry, I meant could you swap the current implementation with the one I posted?

include/materials/bingham.tcc
bodhinandach: It's only 1 if ezra, if that's okay I will keep it this way. The Newtonian is also modified in similar way.

cw646 (approved):
Will add the CW changes to the Bingham model in separate pr

bodhinandach:
Tests are added, @cw646 will address his modification in the following PR

PR #590, opened 2020-03-31
:pencil: Run vtk test only if VTK is found

kks32:
VTK tests are only run when they are available


bodhinandach (commented):
@kks32 I actually found a bug recently related to vtk. After the modification of `particle_tensor_data`, we lose the ability to specify a variable to be printed out from `mpm.json` as we did before. Of course, we need to specify that variable while initializing particle in `initialize()`. The only thing to do at the moment is to hard-code it in mpm_base default variables, along with displacements and velocities. Any thoughts on this?

kks32:
We print all variables so we have removed the option to choose what to print from the post-processing option in `mpm.json`. You can still choose which specific state variable is needed for printing in `mpm.json`

tests/io/vtk_writer_test.cc
```diff
@@ -1,3 +1,4 @@
+#ifdef USE_MPI
```
cbgeo: This should be `USE_VTK`

cbgeo (changes requested):

cbgeo (approved):

PR #591, opened 2020-03-31
[Constitutives] MC J2 correction

bodhinandach:
The PR fixes the bug to compute the stress derivation of the j2 deviatoric stress invariants, which are included in the computation of `drho_dsigma`, `dj2_dsigma`, and `dj3_dsigma`. The 2.0 coefficient is missing to account for the symmetric of the off-diagonal components, i.e. S12=S21 -> S12+S21 = 2.0*S12.

### DIFF
--- a/include/materials/mohr_coulomb.tcc
+++ b/include/materials/mohr_coulomb.tcc
@@ -251,6 +251,9 @@ void mpm::MohrCoulomb<Tdim>::compute_df_dp(
   if (fabs(rho) > std::numeric_limits<int>::epsilon()) {
     // Compute dRho / dSigma
     drho_dsigma = 1. / rho * dev_stress;
+    drho_dsigma(3) *= 2.0;
+    drho_dsigma(4) *= 2.0;
+    drho_dsigma(5) *= 2.0;
     // Compute r value
     r_val = (3. * std::sqrt(6.)) * (j3 / std::pow(rho, 3));
     // Compute dr / dJ2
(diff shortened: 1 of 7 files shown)


tianchiTJ (approved):

kks32 (approved):

include/materials/mohr_coulomb.tcc
```diff
@@ -264,6 +267,9 @@ void mpm::MohrCoulomb<Tdim>::compute_df_dp(
   const double dtheta_dr = -1 / (3. * std::sqrt(divider));
   // Compute dJ2 / dSigma
   Vector6d dj2_dsigma = dev_stress;
+  dj2_dsigma(3) *= 2.0;
```
ezrayst, line 270: I am currently rechecking everything. So please wait for me.

bodhinandach: Sure, @ezrayst, take your time. Thanks for reconfirming.

tianchiTJ: Can we approve this request first, because this is real an error in computation.

bodhinandach: Ping @ezrayst

ezrayst: I can OK it, but maybe we need to revisit it again for sure. What do you think?

ezrayst (approved):

include/materials/mohr_coulomb.tcc
tianchiTJ: I mean at least we need to make it correct first and good for other users, and then improve it.

ezrayst: We don't know whether it's correct yet, but at least we can fix what is for sure wrong now. I approved it.

PR #595, opened 2020-04-07
Add range to sine function

thiagordonho:
Add range to the definition of a sine function thus allowing for the function to be specified only within this range. The value of the function is equal to zero at times that are outside of this range.

### DIFF
--- a/src/functions/sin_function.cc
+++ b/src/functions/sin_function.cc
@@ -4,18 +4,17 @@ mpm::SinFunction::SinFunction(unsigned id, const Json& properties)
     : mpm::FunctionBase(id, properties) {
   x0_ = properties.at("x0");
   a_ = properties.at("a");
-  if (properties.contains("xrange"))
+  if (properties.contains("xrange")) {
     if (properties.at("xrange").is_array() &&
         properties.at("xrange").size() == 2) {
       for (unsigned index = 0; index < 2; ++index)
         xrange_.at(index) = properties.at("xrange").at(index);
-    } else
-      throw std::runtime_error(
-          "Cannot initialise sine function; x range is invalid");
+    }
+  }
 }
 
 // Return f(x) for a given x
 double mpm::SinFunction::value(double x_input) const {
-  if ((x_input < xrange_[0]) || (x_input > xrange_[1])) return 0.0;
+  if ((x_input < xrange_.at(0)) || (x_input > xrange_.at(1))) return 0.0;
   return std::sin(a_ * (x_input - x0_));
 }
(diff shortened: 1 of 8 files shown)


kks32:
I have the following questions:

1. Is this to curtail the duration in which the sin function is applied?

2. Could you provide an example input file in your description?

3. Is the range function an optional argument?

4. could we have either just xmin/xmax without specifying the other?

src/functions/sin_function.cc
```diff
@@ -4,9 +4,17 @@ mpm::SinFunction::SinFunction(unsigned id, const Json& properties)
     : mpm::FunctionBase(id, properties) {
   x0_ = properties.at("x0");
   a_ = properties.at("a");
+  if (properties.at("xrange").is_array() &&
```
kks32: Could you make this an optional argument?

src/functions/sin_function.cc
```diff
@@ -4,9 +4,17 @@ mpm::SinFunction::SinFunction(unsigned id, const Json& properties)
...
+      properties.at("xrange").size() == 2) {
+    for (unsigned index = 0; index < 2; ++index)
+      xrange_[index] = properties.at("xrange").at(index);
```
kks32:
```suggestion
      xrange_.at(index) = properties.at("xrange").at(index);
```

kks32 (changes requested):

src/functions/sin_function.cc
```diff
@@ -4,13 +4,14 @@ mpm::SinFunction::SinFunction(unsigned id, const Json& properties)
...
   x0_ = properties.at("x0");
   a_ = properties.at("a");
-  if (properties.at("xrange").is_array() &&
-      properties.at("xrange").size() == 2) {
-    for (unsigned index = 0; index < 2; ++index)
-      xrange_[index] = properties.at("xrange").at(index);
-  } else
-    throw std::runtime_error(
-        "Cannot initialise sine function; x range is invalid");
+  if (properties.contains("xrange"))
```
kks32: Could you please add { } to define range of the if-condition

src/functions/sin_function.cc
```diff
@@ -4,9 +4,18 @@ mpm::SinFunction::SinFunction(unsigned id, const Json& properties)
...
+      for (unsigned index = 0; index < 2; ++index)
+        xrange_.at(index) = properties.at("xrange").at(index);
+    } else
+      throw std::runtime_error(
+          "Cannot initialise sine function; x range is invalid");
 }
 
 // Return f(x) for a given x
 double mpm::SinFunction::value(double x_input) const {
+  if ((x_input < xrange_[0]) || (x_input > xrange_[1])) return 0.0;
```
kks32: Please use .at()

kks32 (approved):

PR #597, opened 2020-04-09
[Materials] MohrCoulomb derivatives of R

ezrayst:
Not sure about the derivatives highlighted in this PR. 

If @tianchiTJ is OK, I would make the test cases too. Below is my document on the derivatives.

image


tianchiTJ:
I think there should be a negative sign in A.42, and others should be good in my opinion.

ezrayst:
That is another thing we need to discuss. I use sin convention for Lode Angle and that's why there is no negative. But you agree in the pull request? I will make the test cases then

tianchiTJ:
You equations are right, but the rho=sqrt(2J2), so I think the original code should be correct

ezrayst:
I see. Understood thanks!

Issue #587, opened 2020-03-17
Constitutive Laws Development and Optimization


bodhinandach:
I am creating this thread to discuss what we can generalize and improve furthers from the Constitutive Laws. So far my concerns are to:
- generalize state-variables naming, e.g. between epds and pdstrain
- create a common material_utility where functions like: compute stress invariants, principle stresses,  rotation transformation, tensor operators, etc. are implemented there.

If you have any other suggestions, we can discuss it here too. Feel free to add/tag anyone interested.

kks32:
We could use the `geometry` namespace for rotations? Maybe we have a separate `tensor` space for common Tensorial operations?

bodhinandach:
Are those namespaces are available now? Static functions work too, I am not sure which is better.

kks32:
Yes `geometry` is already available

ezrayst:
Finished with #600, do we need more? Otherwise we can close the issue I guess

kks32:
Thanks Ezra!

bodhinandach:
Thanks, Ezra, further functions can be added in the future. But as for now, it is good.

PR #600, opened 2020-04-15
Material utility operations namespace

ezrayst:
Response to #587. Computations and tests for:
- [x] mean p 
- [x] deviatoric stress
- [x] deviatoric q
- [x] J2
- [x] J3 
- [x] Lode angle theta
- [x] dp_dsigma
- [x] dq_dsigma
- [x] dj2_dsigma
- [x] dj3_dsigma
- [x] dtheta_dsigma

Upcoming work for next PR:
- [ ] Refactor NorSand
- [ ] Refactor Modified Cam Clay
- [ ] Refactor Mohr Coulomb


include/materials/tensor.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <cmath>
+
+#include "Eigen/Dense"
+
+namespace mpm {
+namespace tensor {
+
+//! Compute mean stress p
+//! \param[in] stress Stress in Voigt notation where positive is tension
```
kks32: Why is tension positive? In geotech applications most of our stresses are compressive

include/materials/tensor.h
```diff
@@ -0,0 +1,66 @@
+#ifndef MPM_TENSOR_H_
+#define MPM_TENSOR_H_
+
+#include <cmath>
+
+#include "Eigen/Dense"
+
+namespace mpm {
+namespace tensor {
```
kks32: We may have to reconsider the namespace naming, none of the operations are tensorial operations. Maybe just `material` or `invariants`

include/materials/tensor.tcc
```diff
@@ -0,0 +1,226 @@
+//! Compute mean stress p
+inline double mpm::tensor::compute_mean_p(
```
kks32: This should be const everything

include/materials/tensor.tcc
```diff
@@ -0,0 +1,226 @@
...
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute mean p in tension positive
+  double mean_p = 1. / 3. * (stress(0) + stress(1) + stress(2));
```
kks32: return this instead of creating a variable.

include/materials/tensor.tcc
```diff
@@ -0,0 +1,226 @@
...
+
+  return mean_p;
+}
+
+//! Compute deviatoric stress
+inline Eigen::Matrix<double, 6, 1> mpm::tensor::compute_deviatoric_stress(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute mean p in tension positive
+  double mean_p = compute_mean_p(stress);
```
kks32: const

include/materials/tensor.tcc
kks32: Almost all functions should be `const` everything

include/materials/tensor.tcc
```diff
@@ -0,0 +1,226 @@
...
+
+  // Define derivatives of R in terms of J2 and J3
+  double dR_dj2 = -9.0 / 4.0 * sqrt(3.0) * j3;
+  double dR_dj3 = 3.0 / 2.0 * sqrt(3.0);
+
+  // Compute derivative of theta in terms of R
+  double dtheta_dR = -1.0 / 3.0;
+
+  // Update when J2 is non zero
+  if (abs(j2) > 1.0E-6) {
```
kks32: Why 1.0E-6?

kks32 (changes requested):
Could you please make functions const, and variables that don't change as `const` as well.

include/materials/tensor.h
ezrayst: I am open to discussion. My thought is to keep this tension positive for the purpose of other functions like computing deviatoric stress. In most geotech application, then users can multiply by -1 separately in material functions?

include/materials/tensor.h
ezrayst: I am fine with `material`, `invariants` might seem a little too specific because this may grow bigger than just computing the `invariants`. Maybe others can comment what they think too.

include/materials/tensor.tcc
ezrayst: Alright, I will `const` everything.

include/materials/tensor.tcc
ezrayst: Noted

include/materials/tensor.tcc
ezrayst: We can change them of course, but I just chose this as a safe-guard. Note that since we are talking about Pa as unit for stress, it is rarely that we have stress so small that it would trigger this if condition. It usually only happens when we have zero stresses all throughout.

Should we change them to 1.E-16 instead?

include/materials/tensor.h
ezrayst: But then the name of the files would be `material` which would be the same as the current class `material`

bodhinandach: I vote for `material_utility`, basically, it can also contain other than invariants if future constitutive laws need it.

include/materials/tensor.h
kks32: The MPM code is mainly for Geotech applications, also I don't think we need to explicity define positive / negative

ezrayst: OK let's get only `mean_p` positive, but the rest to be the same definition

ezrayst: Done

include/materials/tensor.h
ezrayst: Done :) unless others have better names

include/materials/tensor.tcc
ezrayst: Done

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+    lode_angle_val = (3. * std::sqrt(3.) / 2.) * (j3 / std::pow(j2, 1.5));
+  }
+  if (lode_angle_val > 1.0) lode_angle_val = 1.0;
+  if (lode_angle_val < -1.0) lode_angle_val = -1.0;
+
+  // Compute Lode angle (sin convention)
+  double lode_angle = (1. / 3.) * acos(lode_angle_val);
+  if (lode_angle > M_PI / 3.) lode_angle = M_PI / 3.;
+  if (lode_angle < -0.) lode_angle = 0.;
+
```
tianchiTJ: I think no need to check it again if you have limited the lode_angle_val in [-1,1]

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  // Compute q
+  const double deviatoric_q = compute_deviatoric_q(stress);
+
+  // Compute deviatoric stress
+  Eigen::Matrix<double, 6, 1> dev_stress = compute_deviatoric_stress(stress);
+
+  // Compute dq / dsigma
+  Eigen::Matrix<double, 6, 1> dq_dsigma = Eigen::Matrix<double, 6, 1>::Zero();
+  if (abs(deviatoric_q) > 1.E-6) {
```
tianchiTJ: Do you think this tolerance is too large? Do we need to define a same tolerance for all denominator?

include/materials/material_utility.tcc
ezrayst: I see, but it's not harmful to have right?

include/materials/material_utility.tcc
ezrayst: I don't mind using 1.E-15 actually, but it will blow up more right?

include/materials/material_utility.tcc
bodhinandach: If conds are expensive, I second @tianchiTJ. It's always a good practice to minimize if :)

include/materials/material_utility.tcc
bodhinandach: maybe -8 or -9 to be in the middle?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  return dq_dsigma;
+}
+
+//! Compute derivative of J2 in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_dj2_dsigma(
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute deviatoric stress
+  Eigen::Matrix<double, 6, 1> dev_stress = compute_deviatoric_stress(stress);
```
bodhinandach: const?

include/materials/material_utility.h
```diff
@@ -0,0 +1,68 @@
...
+#include <cmath>
+
+#include "Eigen/Dense"
+
+namespace mpm {
+namespace material_utility {
+
+//! Compute mean stress p
+//! \param[in] stress Stress in Voigt notation where positive is tension
+inline const double compute_mean_p(const Eigen::Matrix<double, 6, 1>& stress);
```
bodhinandach: Can we include the details that p is assumed to be positive in compression here too?

bodhinandach (approved):
I am good with this, many thanks @ezrayst!

tianchiTJ (approved):

include/materials/material_utility.tcc
kks32: We should pass a default tolerance value

include/materials/material_utility.tcc
kks32: Why do we write to `deviatoric_stress` and then copy it to dj2/dsigma, why can't we assign directly?

include/materials/material_utility.h
```diff
@@ -0,0 +1,68 @@
...
+
+//! Compute deviatoric stress
+//! \param[in] stress Stress in Voigt notation where positive is tension
```
kks32, line 16: Please change this to positive is compression

include/materials/material_utility.h
kks32: Shall we rename the function as `p` and drop the `compute_` part from all the function names?

include/materials/material_utility.h
```diff
@@ -0,0 +1,68 @@
+#ifndef MPM_MATERIAL_UTILITY_H_
+#define MPM_MATERIAL_UTILITY_H_
+
+#include <cmath>
+
+#include "Eigen/Dense"
```
kks32:
```suggestion
#include "data_types.h"
```
We can use that as the default include

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
+//! Compute mean stress p
+inline const double mpm::material_utility::compute_mean_p(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+  // Compute mean p in compresion positive
```
kks32: Could we move this comment to the top, as this isn't adding any additional value here.

include/materials/material_utility.h
kks32: I'm in two minds about the namespace `material_utility` instead of just `materials` because every time we use it, we have to do `mpm::material_utility::compute_` instead of `mpm::material::mean_p`, which would be more succinct to read.

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  return (-1. / 3. * (stress(0) + stress(1) + stress(2)));
+}
+
+//! Compute deviatoric stress
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_deviatoric_stress(
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute mean p in tension positive
```
kks32: Could we please have compression positive?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  const double mean_p = -1.0 * compute_mean_p(stress);
+
+  // Compute deviatoric by subtracting volumetric part
+  Eigen::Matrix<double, 6, 1> dev_stress = stress;
```
kks32: Could you please rename to deviatoric stress

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+//! Compute J3 invariant
+inline const double mpm::material_utility::compute_j3(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute deviatoric stress
+  Eigen::Matrix<double, 6, 1> dev_stress = compute_deviatoric_stress(stress);
+
+  // Compute J3
+  double j3 = (dev_stress(0) * dev_stress(1) * dev_stress(2)) -
```
kks32: const double

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+}
+
+//! Compute deviatoric q
+inline const double mpm::material_utility::compute_deviatoric_q(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute J2 from
+  const double j2 = compute_j2(stress);
+
+  const double deviatoric_q = std::sqrt(3 * j2);
```
kks32: No need to define a new variable, return `std::sqrt(3*j2)

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  return deviatoric_q;
+}
+
+//! Compute Lode angle
+inline const double mpm::material_utility::compute_lode_angle(
+    const Eigen::Matrix<double, 6, 1>& stress) {
```
kks32: I'd suggest passing a default tolerance value

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  // Compute j2 and j3
+  const double j2 = compute_j2(stress);
+  const double j3 = compute_j3(stress);
+
+  // Compute Lode angle value
+  double lode_angle_val = 0.0;
+  if (abs(j2) > 1.0E-6) {
```
kks32:
```suggestion
  if (std::abs(j2) > tolerance) {
```

include/materials/material_utility.tcc
kks32: If we have checked and restricted the values to be between -1 and 1, why should we check again?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  return lode_angle;
+}
+
+//! Compute derivative of p in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_dp_dsigma(
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  Eigen::Matrix<double, 6, 1> dp_dsigma = Eigen::Matrix<double, 6, 1>::Zero();
```
kks32, line 87: Please use the data type function `zero()`

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  dp_dsigma(1) = 1. / 3.;
+  dp_dsigma(2) = 1. / 3.;
+
+  return dp_dsigma;
+}
+
+//! Compute derivative of q in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_dq_dsigma(
+        const Eigen::Matrix<double, 6, 1>& stress) {
```
kks32: I recommend passing a default argument value for tolerance

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  // Compute q
+  const double deviatoric_q = compute_deviatoric_q(stress);
+
+  // Compute deviatoric stress
+  Eigen::Matrix<double, 6, 1> dev_stress = compute_deviatoric_stress(stress);
+
+  // Compute dq / dsigma
+  Eigen::Matrix<double, 6, 1> dq_dsigma = Eigen::Matrix<double, 6, 1>::Zero();
```
kks32, line 106: Please use data_type function zero

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+//! Compute derivative of q in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_dq_dsigma(
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute q
+  const double deviatoric_q = compute_deviatoric_q(stress);
+
+  // Compute deviatoric stress
+  Eigen::Matrix<double, 6, 1> dev_stress = compute_deviatoric_stress(stress);
```
kks32: do we need to specify type here? const auto?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  // Compute dq / dsigma
+  Eigen::Matrix<double, 6, 1> dq_dsigma = Eigen::Matrix<double, 6, 1>::Zero();
+  if (abs(deviatoric_q) > 1.E-6) {
+    dq_dsigma(0) = 3. / 2. / deviatoric_q * dev_stress(0);
```
kks32: Could you please use parenthesis continuous division is confusing to read.

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+  Eigen::Matrix<double, 3, 1> dev2;
+  dev2(0) = dev_stress(3);
+  dev2(1) = dev_stress(1);
+  dev2(2) = dev_stress(4);
+  Eigen::Matrix<double, 3, 1> dev3;
+  dev3(0) = dev_stress(5);
+  dev3(1) = dev_stress(4);
+  dev3(2) = dev_stress(2);
+
+  Eigen::Matrix<double, 6, 1> dj3_dsigma = Eigen::Matrix<double, 6, 1>::Zero();
```
kks32, line 156: use data_type function zero

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+//! Compute derivative of J3 in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1>
+    mpm::material_utility::compute_dj3_dsigma(
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute J2
+  const double j2 = compute_j2(stress);
+
+  // Compute deviatoric stress
+  const Eigen::Matrix<double, 6, 1> dev_stress =
```
kks32: const auto

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+        const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute J2
+  const double j2 = compute_j2(stress);
+
+  // Compute J3
+  const double j3 = compute_j3(stress);
+
+  // Compute dj2_dsigma
+  const Eigen::Matrix<double, 6, 1> dj2_dsigma = compute_dj2_dsigma(stress);
```
kks32: const auto

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  // Compute dj3_dsigma
+  const Eigen::Matrix<double, 6, 1> dj3_dsigma = compute_dj3_dsigma(stress);
```
kks32: const auto

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+
+  // Define derivatives of R in terms of J2 and J3
+  double dR_dj2 = -9.0 / 4.0 * sqrt(3.0) * j3;
+  double dR_dj3 = 3.0 / 2.0 * sqrt(3.0);
+
+  // Compute derivative of theta in terms of R
+  double dtheta_dR = -1.0 / 3.0;
+
+  // Update when J2 is non zero
+  if (abs(j2) > 1.0E-6) {
```
kks32: std::abs(j2) > tolerance

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,230 @@
...
+    // Update R
+    R = j3 / 2.0 * std::pow(j2 / 3.0, -1.5);
```
kks32: please only use lower case letters for variables

tests/materials/material_utility_test.cc
```diff
@@ -0,0 +1,213 @@
+#include <cmath>
```
kks32: You don't need to include all of these again if you are including material_utility.h

tests/materials/material_utility_test.cc
```diff
@@ -0,0 +1,213 @@
...
+#include <limits>
+#include <memory>
+
+#include "Eigen/Dense"
+#include "catch.hpp"
+
+#include "material_utility.h"
+
+//! \brief Check material_utility class for 3D case
+TEST_CASE("Material utility is checked for 3D case", "[material_utility][3D]") {
```
kks32: Why specifically is it a 3D case?

kks32 (changes requested):
Could you please also address comments raised by @bodhinandach and @tianchiTJ

include/materials/material_utility.tcc
ezrayst: Will do so

include/materials/material_utility.h
ezrayst: I see, ok

include/materials/material_utility.tcc
ezrayst: OK

include/materials/material_utility.h
ezrayst: Wait, the input stress is always positive tension

tests/materials/material_utility_test.cc
ezrayst: OK

include/materials/material_utility.tcc
ezrayst: OK

include/materials/material_utility.tcc
ezrayst: I can use const auto. I think it's better

include/materials/material_utility.h
ezrayst: I understand. Should the filename be changed too? I will keep the filename as is for now

bodhinandach: material is a bit off I guess. I am still for material utility. Btw, @kks32 for the namespace are using CamelCase or snake_case?

include/materials/material_utility.tcc
bodhinandach: @kks32 isn't better to create `mpm::tolerance()` so that by default all tolerance is the same by default? Or we can utilize `std::numeric_limits<double>::epsilon()`

include/materials/material_utility.tcc
ezrayst: OK, it has been changed

tests/materials/material_utility_test.cc
ezrayst: I guess there is no 3D or 2D in this case, everything is generalized as 3D. I will delete them.

include/materials/material_utility.tcc
ezrayst: Ok, I have removed it.

include/materials/material_utility.tcc
ezrayst: I have used the latter actually `std::numeric_limits<double>::epsilon()`

ezrayst:
I think I have addressed all the comments. Now, the namespace is called `material` but the file is still called `material_utility`. 

Throwing some ideas here: `material_functions`, `material_operations`, `materials` (with s) if you want something short.

include/materials/material_utility.h
kks32: I'm really sorry, I got this mixed up. You are right, stresses we get are positive in tension. Let's be consistent and have mean stress be also positive in tension. Sorry about that!! I was incorrectly looking at the magnitude not a component.

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+}
+
+//! Compute deviatoric q
+inline const double mpm::material::q(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute J2 from
+  const double j2 = mpm::material::j2(stress);
+
+  return (std::sqrt(3 * j2));
```
kks32, line 60: Could we just write out a formula instead of a function call?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+  double lode_angle = (1. / 3.) * acos(lode_angle_val);
+
+  return lode_angle;
+}
+
+//! Compute derivative of p in terms of stress sigma
+inline const Eigen::Matrix<double, 6, 1> mpm::material::dp_dsigma(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  Eigen::Matrix<double, 6, 1> dp_dsigma;
```
kks32: Using set zero will be 2 steps every time, the way you had it was how we want it, but instead use the `zero<6>()` to initialize them to zeros.

kks32: This function is available in `data_types.h`, you may want to specialize for `6`

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+
+  // Compute dj2_dsigma
+  const auto dj2_dsigma = mpm::material::dj2_dsigma(stress);
+
+  // Compute dj3_dsigma
+  const auto dj3_dsigma = mpm::material::dj3_dsigma(stress);
+
+  // Declare R as zero to avoid division by zero J2
+  // R is defined as R = cos(3 theta)
+  double r = 0.0;
```
kks32: Why are we declaring r here? Couldn't we declare it line 204?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+
+  // Update when J2 is non zero
+  if (std::abs(j2) > std::numeric_limits<double>::epsilon()) {
+    // Update R
+    r = j3 / 2.0 * std::pow(j2 / 3.0, -1.5);
+    // Update derivatives of R
+    dr_dj2 *= std::pow(j2, -2.5);
+    dr_dj3 *= std::pow(j2, -1.5);
+    // Update derivative of theta in terms of R, check for sqrt of zero
+    if (std::abs(1 - r * r) < std::numeric_limits<double>::epsilon()) {
```
kks32: const double factor = (std::abs(1 - r * r) < std::numeric_limits<double>::epsilon()) ?  std::numeric_limits<double>::epsilon() : (1 - r*r);
dtheta_dr = -1.0/(3*sqrt(factor));

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+inline const double mpm::material::lode_angle(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute j2 and j3
+  const double j2 = mpm::material::j2(stress);
+  const double j3 = mpm::material::j3(stress);
+
+  // Compute Lode angle value
+  double lode_angle_val = 0.0;
+  if (std::abs(j2) > std::numeric_limits<double>::epsilon()) {
```
kks32: Isn't this such a small value? Why can't we set a default tolerance, which constitutive modelers can modify?

kks32 (changes requested):

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,221 @@
...
+    } else {
+      dtheta_dr = -1.0 / (3.0 * sqrt(1 - r * r));
+    }
+  }
+
+  // Compute dtheta / dsigma
+  const Eigen::Matrix<double, 6, 1> dtheta_dsigma =
+      dtheta_dr * ((dr_dj2 * dj2_dsigma) + (dr_dj3 * dj3_dsigma));
+
+  return dtheta_dsigma;
```
kks32: There is no need for `dtheta_dsigma`, return directly

include/materials/material_utility.tcc
ezrayst: That defeats the purpose of having this PR right, where we want to reduce repetition of codes, and easier to maintain.

include/materials/material_utility.tcc
ezrayst: We can't because r is zero when j2 is zero (division by zero) so the same as the next three variables, we need to declare here and then update

include/materials/material_utility.tcc
kks32: function calls are expensive, and this is a single line function for J2, if this gets called a lot, it's a lot more expensive!

include/materials/material_utility.tcc
kks32: But you are not using `r` anywhere else outside the loop starting in line 202. Or am I missing something?

ezrayst: Sorry, we can!

include/materials/material_utility.tcc
ezrayst: Understood

include/materials/material_utility.tcc
ezrayst: Yes, it is such a small value and not likely to happen except almost zero stress (j2 is a squared term). But not all constitutive models have tolerance.

kks32: If we have a tolerance value as a default argument, then modelers needing a tolerance can use it, others can ignore. Any reason, why we shouldn't do this?

include/materials/material_utility.tcc
ezrayst: I see! Give me a moment

ezrayst: I updated the `data_types.h` but is this how I use it?
`Eigen::Matrix<double, 6, 1> dp_dsigma = mpm::zero<6>();`? (It does not work)

include/materials/material_utility.tcc
ezrayst: Another way is to clump the different functions together, so all the other functions that need j2 would be in the same call. I feel if we repeat the single line function, it would be hard to maintain.

ezrayst:
@kks32, how about the namespace name? should we keep them as `material` and the filenames as `material_utility`?

include/materials/material_utility.tcc
ezrayst: I don't mind actually, I will add that.

kks32:
I'd go with `materials` for namespace and keep the filename as `material_utilities.h`?

ezrayst:
Understood.

include/materials/material_utility.tcc
ezrayst: I am still on the fence about this. Does it mean that we have to copy the function (one line) for both j2 and j3?

include/materials/material_utility.tcc
```diff
@@ -0,0 +1,271 @@
...
+  // Compute and return mean p
+  return (1. / 3. * (stress(0) + stress(1) + stress(2)));
+}
+
+//! Compute deviatoric stress
+inline const Eigen::Matrix<double, 6, 1> mpm::materials::deviatoric_stress(
+    const Eigen::Matrix<double, 6, 1>& stress) {
+
+  // Compute mean p
+  const double p = mpm::materials::p(stress);
```
tianchiTJ, line 12: I found some paper use p=0.5*(sigma1+sigma3) and q=0.5*(sigma1-sigma3) for 2D. I think q is ok, and the current equation can represent the same result, but the p may be different. Sorry I'm not sure about this part, do you know which one is better?

kks32: Hi @tianchiTJ those are MIT definition of mean and deviatoric stress. I suggest we stick to octohedral stress definition in 3D.

tianchiTJ: OK, got it

include/materials/material_utility.tcc
bodhinandach: I second @kks32

ezrayst: I have overloaded the functions

include/data_types.h
```diff
@@ -16,19 +16,25 @@ using Index = unsigned long long;
 template <typename Ttype>
```
kks32: If you are not using this file, there is no need to change this.

include/materials/material_utility.tcc
kks32: Set a default argument, please don't overload the functions

kks32 (changes requested):

include/materials/material_utility.h
```diff
@@ -29,15 +29,12 @@ inline const double j3(const Eigen::Matrix<double, 6, 1>& stress);
...
-inline const double lode_angle(const Eigen::Matrix<double, 6, 1>& stress);
-
 //! Compute Lode angle theta (cosine convention)
 //! \param[in] stress Stress in Voigt notation where positive is tension
 //! \param[in] tolerance Default tolerance value specified by user
-inline const double lode_angle(const Eigen::Matrix<double, 6, 1>& stress,
-                               const double tolerance);
+inline const double lode_angle(
+    const Eigen::Matrix<double, 6, 1>& stress,
+    const double tolerance = std::numeric_limits<double>::epsilon());
```
kks32: Please don't const double in input argument or anything of size double or lower: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

include/materials/material_utility.h
```diff
@@ -59,16 +56,12 @@ inline const Eigen::Matrix<double, 6, 1> dj2_dsigma(
...
-//! \param[in] stress Stress in Voigt notation where positive is tension
-inline const Eigen::Matrix<double, 6, 1> dtheta_dsigma(
-    const Eigen::Matrix<double, 6, 1>& stress);
-
 //! Compute derivative of Lode angle theta in terms of stress sigma
 //! \param[in] stress Stress in Voigt notation where positive is tension
 //! \param[in] tolerance Default tolerance value specified by user
 inline const Eigen::Matrix<double, 6, 1> dtheta_dsigma(
-    const Eigen::Matrix<double, 6, 1>& stress, const double tolerance);
+    const Eigen::Matrix<double, 6, 1>& stress,
```
kks32, line 63: Please avoid const here in tolerance

include/materials/material_utility.h
```diff
@@ -0,0 +1,71 @@
...
+#include <cmath>
+
+#include "data_types.h"
+
+namespace mpm {
+namespace materials {
+
+//! Compute mean stress p (tension positive)
+//! \param[in] stress Stress in Voigt notation where positive is tension
+inline const double p(const Eigen::Matrix<double, 6, 1>& stress);
```
kks32: Please avoid returning by const value for built-in data types. 
```you are promising that the original variable (inside the function frame) will not be modified. And again, because you’re returning it by value, it’s copied so the original value could never be modified via the return value. For built-in types, it doesn’t matter whether you return by value as a const, so you should avoid confusing the client programmer and leave off the const when returning a built-in type by value.```

For more info: https://www.linuxtopia.org/online_books/programming_books/thinking_in_c++/Chapter08_014.html

File: include/materials/material_utility.tcc [hunk 44]
```diff
@@ -0,0 +1,209 @@
+//! Compute mean stress p (compression positive)
```
kks32:
```suggestion
//! Compute mean stress p (tension is positive)
```

include/materials/material_utility.h
kks32: Keep const for `Eigen` data types

kks32 (changes requested):

include/materials/material_utility.h
ezrayst: I see. Understood.

kks32 (approved):

PR #603, opened 2020-04-23
Refactor NorSand to include material_utility

ezrayst:


include/materials/norsand.h
```diff
@@ -69,12 +69,12 @@ class NorSand : public Material<Tdim> {
...
 
-  //! Compute stress invariants (p, q, j2, j3, lode_angle and M_theta)
+  //! Compute stress invariants (p, q, lode_angle and M_theta)
   //! \param[in] stress Stress
   //! \param[in] state_vars History-dependent state variables
-  //! \retval vector of size six, containing p, q, j2, j3, lode_angle and
+  //! \retval vector of size six, containing p, q, lode_angle and
   //! M_theta
-  Eigen::Matrix<double, 6, 1> compute_stress_invariants(
+  Eigen::Matrix<double, 4, 1> compute_stress_invariants(
```
kks32: If you need 4 different return values, could you please pass them as a pointer instead of returning an Eigen Matrix.

include/materials/norsand.tcc
```diff
@@ -49,6 +48,8 @@ mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
     m_dilation_ = material_properties.at("m_dilation").template get<double>();
     // Parameter for shear modulus
     m_modulus_ = material_properties.at("m_modulus").template get<double>();
+    // Default tolerance
+    tolerance_ = material_properties.at("tolerance").template get<double>();
```
kks32: Could we have this as an optional argument?

include/materials/norsand.tcc
```diff
@@ -114,54 +115,29 @@ bool mpm::NorSand<Tdim>::compute_elastic_tensor() {
...
 template <unsigned Tdim>
-Eigen::Matrix<double, 6, 1> mpm::NorSand<Tdim>::compute_stress_invariants(
+Eigen::Matrix<double, 4, 1> mpm::NorSand<Tdim>::compute_stress_invariants(
     const Vector6d& stress, mpm::dense_map* state_vars) {
 
   // Note that in this subroutine, stress is compression positive
 
   // Compute mean stress p
-  double mean_p = (stress(0) + stress(1) + stress(2)) / 3.;
+  double mean_p = -1. * mpm::materials::p(-stress);
```
kks32:
```suggestion
const  double mean_p = check_low(-1. * mpm::materials::p(-stress));
```

include/materials/norsand.tcc
```diff
@@ -114,54 +115,29 @@ bool mpm::NorSand<Tdim>::compute_elastic_tensor() {
...
-               std::pow((stress(1) - stress(2)), 2) +
-               std::pow((stress(0) - stress(2)), 2)) /
-                  6.0 +
-              std::pow(stress(3), 2) + std::pow(stress(4), 2) +
-              std::pow(stress(5), 2);
-  j2 = check_low(j2);
-
   // Compute q
-  double deviatoric_q = std::sqrt(3 * j2);
+  double deviatoric_q = mpm::materials::q(-stress);
```
kks32:
```suggestion
  const double deviatoric_q = check_low(mpm::materials::q(-stress));
```

include/materials/norsand.tcc
```diff
@@ -203,22 +179,20 @@ void mpm::NorSand<Tdim>::compute_state_variables(
     // Compute and update void ratio image
     // e_image = e_max_ - (e_max_ - e_min_) / log(crushing_pressure_ / p_image);
     e_image = gamma_ - lambda_ * log(p_image / reference_pressure_);
```
kks32:
```suggestion
    e_image = check_low(gamma_ - lambda_ * log(p_image / reference_pressure_));
```

kks32 (changes requested):

include/materials/norsand.tcc
ezrayst: Yes

include/materials/norsand.h
ezrayst: Can you comment on the current design? I still don't pass the test function which means something is wrong

include/materials/norsand.tcc
```diff
@@ -135,8 +135,7 @@ Eigen::Matrix<double, 4, 1> mpm::NorSand<Tdim>::compute_stress_invariants(
   double M_theta = Mtc_ - std::pow(Mtc_, 2) / (3. + Mtc_) * cos_lode_angle;
 
   // Store to return
-  Eigen::Matrix<double, 4, 1> invariants;
-  invariants << mean_p, deviatoric_q, lode_angle, M_theta;
+  static double invariants[4] = {mean_p, deviatoric_q, lode_angle, M_theta};
```
ezrayst, line 138: @kks32 I am returning as array pointer of double

include/materials/norsand.h
ezrayst: I have successfully pass the test functions so it's OK now. Problem with `const`

include/materials/norsand.tcc
```diff
@@ -114,54 +116,26 @@ bool mpm::NorSand<Tdim>::compute_elastic_tensor() {
 
 //! Compute stress invariants
 template <unsigned Tdim>
-Eigen::Matrix<double, 6, 1> mpm::NorSand<Tdim>::compute_stress_invariants(
+double* mpm::NorSand<Tdim>::compute_stress_invariants(
```
kks32: Please don't return a pointer, because now we have no idea what the size is. Please pass pointers when you would like to modify something.

kks32: For an in-out parameter, pass by pointer

include/materials/norsand.tcc
```diff
@@ -114,54 +116,26 @@ bool mpm::NorSand<Tdim>::compute_elastic_tensor() {
...
-
-  // Compute J2
-  double j2 = (std::pow((stress(0) - stress(1)), 2) +
-               std::pow((stress(1) - stress(2)), 2) +
-               std::pow((stress(0) - stress(2)), 2)) /
-                  6.0 +
-              std::pow(stress(3), 2) + std::pow(stress(4), 2) +
-              std::pow(stress(5), 2);
-  j2 = check_low(j2);
+  invariants[0] = check_low(-1. * mpm::materials::p(-stress));
```
kks32: Invariants can be returned in any order, so please use actual pointers to individual invariants

include/materials/norsand.tcc
```diff
@@ -173,13 +147,13 @@ void mpm::NorSand<Tdim>::compute_state_variables(
     mpm::norsand::FailureState yield_type) {
 
   // Get invariants
-  auto invariants = this->compute_stress_invariants(stress, state_vars);
+  auto* invariants = this->compute_stress_invariants(stress, state_vars);
```
kks32: This auto* means we have no idea how big an array we are returning. Please pass as in-out arguments

kks32 (changes requested):

include/materials/norsand.tcc
```diff
@@ -585,4 +482,4 @@ Eigen::Matrix<double, 6, 1> mpm::NorSand<Tdim>::compute_stress(
```
tianchiTJ, line 489: Could we also summarized the computation of plastic strain inside the material utility?

include/materials/norsand.tcc
ezrayst: I could not imagine this, could you give an example? Like shared pointer?

include/materials/norsand.tcc
bodhinandach: Perhaps what @kks32 suggested is this:
```suggestion
void mpm::NorSand<Tdim>::compute_stress_invariants(double* p, double* q, double* lode_angle, double* m_theta)
```
This is pass by pointer: https://www.geeksforgeeks.org/passing-by-pointer-vs-passing-by-reference-in-c/

bodhinandach: You can also pass by reference, both are ok I think :)

include/materials/norsand.tcc
bodhinandach: I think it is a good idea to make it standard for all model.

include/materials/norsand.tcc
kks32: Our general guidelines and ISO-C++ is that if you want to modify pass it as a pointer, if you want to not modify pass it as a `const &`. This will apply both for normal variables and smart pointers

include/materials/norsand.tcc
ezrayst: I think we can make a function to compute `pdstrain` (a scalar plastic deviatoric strain) in `material_utility` and also the `dpstrain` (a vector of plastic strain). But do you think the MohrCoulomb model can be consistent with the current definition?

tianchiTJ: I used your way to compute the pdstrain before, and can get the good result. Please just use your method to compute the dstress first and then dpstrain and then pdstrain. I will use the general function in the MC and MCC model.

ezrayst: Can we do this in another PR?

tianchiTJ: I'm ok with that, but why another? I think it won't have too many changes.

tianchiTJ: OK, sure.

bodhinandach (approved):

include/materials/norsand.h
```diff
@@ -69,13 +69,11 @@ class NorSand : public Material<Tdim> {
   void compute_plastic_tensor(const Vector6d& stress,
                               mpm::dense_map* state_vars);
 
-  //! Compute stress invariants (p, q, j2, j3, lode_angle and M_theta)
+  //! Compute stress invariants (p, q, lode_angle and M_theta)
```
kks32, line 72: Could you please update the doxygen input parameters and return value

include/materials/norsand.tcc
```diff
@@ -202,23 +176,19 @@ void mpm::NorSand<Tdim>::compute_state_variables(
...
+  // Update M_theta at the updated stress state
+  (*state_vars).at("M_theta") = mtheta;
+
   // Update void ratio
   // Note that dstrain is in tension positive - depsv = de / (1 + e_initial)
   double dvolumetric_strain = dstrain(0) + dstrain(1) + dstrain(2);
-  double void_ratio = (*state_vars).at("void_ratio") -
-                      (1 + void_ratio_initial_) * dvolumetric_strain;
-  if (void_ratio < 1.0E-15) void_ratio = 1.0E-15;
+  double void_ratio = check_low((*state_vars).at("void_ratio") -
```
kks32: We don't need to store this variable `void_ratio`, directly write to state_vars

kks32 (changes requested):

include/materials/norsand.h
ezrayst: Ok, please check whether it is as expected

include/materials/norsand.h
```diff
@@ -69,13 +69,12 @@ class NorSand : public Material<Tdim> {
...
   //! \param[in] stress Stress
-  //! \param[in] state_vars History-dependent state variables
-  //! \retval vector of size six, containing p, q, j2, j3, lode_angle and
-  //! M_theta
-  Eigen::Matrix<double, 6, 1> compute_stress_invariants(
-      const Vector6d& stress, mpm::dense_map* state_vars);
+  //! \param[in] 4 variables to update: p, q, lode_angle and M_theta
```
kks32:
```suggestion
  //! \param[in|out] p Mean stress
  //! \param[in|out] q Deviatoric stress
  //! \param[in|out] lode_angle Lode angle
  //! \param[in|out] M_theta Theta(?)
```

include/materials/norsand.h
```diff
@@ -69,13 +69,12 @@ class NorSand : public Material<Tdim> {
...
+  //! \retval 4 variables, namely p, q, lode_angle and M_theta
```
kks32:
```suggestion
```

kks32 (changes requested):

cbgeo (approved):

kks32 (approved):

PR #606, opened 2020-04-24
Add plastic computations to `material_utility`

ezrayst:
- [x] add dplastic_strain computation
- [x] add pdstrain computation

### DIFF
--- a/include/materials/material_utility.h
+++ b/include/materials/material_utility.h
@@ -74,9 +74,20 @@ inline const Eigen::Matrix<double, 6, 1> dtheta_dsigma(
     const Eigen::Matrix<double, 6, 1>& stress,
     double tolerance = std::numeric_limits<double>::epsilon());
 
+//! Compute incremental dplastic_strain
+//! \param[in] dstress Incremental stress in Voigt notation where positive is
+//! tension \param[in] dstrain Incremental total strain in Voigt notation where
+//! positive is tension \param[in] de Elastic stiffness matrix \param[in]
+//! tolerance Default tolerance value specified by user \retval dplastic_strain
+inline const Eigen::Matrix<double, 6, 1> dplastic_strain(
+    const Eigen::Matrix<double, 6, 1>& dstress,
+    const Eigen::Matrix<double, 6, 1>& dstrain,
+    const Eigen::Matrix<double, 6, 6>& de);
+
 //! Compute plastic deviatoric strain
 //! \param[in] plastic_strain Plastic strain in Voigt notation where positive is
-//! tension \retval plastic deviatoric strain
+//! tension
+//! \retval plastic deviatoric strain
 inline double pdstrain(const Eigen::Matrix<double, 6, 1>& plastic_strain);
 
 }  // namespace materials
--- a/include/materials/material_utility.h
+++ b/include/materials/material_utility.h
@@ -10,58 +10,74 @@ namespace materials {
 
 //! Compute mean stress p (tension positive)
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval mean stress p (tension positive)
 inline double p(const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute deviatoric stress
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval deviatoric stress
 inline const Eigen::Matrix<double, 6, 1> deviatoric_stress(
     const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute J2 invariant
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval J2 invariant
 inline double j2(const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute J3 invariant
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval J3 invariant
 inline double j3(const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute deviatoric q
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval deviatoric q
 inline double q(const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute Lode angle theta (cosine convention)
 //! \param[in] stress Stress in Voigt notation where positive is tension
 //! \param[in] tolerance Default tolerance value specified by user
+//! \retval Lode angle theta (cosine convention)
 inline double lode_angle(
     const Eigen::Matrix<double, 6, 1>& stress,
     double tolerance = std::numeric_limits<double>::epsilon());
 
 //! Compute derivative of p in terms of stress sigma
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval derivative of p in terms of stress sigma
 inline const Eigen::Matrix<double, 6, 1> dp_dsigma(
     const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute derivative of q in terms of stress sigma
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval derivative of q in terms of stress sigma
 inline const Eigen::Matrix<double, 6, 1> dq_dsigma(
     const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute derivative of J2 in terms of stress sigma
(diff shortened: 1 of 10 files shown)


include/materials/material_utility.tcc
```diff
@@ -203,3 +203,29 @@ inline const Eigen::Matrix<double, 6, 1> mpm::materials::dtheta_dsigma(
   // Compute and return dtheta / dsigma
   return (dtheta_dr * ((dr_dj2 * dj2_dsigma) + (dr_dj3 * dj3_dsigma)));
 }
+
+//! Compute incremental dplastic_strain
+inline const Eigen::Matrix<double, 6, 1> mpm::materials::dplastic_strain(
```
bodhinandach: I think this should not be here as different methods have different ways of updating plastic strain. Return mapping algorithm append plastic strain incrementally for example.

include/materials/material_utility.tcc
```diff
@@ -203,3 +203,29 @@ inline const Eigen::Matrix<double, 6, 1> mpm::materials::dtheta_dsigma(
...
+    const Eigen::Matrix<double, 6, 1>& dstress,
+    const Eigen::Matrix<double, 6, 1>& dstrain,
+    const Eigen::Matrix<double, 6, 6>& de) {
+  // Compute and return dplastic_strain
+  return (dstrain - de.inverse() * dstress);
+}
+
+//! Compute plastic deviatoric strain
+inline double mpm::materials::pdstrain(
```
bodhinandach, line 208: This is okay though. Each method can call this function to compute pdstrain providing they have the plastic strain.

bodhinandach (changes requested):

include/materials/material_utility.tcc
tianchiTJ: But this is correct no matter what model you use. You can choose to use it or not

tianchiTJ (approved):

include/materials/material_utility.tcc
bodhinandach: That's true. I am okay with it, but I think we should just put this on the law. Passing all the tensors and matrices to a function and only to perform an inversion is not that efficient I guess...

include/materials/material_utility.h
```diff
@@ -10,59 +10,86 @@ namespace materials {
 
 //! Compute mean stress p (tension positive)
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval mean stress p (tension positive)
```
kks32: `\retval` should be followed by the variable name such as `p` then followed by the explanation

include/materials/material_utility.h
```diff
@@ -10,59 +10,86 @@ namespace materials {
...
 //! \param[in] stress Stress in Voigt notation where positive is tension
 //! \param[in] tolerance Default tolerance value specified by user
+//! \retval derivative of Lode angle theta in terms of stress sigma
 inline const Eigen::Matrix<double, 6, 1> dtheta_dsigma(
     const Eigen::Matrix<double, 6, 1>& stress,
     double tolerance = std::numeric_limits<double>::epsilon());
 
+//! Compute incremental dplastic_strain
+//! \param[in] dstress Incremental stress in Voigt notation where positive is
+//! tension \param[in] dstrain Incremental total strain in Voigt notation where
```
kks32: could you please make sure param starts in a new line

include/materials/material_utility.tcc
kks32: I agree with @bodhinandach this might be easier to do locally in the model

kks32 (changes requested):

include/materials/material_utility.tcc
ezrayst: But this should always be correct right, since even for different methods, dstrain is always divided into dstrain_elastic and dstrain_plastic anyways. The formulation of D stiffness matrix is defined in such a way that we subtract D_elastic. So this is correct. 

It's a hassle though, for one line of code, we have to transfer a lot of variables

include/materials/material_utility.h
ezrayst: OK

include/materials/material_utility.h
```diff
@@ -10,59 +10,89 @@ namespace materials {
...
 
 //! Compute deviatoric q
 //! \param[in] stress Stress in Voigt notation where positive is tension
+//! \retval q Deviatoric q
 inline double q(const Eigen::Matrix<double, 6, 1>& stress);
 
 //! Compute Lode angle theta (cosine convention)
 //! \param[in] stress Stress in Voigt notation where positive is tension
 //! \param[in] tolerance Default tolerance value specified by user
+//! \retval Lode angle theta (cosine convention)
```
kks32: should be a variable name

include/materials/material_utility.tcc
kks32: @ezrayst were you replying to the previous comment?

kks32 (changes requested):

include/materials/material_utility.tcc
ezrayst: Yes, sorry, wrong one

include/materials/material_utility.tcc
ezrayst: But this should always be correct right, since even for different methods, dstrain is always divided into dstrain_elastic and dstrain_plastic anyways. The formulation of D stiffness matrix is defined in such a way that we subtract D_elastic. So this is correct.

It's a hassle though, for one line of code, we have to transfer a lot of variables

bodhinandach: Yes, exactly, I am not saying that this is wrong. I just feel that this shouldn't be here due to inefficiency and it is easier that each model can perform their own way of computing `dplastic_strain`.

ezrayst: How about we have this option, and each material can choose whether to use this or not. The same with `j2` and `j3` computations which would not be used too for most models. 

I do feel it is inefficient but so does passing `stress` to only compute `p` right?

Isn't it the purpose of `mpm::materials` to have the common operations? Unless any of you feel that you will never use this at all.

kks32: Here, you are passing 48 elements to get 6 doubles back, so not sure if we really need this function. Compilers may not inline these functions and that would make it slower. We can have it

kks32 (changes requested):

kks32 (approved):

include/materials/material_utility.tcc
ezrayst: OK! Waiting for your approval @bodhinandach

bodhinandach: I am still not sure that this is the best implementation practice. My concern is related to points made by @kks32 and some other points. 

As @kks32 mentioned, this passes 48 elements, 8 times more to compute p, so the scale is a bit different. Imagine if you call this for 1b billion particles. Also think about whether all of those elements can fit into the lowest cache size.

As for `p`, `j2`, and `j3` they are universal and can be used for all types of stresses. However, this specific function only works and correct for models with small-strain elastoplastic assumptions, most likely with explicit implementation. The kinematic decomposition there won't be quite useful for models with large deformation (they use multiplicative decomp), materials with viscous effect (viscous terms are missing), or materials with constrained strains. I think most of the return mapping algorithms also do it incrementally as I mentioned before. And therefore, bring back to my earlier point, why don't we just implement this inside each model?

In conclusion, I can approve this, but this won't give any benefits as compared to computing invariants. There are many limitations in both efficiency and usefulness. I'll let the group decide.

bodhinandach (approved):

include/materials/material_utility.tcc
ezrayst: A few thoughts:
- I agree this won't work for viscosity, but I think it should work for any elastoplastic materials which are the materials that we usually use anyways. I think it should work for large strains too, and it will give the same results with using `lambda` for instance.
- I agree it's not that efficient with 48 doubles to get one single double. However, coming back to comparison with the computation of p, in NorSand at least, one step of compute stress would require at least 4 calls to compute p, which might be more efficient if we do them within the material model itself.
- My point being, not all functions in `mpm::materials` would be used by all material models. We can choose whether to use this function or not and that is fine because we are doing inline anyways. Didn't we want to make as uniform definitions so it is easier to maintain?

I will wait till tomorrow's discussion before merging this PR.

Issue #602, opened 2020-04-23
HDF5 - statevars


ezrayst:
**Describe the bug**
`statevars` are not inputted in order within the HDF5 output file.

**To Reproduce**
Run this [simulation](https://www.dropbox.com/sh/8oj4kdhk02996ju/AAAwYfDV-kLt7oIqiO935-Zva?dl=0) with NorSand model.

**Expected behavior**
Here is my report on the different order than expected:
- `svars_0` is supposed to be `M_theta` and it is correct
- `svars_1` is supposed to be `void_ratio` but it is stored in `svars_7`
- `svars_2` is supposed to be `e_image` but it is stored in `svars_8`
- `svars_3` is supposed to be `p_image` but it is stored in `svars_4`
- `svars_4` is supposed to be `p_cohesion` but it is stored in `svars_9`
- `svars_5` is supposed to be `p_dilation` but it is stored in `svars_10`
and so on

kks32:
Are there any results in `svars_1`, `svars_2` and `svars_3`? or are they empty?

ezrayst:
They are zeros

PR #607, opened 2020-04-25
:hammer: Fix HDF5 material statevars

kks32:
State variables are stored as a dense hashmap (unsorted), which led to #602.

@cw646 @ezrayst Could you please check if this implementation works?

### DIFF
--- a/include/materials/norsand.tcc
+++ b/include/materials/norsand.tcc
@@ -94,6 +94,18 @@ mpm::dense_map mpm::NorSand<Tdim>::initialise_state_variables() {
   return state_vars;
 }
 
+//! Initialise state variables
+template <unsigned Tdim>
+std::vector<std::string> mpm::NorSand<Tdim>::state_variables() const {
+  const std::vector<std::string> state_vars = {
+      "M_theta",         "void_ratio",      "e_image",
+      "p_image",         "p_cohesion",      "p_dilation",
+      "pdstrain",        "plastic_strain0", "plastic_strain1",
+      "plastic_strain2", "plastic_strain3", "plastic_strain4",
+      "plastic_strain5"};
+  return state_vars;
+}
+
 //! Compute elastic tensor
 template <unsigned Tdim>
 bool mpm::NorSand<Tdim>::compute_elastic_tensor() {
--- a/include/particles/particle.tcc
+++ b/include/particles/particle.tcc
@@ -115,8 +115,9 @@ bool mpm::Particle<Tdim>::initialise_particle(
       auto mat_state_vars = material_->initialise_state_variables();
       if (mat_state_vars.size() == particle.nstate_vars) {
         unsigned i = 0;
-        for (const auto& mat_state_var : mat_state_vars) {
-          this->state_variables_[mat_state_var.first] = particle.svars[i];
+        auto state_variables = material_->state_variables();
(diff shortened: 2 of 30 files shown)


tests/materials/modified_cam_clay_test.cc
```diff
@@ -118,6 +118,26 @@ TEST_CASE("Modified cam clay undrained condition is checked in 3D",
       REQUIRE(state_variables.at("pcc") == Approx(0.0).epsilon(Tolerance));
       REQUIRE(state_variables.at("subloading_r") ==
               Approx(1.0).epsilon(Tolerance));
+
+      const std::vector<std::string> state_vars = {"bulk_modulus",
```
ezrayst, line 121: Clang format issue here.
Also, not sure whether we should keep `p`, `q`, `theta` and so on as mentioned. Bulk and shear modulus might be stored as separate function too.

include/particles/particle.tcc
```diff
@@ -115,8 +115,9 @@ bool mpm::Particle<Tdim>::initialise_particle(
       auto mat_state_vars = material_->initialise_state_variables();
       if (mat_state_vars.size() == particle.nstate_vars) {
         unsigned i = 0;
-        for (const auto& mat_state_var : mat_state_vars) {
-          this->state_variables_[mat_state_var.first] = particle.svars[i];
+        auto state_variables = material_->state_variables();
+        for (const auto& state_var : state_variables) {
+          this->state_variables_.at(state_var) = particle.svars[i];
```
ezrayst, line 120: This works! I have tried with my case and they are giving the right order of state variables

include/materials/norsand.tcc
```diff
@@ -94,6 +94,18 @@ mpm::dense_map mpm::NorSand<Tdim>::initialise_state_variables() {
   return state_vars;
 }
 
+//! Initialise state variables
+template <unsigned Tdim>
+std::vector<std::string> mpm::NorSand<Tdim>::state_variables() const {
+  const std::vector<std::string> state_vars = {
```
ezrayst, line 100: I can't think of a better way actually. We might need to have both the vectors of state variables to get the key for the map of state variables where they are actually stored. Any chance the map is ordered so we don't need the vector (we are already using dense map)?

tests/materials/modified_cam_clay_test.cc
kks32: Clang-format doesn't fail, so this is not an issue. 
That's for @tianchiTJ  to change?

include/materials/norsand.tcc
kks32: Our dense map is open addressing, so we can have a contiguous array, we can't use ordered map as it will be stored as a tree rather than a contiguous array.

tests/materials/modified_cam_clay_test.cc
ezrayst: OK Understood. We will work with @tianchiTJ on that!

include/materials/norsand.tcc
ezrayst: Understood. OK, let's do it this way!

ezrayst (approved):

cw646 (approved):
Tested with Matlab for parsing .h5 files, and ParaView for state_vars outputs. Both look good, thanks, Krishna!

PR #610, opened 2020-04-27
:computer: Use material utility in MC and MCC model

tianchiTJ:
1) Use material utility in MC and MCC model
2) Delete unnecessary computation of J3
3) Correct some numeric_limit type
4) Add total pdstrain and pvstrain in MCC model

### DIFF
--- a/include/materials/modified_cam_clay.tcc
+++ b/include/materials/modified_cam_clay.tcc
@@ -274,8 +274,7 @@ bool mpm::ModifiedCamClay<Tdim>::compute_stress_invariants(
   (*state_vars).at("q") = mpm::materials::q(stress);
   // Compute theta (Lode angle)
   if (three_invariants_)
-    (*state_vars).at("theta") = mpm::materials::lode_angle(
-        stress, std::numeric_limits<double>::epsilon());
+    (*state_vars).at("theta") = mpm::materials::lode_angle(stress);
 
   return true;
 }
--- a/include/materials/modified_cam_clay.tcc
+++ b/include/materials/modified_cam_clay.tcc
@@ -485,63 +485,20 @@ void mpm::ModifiedCamClay<Tdim>::compute_df_dsigma(
   // Compute dF / dpc
   double df_dpc = -(p + pcc);
   // Compute dp / dSigma
-  Vector6d dp_dsigma = Vector6d::Zero();
-  dp_dsigma(0) = -1. / 3.;
-  dp_dsigma(1) = -1. / 3.;
-  dp_dsigma(2) = -1. / 3.;
+  Vector6d dp_dsigma = -mpm::materials::dp_dsigma(stress);
   // Compute dq / dSigma
-  Vector6d dq_dsigma = Vector6d::Zero();
-  dq_dsigma = 1.5 / q * dev_stress;
+  Vector6d dq_dsigma = mpm::materials::dq_dsigma(stress);
   // Compute dF/dSigma
   (*df_dsigma) = (df_dp * dp_dsigma) + (df_dq * dq_dsigma);
   // Compute dTheta / dSigma
   if (three_invariants_) {
     // Compute dF / dM
-    double df_dm = 0.;
+    double df_dm = -2 * q * q / std::pow(m_theta, 3);
     // Compute dM / dtheta
-    double dm_dtheta = 0.;
-    // Compute r
-    double r_val = 0.;
-    // Compute j2
-    const double j2 = mpm::materials::j2(stress);
-    const double j3 = mpm::materials::j3(stress);
-    if (std::fabs(j2) > std::numeric_limits<double>::epsilon())
-      r_val = (3. * std::sqrt(3.) / 2.) * (j3 / std::pow(j2, 1.5));
-    // Compute dTheta / dr
-    double divider = 1 - (r_val * r_val);
-    if (divider <= 0.) divider = 1.E-3;
-    double dtheta_dr = -1 / (3. * std::sqrt(divider));
-    // Compute dr / dJ2
-    double dr_dj2 = (-9 * std::sqrt(3.) / 4.) * j3;
-    if (std::fabs(j2) > std::numeric_limits<double>::epsilon())
-      dr_dj2 = dr_dj2 / std::pow(j2, 2.5);
-    // Compute dr / dJ3
-    double dr_dj3 = 1.5 * std::sqrt(3.);
-    if (std::fabs(j2) > std::numeric_limits<double>::epsilon())
-      dr_dj3 = dr_dj3 / std::pow(j2, 1.5);
-    // Compute dJ2 / dSigma
-    Vector6d dj2_dsigma = dev_stress;
-    // Compute dJ3 / dSigma
-    Eigen::Matrix<double, 3, 1> dev1;
-    dev1(0) = dev_stress(0);
-    dev1(1) = dev_stress(3);
-    dev1(2) = dev_stress(5);
-    Eigen::Matrix<double, 3, 1> dev2;
-    dev2(0) = dev_stress(3);
-    dev2(1) = dev_stress(1);
-    dev2(2) = dev_stress(4);
-    Eigen::Matrix<double, 3, 1> dev3;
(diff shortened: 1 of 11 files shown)


include/materials/modified_cam_clay.tcc
```diff
@@ -81,8 +81,6 @@ mpm::dense_map mpm::ModifiedCamClay<Tdim>::initialise_state_variables() {
       {"shear_modulus", 3 * youngs_modulus_ * (1 - 2 * poisson_ratio_) /
                             (2 * (1 + poisson_ratio_))},
       // Stress invariants
-      // J3
-      {"j3", 0.},
       // Volumetric stress
       {"p", 0.},
```
ezrayst, line 85: Do we need to store `p`, `q` and `theta` as state variables?

include/materials/modified_cam_clay.tcc
```diff
@@ -267,51 +269,13 @@ template <unsigned Tdim>
...
-      (*state_vars).at("theta") = M_PI / 3.;
-    if ((*state_vars).at("theta") < 0.0) (*state_vars).at("theta") = 0.;
-  }
+  (*state_vars).at("p") = -mpm::materials::p(stress);
+  // Compute deviatoric q
+  (*state_vars).at("q") = mpm::materials::q(stress);
+  // Compute theta (Lode angle)
+  if (three_invariants_)
+    (*state_vars).at("theta") = mpm::materials::lode_angle(
+        stress, std::numeric_limits<double>::epsilon());
```
ezrayst: Suggestion to add optional `tolerance` value users can specify.
Other than that, you can use `lode_angle` with default tolerance value.

ezrayst:
```suggestion
        stress);
```

include/materials/modified_cam_clay.tcc
```diff
@@ -540,7 +503,8 @@ void mpm::ModifiedCamClay<Tdim>::compute_df_dsigma(
     // Compute r
     double r_val = 0.;
     // Compute j2
-    double j2 = 3 * std::pow(q, 2);
+    const double j2 = mpm::materials::j2(stress);
+    const double j3 = mpm::materials::j3(stress);
```
ezrayst: I don't think you need this anymore. You can get `dtheta_dsigma` from `material_utility`

include/materials/mohr_coulomb.tcc
```diff
@@ -63,8 +63,6 @@ mpm::dense_map mpm::MohrCoulomb<Tdim>::initialise_state_variables() {
                                // Cohesion
                                {"cohesion", this->cohesion_peak_},
                                // Stress invariants
-                               // j3
-                               {"j3", 0.},
```
ezrayst, line 67: Not sure you need `theta` too in this case

include/materials/modified_cam_clay.tcc
ezrayst: Same comments about `dsigma_dtheta` on this model

ezrayst (changes requested):

include/materials/modified_cam_clay.tcc
tianchiTJ: I think we need them as the output results

include/materials/modified_cam_clay.tcc
tianchiTJ: Well I think there is no rule to choose it, so I make it implicit. But if you want it to be an option @kks32 @ezrayst , I can do it.

include/materials/modified_cam_clay.tcc
tianchiTJ: Use it now

include/materials/modified_cam_clay.tcc
ezrayst: What do you mean by output results? Like visualizations or like the `vtk`?

include/materials/modified_cam_clay.tcc
ezrayst: No need actually if you feel this model does not need. If so, you don't have to put tolerance argument to use this function. Please see my suggestion change of this line.

include/materials/mohr_coulomb.tcc
```diff
@@ -101,42 +99,12 @@ template <unsigned Tdim>
...
-  // Compute theta
-  (*state_vars).at("theta") = (1. / 3.) * acos(theta_val);
-  // Check theta
-  if ((*state_vars).at("theta") > M_PI / 3.)
-    (*state_vars).at("theta") = M_PI / 3.;
-  if ((*state_vars).at("theta") < 0.0) (*state_vars).at("theta") = 0.;
+  (*state_vars).at("epsilon") = mpm::materials::p(stress) * std::sqrt(3.);
+  // Compute theta value
+  (*state_vars).at("theta") = mpm::materials::lode_angle(
+      stress, std::numeric_limits<double>::epsilon());
```
ezrayst:
```suggestion
      stress);
```

include/materials/mohr_coulomb.tcc
```diff
@@ -101,42 +99,12 @@ template <unsigned Tdim>
...
   // Compute rho
-  (*state_vars).at("rho") = std::sqrt(2. * (j2));
-  // Compute epsilon
-  (*state_vars).at("epsilon") =
-      (1. / std::sqrt(3.)) * (stress(0) + stress(1) + stress(2));
+  (*state_vars).at("rho") = std::sqrt(2. * mpm::materials::j2(stress));
```
ezrayst, line 106: I am not sure you need to save `rho` as a state variable too.

include/materials/modified_cam_clay.tcc
tianchiTJ: Yes, and I also need it in may functions inside the model.

include/materials/modified_cam_clay.tcc
tianchiTJ: I see, agree with it.

include/materials/mohr_coulomb.tcc
```diff
@@ -399,7 +311,7 @@ void mpm::MohrCoulomb<Tdim>::compute_df_dp(
             ((sin(phi) * sin(theta + M_PI / 3.) /
               (std::sqrt(3.) * cos(phi) * cos(phi))) +
              (cos(theta + M_PI / 3.) / (3. * cos(phi) * cos(phi)))) +
-        (mean_p / (cos(phi) * cos(phi)));
+        (mpm::materials::p(stress) / (cos(phi) * cos(phi)));
```
ezrayst, line 313: The `p` function in `mpm::materials` give you tension positive. Just making sure, your `p` definition here also tension positive right?

ezrayst (changes requested):

include/materials/mohr_coulomb.tcc
tianchiTJ: Yes, I have passed all test, and they should be correct.

ezrayst: I see, good then!

include/materials/modified_cam_clay.tcc
ezrayst: Should we save them as state variables though? Maybe need @kks32 to decide on this.
I personally do not store any invariants in the norsand model. But this can work too.

kks32: VTK attribute can be computed directly in ParaView rather than storing these values, if we can reduce this, we can reduce the number of VTK state variables set to 20. This will help reduce communication.

kks32 (changes requested):

include/materials/modified_cam_clay.tcc
tianchiTJ: Now it's 19 state variables. I can improve it if needed. Can we finished this first and then I will reduce its number.

ezrayst: I am OK with it

ezrayst (approved):

ezrayst:
Let's get this PR which is to use `mpm::materials`. Then please optimize the `state_vars` after that

kks32 (approved):

PR #613, opened 2020-04-28
NorSand use pdstrain and make some parameters optional

ezrayst:
**Describe the PR**
NorSand model has been modified:
- To use `pdstrain` within `mpm::materials`
- As discussed, `bonded_model` is now an optional input variable along with the other five variables including `p_cohesion_initial`, `p_dilation_initial`, `m_cohesion`, `m_dilation` and `m_modulus`.
- The test cases have also been changed to reflect this change, only for the non bonded part of the test

### DIFF
--- a/include/materials/norsand.tcc
+++ b/include/materials/norsand.tcc
@@ -35,8 +35,11 @@ mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
     p_image_initial_ =
         material_properties.at("p_image_initial").template get<double>();
     // Flag for bonded model
-    if (material_properties.find("bond_model") != material_properties.end()) {
+    if (material_properties.find("bond_model") != material_properties.end())
       bond_model_ = material_properties.at("bond_model").template get<bool>();
+
+    // Obtain the rest of the bonded parameters
+    if (bond_model_) {
       // Initial p_cohesion
       p_cohesion_initial_ =
           material_properties.at("p_cohesion_initial").template get<double>();
--- a/include/materials/norsand.tcc
+++ b/include/materials/norsand.tcc
@@ -52,18 +52,8 @@ mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
       m_dilation_ = material_properties.at("m_dilation").template get<double>();
       // Parameter for shear modulus
       m_modulus_ = material_properties.at("m_modulus").template get<double>();
-    } else {
-      // Initial p_cohesion
-      p_cohesion_initial_ = 0.;
-      // Initial p_dilation
-      p_dilation_initial_ = 0.;
-      // Cohesion degradation parameter m upon shearing
-      m_cohesion_ = 0.;
-      // Dilation degradation parameter m upon shearing
-      m_dilation_ = 0.;
-      // Parameter for shear modulus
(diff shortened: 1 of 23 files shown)


include/materials/norsand.tcc
```diff
@@ -35,19 +35,32 @@ mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
...
-        material_properties.at("p_dilation_initial").template get<double>();
-    // Cohesion degradation parameter m upon shearing
-    m_cohesion_ = material_properties.at("m_cohesion").template get<double>();
-    // Dilation degradation parameter m upon shearing
-    m_dilation_ = material_properties.at("m_dilation").template get<double>();
-    // Parameter for shear modulus
-    m_modulus_ = material_properties.at("m_modulus").template get<double>();
+    if (material_properties.find("bond_model") != material_properties.end()) {
+      bond_model_ = material_properties.at("bond_model").template get<bool>();
+      // Initial p_cohesion
```
kks32, line 43: The following should only be checked if `bond_model_` is `true`

tests/materials/norsand_test.cc
```diff
@@ -39,13 +39,6 @@ TEST_CASE("NorSand is checked in 3D non-bonded model",
   jmaterial["hardening_modulus"] = 200.0;
   jmaterial["void_ratio_initial"] = 0.85;
   jmaterial["p_image_initial"] = 87014.6;
-  jmaterial["bond_model"] = false;
```
kks32, line 42: Why are they removed from testing?

kks32 (changes requested):

tests/materials/norsand_test.cc
ezrayst: To test that they can be initialized without specifying those `bonded` parameters

include/materials/norsand.tcc
ezrayst: Modified!

include/materials/norsand.tcc
```diff
@@ -35,19 +35,35 @@ mpm::NorSand<Tdim>::NorSand(unsigned id, const Json& material_properties)
...
+      p_dilation_initial_ =
+          material_properties.at("p_dilation_initial").template get<double>();
+      // Cohesion degradation parameter m upon shearing
+      m_cohesion_ = material_properties.at("m_cohesion").template get<double>();
+      // Dilation degradation parameter m upon shearing
+      m_dilation_ = material_properties.at("m_dilation").template get<double>();
+      // Parameter for shear modulus
+      m_modulus_ = material_properties.at("m_modulus").template get<double>();
+    } else {
+      // Initial p_cohesion
```
kks32: Could we set these as the default values in the header and we don't need an else clause

tests/materials/norsand_test.cc
kks32: Are there tests with these values?

kks32 (changes requested):

tests/materials/norsand_test.cc
ezrayst: Test for `false`? The default initialization is `false` anyway right?

bodhinandach (approved):

kks32 (approved):

Issue #609, opened 2020-04-27
Generate Random Particles for Accurate Integration 


ezrayst:
# Generate Random Particles and Accurate Integration MPM Scheme

## Design Detail
- Have a new particle generator (currently either from 'file` or at `gauss points`)
- This generates random particles within a cell, on which the weights will be computed separately to ensure accurate integration
- The weights will ensure accurate integration

## Implementation Detail
- Start with generate random particles first
- Then once that's done, we will compute the gauss weights

ezrayst:
After discussion with @kks32, we will change this from `cell` (generate points function). 

The new input JSON file might look something like this with label `random`:
```
  "particles": [
    {
      "generator": {
        "check_duplicates": true,
        "cset_id": -1,
        "particle_type": "P2D",
        "material_id": 2,
        "pset_id": 2,
        "particles_per_cell": 20,
        "type": "random"
      }
  ]
```

kks32:
Could we have separate issues for updating weights and random particle generations?

ezrayst:
Understood, will start with random particle generations first

ezrayst:
Continued in #615

PR #614, opened 2020-04-29
Generate random particles

ezrayst:
**Describe the PR**
Generate random particles within an element as first step to solve Issue #609.

General design of the random generator:
- Make a new class called `QuadrilateralRandomQuadratures` 
- Update `assign_quadrilateral` in `cell` to include `generator_type` with default string value of "gauss"

Alternative design:
- Generate random quadratures in `geometry` but the design of the code would be rather messy

Things to do in this PR (this will notsolve issue #609 entirely)
- [x] Implement for 1, 4, 9, 16 and 25 particles per cell for quadrilateral element
- [ ] Implement for triangular and hexahedron (for completion) 
- [ ] Test functions (note that random generator cannot be tested for locations of particles, but only the number of particles within the cell)

Things to do in next PR (to complete issue #609)
- [ ] Compute accurate gauss weights for each randomly generated particles
- [ ] Test functions on the computed weights

Example of generation of 25 particles per cell: 
image

Example of generation of 4 particles per cell:
image

### DIFF
--- a/include/cell.h
+++ b/include/cell.h
@@ -62,7 +62,7 @@ class Cell {
   bool is_initialised() const;
 
   //! Assign quadrature
-  void assign_quadrature(unsigned nquadratures);
+  void assign_quadrature(unsigned nquadratures, const std::string& generator_type= "gauss");
 
   //! Generate points
   std::vector<Eigen::Matrix<double, Tdim, 1>> generate_points();
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -78,11 +78,11 @@ bool mpm::Cell<Tdim>::is_initialised() const {
 
 //! Assign quadrature
 template <unsigned Tdim>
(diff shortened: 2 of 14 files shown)


kks32 (changes requested):
Quadratures are not a reasonable way to generate random points within a cell:

1. When we integrate FE solvers, these random quadratures would be incorrect / invalid in those situations.
2. The weights associated for each random point would mean we will have each cell with its own weights and set of points, which defeats the purpose of having a generic quadrature class. 

I suggest not to continue to on this path to generate material points randomly, please open an RFC issue and we can flush out the details of the implementation before writing any further code.

kks32:
@ezrayst A discussion or a Request for comments is for discussion before a PR and would like for it to happen in an issue before we write code. PR shouldn't be the place where we have a general discussion of how to implement a new feature, especially when you have spent time coding it. I'm closing this issue, as this is not a possible way to solve the random point generation. We need more discussion so please open an RFC

PR #617, opened 2020-05-01
:hammer: Refactor/constraints in Mesh

kks32:
**Describe the PR**
Refactor velocity constraints on nodes to `Constraints` file. The mesh test file has been split to 2D and 3D, so only check if applying velocity constraints have moved from mesh to constraints object in the test. 

**Additional context**
Mesh file is supposed to have only nodes, elements, and particles. Constraints should move to a separate BCs / Loading.


ezrayst (approved):
Thank you Krishna. 

I have a question actually: there should not be performance difference compared to having `constraints` in `mesh` as before right? Furthermore, we have cleaner code. Lastly, I think it's been a while since we update the `overview` of the code within `mpm-doc`. Should we do that too?

kks32:
@ezrayst there are no performance differences.

kks32:
@bodhinandach Waiting for review.

bodhinandach (approved):

bodhinandach:
Thanks @kks32! We need to do a bit of cleanup in mesh class. Will help you during summer.

PR #618, opened 2020-05-04
Create nodal properties pool in mesh

thiagordonho:
Implement `create_nodal_properties` function in the Mesh Class. It creates the properties that are necessary for the multimaterial contact algorithm being implemented. The property is created for all nodes considering all materials in the mesh.

The create_nodal_properties() function is called in the mpm_base.tcc file as part of the initialise_mesh() function.


include/mesh.tcc
```diff
@@ -1883,3 +1883,36 @@ bool mpm::Mesh<Tdim>::assign_nodal_friction_constraints(
...
 }
+
+//! Assign nodal velocity constraints
+template <unsigned Tdim>
+bool mpm::Mesh<Tdim>::initialise_nodal_properties(int set_id) {
+  bool status = true;
+  // Initialise the shared pointer to nodal properties
+  nodal_properties_ = std::make_shared<mpm::NodalProperties>();
+  try {
+    if (set_id == -1 || node_sets_.find(set_id) != node_sets_.end()) {
```
kks32: If we have a dynamic interface, then the nodal property size are going to change all the time. So, I'd suggest we set the nodal properties to all the nodes initially, we can improve this later. I wouldn't want to pass a set-id because it doesn't get updated at each time step.

include/mesh.tcc
```diff
@@ -1883,3 +1883,36 @@ bool mpm::Mesh<Tdim>::assign_nodal_friction_constraints(
...
+      // Create pool data for each property in the nodal properties struct
+      // object. Properties must be named in the plural form
+      nodal_properties_->create_property("masses", nset.size(),
+                                         materials_.size());
+      nodal_properties_->create_property("momenta", nset.size() * Tdim,
+                                         materials_.size());
+
+      // Iterate over all nodes in the selected node set to initialise the
+      // property handle in each node
+      for (unsigned i = 0; i < nset.size(); ++i)
```
kks32: Could we do a range-based for loop instead?

kks32 (changes requested):

include/mesh.tcc
thiagordonho: I agree. I only added the option of choosing a set because I though it could be useful for initial testing. I will just leave the initialise function with all the nodes, though, as you instructed.

include/mesh.tcc
```diff
@@ -1745,3 +1745,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
   }
   return status;
 }
+
+// Initialise the nodal properties' pool
+template <unsigned Tdim>
+bool mpm::Mesh<Tdim>::initialise_nodal_properties() {
+  // Initialise the shared pointer to nodal properties
+  nodal_properties_ = std::make_shared<mpm::NodalProperties>();
```
kks32, line 1893: Every time we initialize a nodal property, we recreate this object? Could we split to `create_nodal_properties` and `initialize`, which makes it all zero, so rename to `create_nodal_properties()`

include/mesh.tcc
```diff
@@ -1745,3 +1745,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
...
+
+  // Create pool data for each property in the nodal properties struct
+  // object. Properties must be named in the plural form
+  nodal_properties_->create_property("masses", nodes_.size(),
```
kks32: You could check if `nodes_.size()` and `materials_.size()` are non-zero

kks32 (changes requested):

include/mesh.tcc
```diff
@@ -1743,3 +1743,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
   }
   return status;
 }
+
+// Initialise the nodal properties' pool
+template <unsigned Tdim>
+bool mpm::Mesh<Tdim>::initialise_nodal_properties() {
```
cbgeo: No need to return

cbgeo (changes requested):

include/mesh.tcc
```diff
@@ -1743,3 +1743,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
...
+  nodal_properties_->create_property("masses", nodes_.size(),
+                                     materials_.size());
+  nodal_properties_->create_property("momenta", nodes_.size() * Tdim,
+                                     materials_.size());
+
+  // Iterate over all nodes to initialise the property handle in each node
+  // and assign its node id as the prop id in the nodal property data pool
+  for (auto nitr = nodes_.cbegin(); nitr != nodes_.cend(); ++nitr)
+    (*nitr)->initialise_property_handle((*nitr)->id(), nodal_properties_);
+
```
cbgeo: else: throw std::runtime_error

include/mesh.tcc
```diff
@@ -1746,99 +1746,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
   return status;
 }
 
-//! Assign velocity constraints to nodes
```
kks32: Why are these removed?

include/mesh.tcc
```diff
@@ -1746,99 +1746,24 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
...
-}
-
 //! Assign nodal velocity constraints
 template <unsigned Tdim>
-bool mpm::Mesh<Tdim>::initialise_nodal_properties(int set_id) {
+bool mpm::Mesh<Tdim>::initialise_nodal_properties() {
   bool status = true;
   // Initialise the shared pointer to nodal properties
   nodal_properties_ = std::make_shared<mpm::NodalProperties>();
   try {
```
kks32: An if, else condition, please do not use a try block

include/mesh.tcc
thiagordonho: Because they are not part of the mesh.tcc file anymore. They are in the load_bcs folder as part of the constraints.tcc file. This was removed as a merge and resolve merge conflicts commits when I merged from develop.

kks32 (approved):

include/solvers/mpm_base.tcc
```diff
@@ -187,6 +187,9 @@ bool mpm::MPMBase<Tdim>::initialise_mesh() {
     // Read and assign friction constraints
     this->nodal_frictional_constraints(mesh_props, mesh_io);
 
+    // Create nodal properties
+    mesh_->create_nodal_properties();
```
kks32: We would have to only do this only if someone has `interface` enabled.

kks32: Otherwise, we are creating a huge amount of memory unnecessarily. Suggestion is, we create another `MPMExplicit` for now with an option to do interface

kks32 (changes requested):

include/solvers/mpm_base.tcc
thiagordonho: Hum... Do you mean to create an entire new MPMExplicit file where the nodal properties are going to be dealt with? Not just create a boolean option to be read from the JSON file, then.

kks32: Yes, that's correct. We'll discuss a better way later, for now, interface should not affect the main branch.

kks32 (approved):

cbgeo (approved):

PR #622, opened 2020-05-07
:wrench: Use default MPI_TAG to avoid mpi_tag overflow fixes #620

### LINKED ISSUE #620: MPI_TAG overflows for large number of domain shared nodes with Halo exchange
**Describe the bug**
Running a large problem where the number of shared nodes exceeds the capacity of MPI_TAG a 16-bit signed integer, we have an overflow problem in the MPI_TAG.

```
Abort(671723268) on node 3 (rank 3 in comm 0): Fatal error in PMPI_Isend: Invalid tag, error stack:
PMPI_Isend(148): MPI_Isend(buf=0x7fff7328ea78, count=1, MPI_DOUBLE, dest=5, tag=524547, MPI_COMM_WORLD, request=0xc54a70fc) failed
PMPI_Isend(95).: Invalid tag, value is 524547
```

**To Reproduce**
Run a problem where the number of domain shared nodes exceed 32,768.

**Expected behavior**
Pass `MPI_TAG` without an overflow

 - Frontera

**Additional context**
How can we alter this prototype code to avoid `MPI_TAG` overflows? At the moment, the node id or particle id is used as the `MPI_TAG`.

kks32:
**Describe the PR**
Using node id as the `mpi_tag` causes an overflow. Instead, we use the default tag of 0.

**Related Issues/PRs**
This fixes #620 by using the default `mpi_tag` of 0. This OK because MPI guarantees the order of send, and we won't need the tag to identify the information. `Order Messages are non-overtaking`https://www.mpi-forum.org/docs/mpi-1.1/mpi-11-html/node41.html.

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -145,9 +145,8 @@ void mpm::Mesh<Tdim>::nodal_halo_exchange(Tgetfunctor getter,
       std::set<unsigned> node_mpi_ranks = domain_shared_nodes_[i]->mpi_ranks();
       for (auto& node_rank : node_mpi_ranks) {
         if (node_rank != mpi_rank) {
-          MPI_Isend(&property, Tnparam, MPI_DOUBLE, node_rank,
-                    domain_shared_nodes_[i]->id(), MPI_COMM_WORLD,
-                    &send_requests[j]);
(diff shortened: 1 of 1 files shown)


cbgeo (approved):

bodhinandach (approved):
Wow. This simple method works :) Thanks @kks32!

PR #624, opened 2020-05-08
Resume bug

kks32:
**Describe the PR**
Resume from HDF5 fails on MPI processes because after reading all the particles initially, we don't remove them. This causes a failed pointer issue. This PR reads the HDF5 particle, gets info on the number of rows (particles), and uses this information to accurately resume the process.

**Additional context**
Dynamic load balancing fails when there is significant particle movement. Will open an issue to describe the issue. This PR temporarily disables dynamic load balancing.

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1440,18 +1440,27 @@ std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::nodal_coordinates()
 
 //! Cell node pairs
 template <unsigned Tdim>
-std::vector<std::array<mpm::Index, 2>> mpm::Mesh<Tdim>::node_pairs() const {
+std::vector<std::array<mpm::Index, 2>> mpm::Mesh<Tdim>::node_pairs(
+    bool active) const {
   // Vector of node_pairs
   std::vector<std::array<mpm::Index, 2>> node_pairs;
 
   try {
+    int mpi_rank = 0;
+#ifdef USE_MPI
+    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
+#endif
     if (cells_.size() == 0)
       throw std::runtime_error("No cells have been initialised!");
 
     for (auto citr = cells_.cbegin(); citr != cells_.cend(); ++citr) {
-      const auto pairs = (*citr)->side_node_pairs();
-      node_pairs.insert(std::end(node_pairs), std::begin(pairs),
-                        std::end(pairs));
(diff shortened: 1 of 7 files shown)


include/mesh.tcc
```diff
@@ -1346,29 +1346,50 @@ bool mpm::Mesh<Tdim>::read_particles_hdf5(unsigned phase,
...
   }
   // close the file
   H5Fclose(file_id);
+
+  // Overwrite particles container
+  this->particles_ = particles;
+
+  // Remove associated cell for the particle
+  for (auto citr = this->cells_.cbegin(); citr != this->cells_.cend(); ++citr)
+    (*citr)->clear_particle_ids();
```
bodhinandach, line 1391: Don't we need to allocate the resumed particles to cell here? I mean now cell don't have a pointer to particles?

include/solvers/mpm_explicit.tcc
```diff
@@ -281,9 +281,8 @@ bool mpm::MPMExplicit<Tdim>::solve() {
 #ifdef USE_MPI
 #ifdef USE_GRAPH_PARTITIONING
     // Run load balancer at a specified frequency
-    if (step_ % nload_balance_steps_ == 0 && step_ != 0)
-      this->mpi_domain_decompose(false);
-
+    // if (step_ % nload_balance_steps_ == 0 && step_ != 0)
+    //   this->mpi_domain_decompose(false);
```
bodhinandach, line 285: Do you do this intentionally?

bodhinandach (approved):

include/mesh.tcc
kks32: Cells only have particle ids, and at each step, the list of ids gets refreshed when you do locate particles.

include/solvers/mpm_explicit.tcc
kks32: It's causing an issue at large deformations, so temporarily disabling to avoid this issue and keeps `develop` functioning. I have described it in PR.

include/mesh.tcc
bodhinandach: I see.. We will call that function anyway right.. Thanks for clarifying.

include/solvers/mpm_explicit.tcc
bodhinandach: Ok! No problem. @kks32 For the load balancer, I have an idea to improve the performance. Will open an issue.

cbgeo (approved):

Issue #636, opened 2020-05-14
Contact Algorithm


thiagordonho:
# Contact Algorithm

## Summary

Implementation of a contact algorithm to handle the frictional interaction between adjacent materials at the region of contact between the two, also known as the multi-material contact or interface. This implementation is based on the procedures provided by [Bardenhagen _et al._ (2000)](https://doi.org/10.1016/S0045-7825(99)00338-2) and [Nairn (2013)](http://www.cof.orst.edu/cof/wse/faculty/Nairn/papers/MMInterfaces.pdf).

## Motivation

Currently, the CB-Geo MPM code computes a single velocity field for one-phase problems. This means that, upon computing the nodal velocities towards the end of each time step, the code will update the material point’s (or particle’s) position using the same velocity field, regardless of the distinct bodies being represented. The distinct bodies will thus behave as if they are "stuck" to each other. Nevertheless, this is not always the case for every model, since one may desire to attribute to the body the capacity of moving freely if separating, or of developing friction at the contact surface with another body if they are approaching.

The proposed feature, thus, allows one to develop simulations where multi-material interface interaction is essential in the analysis. Cases such as soil-structure interaction, slip surfaces in slope stability analyses, fracture movement/rupture, among others, represent common geotechnical problems that fit this contact requirement. By applying this feature, the user is able to assign the desired friction characteristics between every pair of bodies.

## Design Detail

The MPM contact algorithm as described by Bardenhagen et al. (2000) can be summarized into the following stages: reset the nodal properties, map the particle’s properties to the nodes, compute the nodal kinematics of separate bodies by satisfying the conservation of linear momentum, correct the kinematics by applying contact law and no-interpenetration conditions at the interface between bodies, and update the particles’ kinematics. This brief overview of the contact algorithm stages also indicates that the nodal properties must be stored for separate bodies, instead of the singular nodal velocity field of the one-phase conventional MPM. Therefore, a **data structure** solution is included in this design detail to explain the chosen approach for implementing the algorithm to the current CB-Geo MPM code. The stages of the representative contact time-step loop are fulfilled by iterating over the particles or nodes to perform each of the tasks. Hence, new methods, hereafter referred to as **contact-related methods**, are added to the `Particle`, `Node`, and `Mesh` classes to perform these tasks. An additional **Contact class** is proposed to store the parameters of the contact law and to orderly call the contact-related methods. Finally, a description of the **input** needed for using this feature is given.

### Data structure

[...]

PR #643, opened 2020-05-28
Initialise nodal, multimaterial properties

thiagordonho:
Implement the initialisation of all the properties in the nodal properties pool and call this function from the `MPMExplicit`. The following describe the rational behind the `initialise_properties` function:

- Iterates over all properties in the `std::map` of properties.
- Creates a null matrix of the same size of the property's `Eigen::Matrix`
- Assign the null matrix to the entire property's `Eigen::Matrix`using the previously implemented `assign_property` in `NodalProperties`.

After proper implementation, call `initialise_properties()` at the beginning of the computational step loop.

This PR is related to the RFC #636 .
A reset to `zero` of all the properties' values is expected with the changes in this PR. This initialisation is necessary in order to recalculate the multimaterial properties with every step in the Contact Algorithm.

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1912,7 +1912,7 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
 
 // Initialise the nodal properties' map
 template <unsigned Tdim>
-void mpm::Mesh<Tdim>::initialise_properties() {
+void mpm::Mesh<Tdim>::initialise_nodal_properties() {
   // Call initialise_properties function from the nodal properties
-  nodal_properties_->initialise_properties();
+  nodal_properties_->initialise_nodal_properties();
 }
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1886,7 +1886,7 @@ bool mpm::Mesh<Tdim>::assign_nodal_concentrated_forces(
   return status;
 }
 
-// Create the nodal properties' pool
+// Create the nodal properties' map
 template <unsigned Tdim>
 void mpm::Mesh<Tdim>::create_nodal_properties() {
   // Initialise the shared pointer to nodal properties
@@ -1908,4 +1908,11 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
   } else {
     throw std::runtime_error("Number of nodes or number of materials is zero");
   }
+}
+
+// Initialise the nodal properties' map
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::initialise_properties() {
+  // Call initialise_properties function from the nodal properties
+  nodal_properties_->initialise_properties();
 }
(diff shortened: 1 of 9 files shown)


include/mesh.tcc
```diff
@@ -1909,3 +1909,10 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
     throw std::runtime_error("Number of nodes or number of materials is zero");
   }
 }
+
+// Initialise the nodal properties' map
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::initialise_properties() {
```
kks32: Rename to initialise_nodal_properties

include/solvers/mpm_explicit.tcc
```diff
@@ -170,11 +170,15 @@ bool mpm::MPMExplicit<Tdim>::solve() {
...
     task_group.wait();
 
-    // Assign material ids to node
-    if (interface_)
+    // Initialise nodal properties and assign material ids to node
+    if (interface_) {
+      // Initialise nodal properties
+      mesh_->initialise_properties();
+
       mesh_->iterate_over_particles(
```
kks32, line 178: Does this reset the material ids already present in the node?

src/nodal_properties.cc
```diff
@@ -42,3 +43,17 @@ void mpm::NodalProperties::update_property(
   properties_.at(property).block(node_id * nprops, mat_id, nprops, 1) =
       property_value + this->property(property, node_id, mat_id, nprops);
 }
+
+// Initialise all the nodal values for all properties in the property pool
+void mpm::NodalProperties::initialise_properties() {
+  // Iterate over all properties in the property map
+  for (std::map<std::string, Eigen::MatrixXd>::iterator prop_itr =
```
kks32: use auto

src/nodal_properties.cc
```diff
@@ -42,3 +43,17 @@ void mpm::NodalProperties::update_property(
...
+           properties_.begin();
+       prop_itr != properties_.end(); ++prop_itr) {
+    // Create Matrix with zero values that has same size of the current property
+    // in the iteration
+    Eigen::MatrixXd zeroed_property =
```
kks32, line 57: Is the size the entire size of particles * # of materials?

kks32 (changes requested):

include/solvers/mpm_explicit.tcc
thiagordonho: It does not. It only resets the nodal properties' values. But now that you have said it, it should also reset the material ids in the node as well, right?

src/nodal_properties.cc
thiagordonho: The size is equal to `rows * cols` where:
- `rows` = the number of nodes * property dimension (1 if scalar, 2 or 3 if 2D or 3D vector, respectively)
- `cols` = number of materials

thiagordonho:
@kks32 , how can I fix this code coverage issue?

include/solvers/mpm_explicit.tcc
thiagordonho: material_ids_.clear() is called in the Node class (`initialise()` function), thus already being reset as the nodes are reset.

kks32 (approved):

kks32 (approved):

PR #648, opened 2020-06-03
Feature/contact/compute momenta change

thiagordonho:
Implement a function in the `Node` class that computes the change in momentum of all the properties in the nodal properties pool, and call this function from the MPMExplicit by iterating over all nodes. The following describe the rationale behind the `compute_multimaterial_change_in_momentum()` function:

- Iterate over all the material ids stored in the node -- stored in `std::set material_ids_`.
- Compute the change in momentum for each material and for this node as: `delta_momentum = velocity_ * mass - momentum`, where `mass` and `momentum` are values obtained from the existing property pool for the given material id and node id, and `velocity_` is the velocity of the node in question. 

After proper implementation, call this function while iterating over nodes at every computational step in the `MPMExplicit` class. 

```
    if (interface_)
      mesh_->iterate_over_nodes(std::bind(
          &mpm::NodeBase<Tdim>::compute_multimaterial_change_in_momentum,
          std::placeholders::_1));
```

This function is only called after the velocities at the nodes are computed but before velocity constraints are applied. Additionally, the computation must only be done if `interface_ == true`. 

The calculation of the change in momentum for each node and different material is expected with this PR. This property is essential for the contact algorithm since it represents one of the verifications of the contact between two adjacent materials while also being the main property used to change the momentum accordingly if contact is identified.

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1889,14 +1889,15 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
 
   // Check if nodes_ and materials_is empty and throw runtime error if they are
   if (nodes_.size() != 0 && materials_.size() != 0) {
+    // Compute number of rows in nodal properties for vector entities
+    const unsigned nrows = nodes_.size() * Tdim;
     // Create pool data for each property in the nodal properties struct
     // object. Properties must be named in the plural form
     nodal_properties_->create_property("masses", nodes_.size(),
                                        materials_.size());
-    nodal_properties_->create_property("momenta", nodes_.size() * Tdim,
+    nodal_properties_->create_property("momenta", nrows, materials_.size());
+    nodal_properties_->create_property("change_in_momenta", nrows,
                                        materials_.size());
-    nodal_properties_->create_property("change_in_momenta",
-                                       nodes_.size() * Tdim, materials_.size());
 
     // Iterate over all nodes to initialise the property handle in each node
     // and assign its node id as the prop id in the nodal property data pool
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1895,6 +1895,8 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
                                        materials_.size());
     nodal_properties_->create_property("momenta", nodes_.size() * Tdim,
                                        materials_.size());
+    nodal_properties_->create_property("change_in_momenta", nodes_.size() * Tdim,
+                                       materials_.size());
 
     // Iterate over all nodes to initialise the property handle in each node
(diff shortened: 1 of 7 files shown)


include/mesh.tcc
```diff
@@ -1895,6 +1895,8 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
                                        materials_.size());
     nodal_properties_->create_property("momenta", nodes_.size() * Tdim,
                                        materials_.size());
+    nodal_properties_->create_property("change_in_momenta",
```
kks32: Could we store `nodes_.size() * Tdim` as a `const unsigned` variable in this function?

include/node.tcc
```diff
@@ -539,3 +539,19 @@ void mpm::Node<Tdim, Tdof, Tnphases>::update_property(
...
+void mpm::Node<Tdim, Tdof,
+               Tnphases>::compute_multimaterial_change_in_momentum() {
+  // iterate over all materials in the material_ids set and update the change in
+  // momentum
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  for (auto mat_itr = material_ids_.begin(); mat_itr != material_ids_.end();
+       ++mat_itr)
+    property_handle_->update_property(
+        "change_in_momenta", prop_id_, *mat_itr,
+        velocity_ * property_handle_->property("masses", prop_id_, *mat_itr) -
```
kks32: Could you take these calculations out separately add a const store it in a variable and use the variables here in the function call, otherwise it's confusing. 

Please use `mitr` instead of `mat_itr` as it is easier to read.

kks32 (changes requested):

include/node.tcc
```diff
@@ -547,11 +547,13 @@ void mpm::Node<Tdim, Tdof,
...
   std::lock_guard<std::mutex> guard(node_mutex_);
-  for (auto mat_itr = material_ids_.begin(); mat_itr != material_ids_.end();
-       ++mat_itr)
-    property_handle_->update_property(
-        "change_in_momenta", prop_id_, *mat_itr,
-        velocity_ * property_handle_->property("masses", prop_id_, *mat_itr) -
-            property_handle_->property("momenta", prop_id_, *mat_itr, Tdim),
-        Tdim);
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    Eigen::MatrixXd mass =
```
kks32: could you make these const if possible use the size: `Eigen::Matrix<double, Tdim, 1>` or  `Eigen::Matrix<double, 1, 1>`

kks32 (changes requested):

kks32 (approved):

PR #649, opened 2020-06-09
Feature/contact/compute displacements

thiagordonho:
Implement a function in the `Particle` class that maps the displacements from the material points to the nodal properties pool. Additionally, implement another function in the `Node` class to compute the nodal multimaterial displacements and separation vectors. Both of these functions are called from the `MPMExplicit` solve function within every computation step by iterating over all particles and all nodes, respectively. The following describes the rationale behind the `map_multimaterial_displacements_to_nodes()` function:

- Iterate over all the nodes in the cell this particle belongs to -- stored in `nodes_`.
- Compute the nodal displacement as `mass_ * shapefn_[i] * displacement_` and update the nodal property at the node.

The following describes the rationale behind the `compute_multimaterial_separation_vector()` function:

- Iterate over all materials in the node -- stored in `std::set material ids`.
- Update the nodal displacements by dividing by the nodal-material mass.
- Calculate the displacement of the centre of mass at this node -- `displacement_`.
- Compute the separation vector as given in [Nairn (2013)](http://www.cof.orst.edu/cof/wse/faculty/Nairn/papers/MMInterfaces.pdf) and assign it to `nodal_properties_` (this computation is done in a separate iteration over all the materials, for the displacement of the centre of mass is necessary).

The calculation of the displacements and separation vectors for each node and material is expected with this PR. The separation vector is essential for the contact algorithm since it is used in a second verification of the contact between two adjacent materials.

### DIFF
--- a/include/node.tcc
+++ b/include/node.tcc
@@ -569,9 +569,9 @@ void mpm::Node<Tdim, Tdof,
   // node
   std::lock_guard<std::mutex> guard(node_mutex_);
   for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
-    const Eigen::Matrix<double, Tdim, 1> material_displacement =
+    const auto& material_displacement =
         property_handle_->property("displacements", prop_id_, *mitr, Tdim);
-    const Eigen::Matrix<double, 1, 1> material_mass =
+    const auto& material_mass =
         property_handle_->property("masses", prop_id_, *mitr);
 
     // displacement of the center of mass
@@ -592,9 +592,9 @@ void mpm::Node<Tdim, Tdof,
         property_handle_->property("masses", prop_id_, *mitr);
 
     // Update the separation vector property
-    Eigen::Matrix<double, Tdim, 1> separation_vector =
-        (displacement_ - material_displacement) * mass_(0, 0) /
-        (mass_(0, 0) - material_mass(0, 0));
+    const auto& separation_vector = (displacement_ - material_displacement) *
+                                    mass_(0, 0) /
+                                    (mass_(0, 0) - material_mass(0, 0));
     property_handle_->update_property("separation_vectors", prop_id_, *mitr,
                                       separation_vector, Tdim);
   }
--- a/include/node.tcc
+++ b/include/node.tcc
@@ -28,7 +28,7 @@ void mpm::Node<Tdim, Tdof, Tnphases>::initialise() noexcept {
   external_force_.setZero();
   internal_force_.setZero();
   pressure_.setZero();
-  displacement_.setZero();
+  contact_displacement_.setZero();
   velocity_.setZero();
   momentum_.setZero();
   acceleration_.setZero();
@@ -575,7 +575,7 @@ void mpm::Node<Tdim, Tdof,
         property_handle_->property("masses", prop_id_, *mitr);
 
     // displacement of the center of mass
-    displacement_ += material_displacement / mass_(0, 0);
+    contact_displacement_ += material_displacement / mass_(0, 0);
     // assign nodal-multimaterial displacement by dividing it by this material's
     // mass
     property_handle_->assign_property(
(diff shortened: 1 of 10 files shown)


include/node.tcc
```diff
@@ -558,3 +559,43 @@ void mpm::Node<Tdim, Tdof,
...
+    displacement_ += material_displacement / mass_(0, 0);
+    // assign nodal-multimaterial displacement by dividing it by this material's
+    // mass
+    property_handle_->assign_property(
+        "displacements", prop_id_, *mitr,
+        material_displacement / material_mass(0, 0), Tdim);
+  }
+
+  // iterate over all materials in the material_ids to compute the separation
+  // vector
```
kks32, line 587: Does this have to be computed after the displacement in previous for, could we merge these two for loops?

kks32 (changes requested):

include/node.tcc
thiagordonho: They can't be merged because the displacement of the center of mass needs to be calculated in the first loop before it can be used in the second loop to calculate the separation vector.

include/node.tcc
```diff
@@ -558,3 +559,43 @@ void mpm::Node<Tdim, Tdof,
...
+               Tnphases>::compute_multimaterial_separation_vector() {
+  // iterate over all materials in the material_ids set, update the
+  // displacements and calculate the displacement of the center of mass for this
+  // node
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    const Eigen::Matrix<double, Tdim, 1> material_displacement =
+        property_handle_->property("displacements", prop_id_, *mitr, Tdim);
+    const Eigen::Matrix<double, 1, 1> material_mass =
+        property_handle_->property("masses", prop_id_, *mitr);
```
bodhinandach:
```suggestion
    const auto& material_displacement =
        property_handle_->property("displacements", prop_id_, *mitr, Tdim);
    const auto& material_mass =
        property_handle_->property("masses", prop_id_, *mitr);
```

include/node.tcc
```diff
@@ -558,3 +559,43 @@ void mpm::Node<Tdim, Tdof,
...
+  // iterate over all materials in the material_ids to compute the separation
+  // vector
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    const Eigen::Matrix<double, Tdim, 1> material_displacement =
+        property_handle_->property("displacements", prop_id_, *mitr, Tdim);
+    const Eigen::Matrix<double, 1, 1> material_mass =
+        property_handle_->property("masses", prop_id_, *mitr);
+
+    // Update the separation vector property
+    Eigen::Matrix<double, Tdim, 1> separation_vector =
```
bodhinandach: `const`?

include/particles/particle.tcc
```diff
@@ -515,7 +516,23 @@ void mpm::Particle<Tdim>::map_multimaterial_mass_momentum_to_nodes() noexcept {
...
+//! Map multimaterial displacements to nodes
+template <unsigned Tdim>
+void mpm::Particle<Tdim>::map_multimaterial_displacements_to_nodes() noexcept {
+  // Check if particle mass is set
+  assert(mass_ != std::numeric_limits<double>::max());
+
+  // Map displacements to nodal property and divide it by the respective
+  // nodal-material mass
+  for (unsigned i = 0; i < nodes_.size(); ++i) {
+    Eigen::Matrix<double, Tdim, 1> displacement =
```
bodhinandach: `const`?

bodhinandach (commented):
So far looks okay for me @thiagordonho. I added my suggestion and have one minor comment: will that be better if we somehow change the naming of `displacement_` to something like `contact_displacement_` to avoid confusion of nodal variable. I was thinking `displacement_` is the integration of `velocity_`.

thiagordonho:
> So far looks okay for me @thiagordonho. I added my suggestion and have one minor comment: will that be better if we somehow change the naming of `displacement_` to something like `contact_displacement_` to avoid confusion of nodal variable. I was thinking `displacement_` is the integration of `velocity_`.

I think that is a very good idea! I will change all the related variables. Thank you for that and for the other remarks.

bodhinandach (approved):

kks32 (approved):

Issue #637, opened 2020-05-16
Refactor particle to handle different types


kks32:
## Summary

Refactor `Particle` data to have the generic functions with the ability to fetch and modify generic scalar and vector properties, map, and interpolate these properties to nodes. Add a separate implementation functions to operate on these properties of the particle. These functions could be grouped under the namespace of solid particle functions, fluid-particle function, and mixed particle functions. The particle agnostic design of functions would allow for mix-and-matching different functions in the Solver class, without having to deal with multiple-inheritance or repetition.

## Motivation

Refactoring and separating the Particle data from the implementation of algorithm enables handling different particle data types, without having to have a lot of virtual functions in the ParticleBase and other subsequent derived particles. Addresses the comments raised in RFCs #633 and #634. For example, deriving `FluidParticle` from `Particle` would result in extra data needed for `FluidParticle`, which is not needed in the case of `Particle` conversely, `FluidParticle` won't need certain `Particle` data. However, the important consideration is that the functions of a `FluidParticle`s have to be virtualized in the abstract `ParticleBase` and every time a new class is derived, more abstract functions have to be added to the base and other existing derived classes. These functions, which are not useful, must, therefore, throw an error. The Abstract Factory design with inconsistent derived functions is inherently prone to errors when a developer fails to add functions to all the derived classes. This RFC proposes an alternate way of deriving particle data and keeping the implementation separate and reusable. Separation of implementation allows for the composition of different functions in the Solver class.

## Design Detail
Particle class has a generic ordered map of scalar and vector properties. An open addressing scheme will be used to store the map properties, that the data stored in here will remain continuous in memory. These properties can be accessed using `scalar_property` and `vector_property`. `Particle` types will still control the initialization of these different properties, i.e., derived classes will be used to create new types of Particles, but all these classes will have a consistent functional interface. This avoids the problem of too many virtual functions. The `ParticleBase` will be altered to have the following. A consistent interface allows transferring particles across MPI ranks.

```c++
// Scalar properties
tsl::ordered_map<std::string, double> scalar_properties_;
tsl::ordered_map<std::string, Eigen::Matrix<double, Tdim, 1>> vector_properties_;
```
Scalar and vector properties can be set and returned using generic functions. 
```c++
void assign_property(const std::string& property, double value);
double property(const std::string& property) const;
```

Nodes will also have a similar property pool structure, which will allow us to write generic map and interpolation functions. 

```c++
//! Map scalar properties to nodes
template <unsigned Tdim>
[...]

PR #650, opened 2020-06-11
Feature/particle refactor

kks32:
**Describe the PR**
Refactor particle class to handle multiple particle properties. Remove the particle base class. Add implementation for handling particle scalar properties. 

**Related Issues/PRs**
This is a work in progress implementation of #637 

**Additional context**
Still to resolve
- The ideal way to handle particle MPI transfer
- Process of handling HDF5 Particle struct


thiagordonho (approved):

tianchiTJ (approved):
If I want to define the `twophase` particle, what type of structure should I use? `template-twophase` or a new `namespace-twophase` in `namespace-mpm`?

kks32:
@tianchiTJ maybe namespace `mpm::Particle::TwoPhase`

tianchiTJ:
> @tianchiTJ maybe namespace `mpm::Particle::TwoPhase`

Sure, and for some general function like `initialize_particle`. Shall we create a new function like `initialize_twophase_particle` or other way? If use new function `initialize_twophase_particle`, is it only use to define variables for fluid phase and still need to use the original `initialize_particle` to define variables for solid phase, or include all defination?

WeijianLiang (approved):

include/node.h
```diff
@@ -272,6 +287,12 @@ class Node : public NodeBase<Tdim> {
   unsigned dof_{std::numeric_limits<unsigned>::max()};
   //! Status
   bool status_{false};
+  //! Scalar properties
+  tsl::ordered_map<mpm::properties::Scalar, Eigen::Matrix<double, 1, Tnphases>>
+      scalar_properties_;
+  //! Vector properties
+  tsl::ordered_map<mpm::properties::Scalar, Eigen::Matrix<double, 1, Tnphases>>
```
bodhinandach: Why `mpm::properties::Scalar`? Are we gonna put `Velocity` in `mpm::properties::Scalar`? Where to specify and store the dimensional components?

include/particles/particle.h
```diff
@@ -254,56 +305,60 @@ class Particle : public ParticleBase<Tdim> {
...
+  //! Unsigned material id
+  unsigned material_id_{std::numeric_limits<unsigned>::max()};
+  //! Material state history variables
+  mpm::dense_map state_variables_;
+  //! Vector of particle neighbour ids
+  std::vector<mpm::Index> neighbours_;
+  //! Scalar properties
+  tsl::ordered_map<mpm::properties::Scalar, double> scalar_properties_;
+  //! Vector properties
+  tsl::ordered_map<mpm::properties::Scalar, Eigen::Matrix<double, 1, Tdim>>
```
bodhinandach: Why `mpm::properties::Scalar`? What enum and data type are you planning to define vectorial variables like velocity?

include/particles/particle.tcc
```diff
@@ -238,10 +238,10 @@ void mpm::Particle<Tdim>::initialise() {
...
 
   // Initialize vector data properties
-  this->properties_["stresses"] = [&]() { return stress(); };
-  this->properties_["strains"] = [&]() { return strain(); };
-  this->properties_["velocities"] = [&]() { return velocity(); };
-  this->properties_["displacements"] = [&]() { return displacement(); };
+  properties_["stresses"] = [&]() { return stress(); };
+  properties_["strains"] = [&]() { return strain(); };
+  properties_["velocities"] = [&]() { return velocity(); };
+  properties_["displacements"] = [&]() { return displacement(); };
```
bodhinandach, line 244: Are you planning to combine this with the new `vector_properties`?

src/particle.cc
```diff
@@ -1,13 +1,13 @@
...
 
 // Particle2D (2 Dim)
-static Register<mpm::ParticleBase<2>, mpm::Particle<2>, mpm::Index,
+static Register<mpm::Particle<2>, mpm::Particle<2>, mpm::Index,
                 const Eigen::Matrix<double, 2, 1>&>
     particle2d("P2D");
 
 // Particle3D (3 Dim)
-static Register<mpm::ParticleBase<3>, mpm::Particle<3>, mpm::Index,
+static Register<mpm::Particle<3>, mpm::Particle<3>, mpm::Index,
```
bodhinandach, line 11: Can you explain how we can register fluid particle and two-phase particles with this constructor?

bodhinandach (commented):
Thanks @kks32. I have some comments to address.

include/node.h
kks32: That's just a place holder for now

include/node.tcc
```diff
@@ -45,6 +49,27 @@ void mpm::Node<Tdim, Tdof, Tnphases>::initialise_property_handle(
...
+  // Decide to update or assign
+  const double factor = (update == true) ? 1. : 0.;
+
+  // Update/assign mass
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  scalar_properties_.at(property)[phase] =
+      (scalar_properties_.at(property)[phase] * factor) + value;
+}
+
+//! Update scalar property at the nodes from particle
```
ezrayst, line 66:
```
//! Return scalar property at nodes
```

include/node_base.h
```diff
@@ -17,6 +17,11 @@
 
 namespace mpm {
 
+namespace properties {
+//! Sclar Properties
```
ezrayst:
```
//! Scalar properties
```

include/node_base.h
```diff
@@ -17,6 +17,11 @@
...
+enum Scalar : unsigned int { Mass };
```
ezrayst, line 22: Any thoughts of scoping the enum - like what we do in `particle` for solid, liquid and gas. I personally think there is no difference, but giving the scope might be cleaner.

include/particles/particle.h
ezrayst: I think @kks32 has not made the enum for Vector. The most important thing is that the map has `Eigen::Matrix<double, 1, Tdim>` which is a vector. On that note (I will write it in general comment), do we need tensor too? How do we store stresses?

ezrayst (commented):
Generally, thanks for doing this, @kks32.

My only comment other than the minute details I put in line is regarding the stresses and strains. We do `mpm::properties::Scalar` for mass, I assume you will make `mpm::properties::Vector` for displacements, velocities and accelerations, how about stresses? Do you want to make `mpm::properties::Tensor` but have 6x1 vector inside that instead?

include/particles/particle.tcc
kks32: We'll remove it when we have vector properties fully integrated.

src/particle.cc
kks32: There is no difference in the constructor. We plan to store the particle type as a member variable

include/node_base.h
kks32: It's both particle and node properties.

src/particle.cc
bodhinandach: Then, are we planning to read it from .json and assign the particle type when calling initialise_particles()?

include/node_base.h
ezrayst: What I meant is to put like `Mass = 0, Something = 1` instead of just `Mass`

include/particles/particle.h
kks32: Not at the moment, we'll use stresses and strains  direclty for no

src/particle.cc
kks32: Yes, like we do now, we initialize the type appropriately from JSON object.

include/node_base.h
kks32: It will automatically do that for you

ezrayst: I agree, but then why we do it [here](https://github.com/cb-geo/mpm/blob/7d4fe7524cc4ebe58bd4593e86b69b8ed80d5639/include/particles/particle_base.h#L22)?

kks32:
I'll close this issue/branch as #654 supersedes this and is probably a better design. Please raise any objections you may have. I'll delete this branch after our weekly meeting.

PR #654, opened 2020-06-29
[Refactor] Particle variables and functions handling

### LINKED ISSUE #637: Refactor particle to handle different types

kks32:
**Describe the PR**
This is an alternative implementation of particle refactor. 

After some more consideration, I am wondering about the possibility to derive particle types, but only keep the common particle functions within the class, and merge all independent/specialized functions to be outside the class. This will help to initialize the classes with the right types, and also in MPI transfer, as the type is known. We derived data of individual Particle types but keep specialized particle functions separately. Maybe the particle functions may have to stay in the same namespace.

**Related Issues/PRs**
#637 and #650 

**Additional context**
Specialized functions will be still separate.

Following are the run-times for 3D hydrostatic column 10k steps on Stampede2

| Branch                   | TACC runtime (s) |
|--------------------------|-------------------|
| develop                  | 742.6            |
| prop map + free fn | 1280             |
| vector + free fn map       | 865             |
| flatmap + free fn       | 972            |


include/node.tcc
```diff
@@ -46,28 +81,63 @@ void mpm::Node<Tdim, Tdof, Tnphases>::initialise_property_handle(
...
+//! Update vector property at the nodes from particle
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+void mpm::Node<Tdim, Tdof, Tnphases>::update_vector_property(
+    mpm::properties::Vector property, bool update, unsigned phase,
+    const Eigen::Matrix<double, Tdim, 1>& value) noexcept {
   // Decide to update or assign
   const double factor = (update == true) ? 1. : 0.;
 
-  // Update/assign volume
+  // Update/assign mass
```
kks32:
```suggestion
  // Update/assign value
```

include/node.tcc
```diff
@@ -108,47 +178,26 @@ template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
 void mpm::Node<Tdim, Tdof, Tnphases>::update_external_force(
     bool update, unsigned phase,
     const Eigen::Matrix<double, Tdim, 1>& force) noexcept {
-  // Assert
-  assert(phase < Tnphases);
```
kks32, line 114: Should we move this check to `update_vector_property`

include/particles/particle.h
```diff
@@ -307,12 +309,12 @@ class Particle : public ParticleBase<Tdim> {
   using ParticleBase<Tdim>::state_variables_;
   //! Neighbour particles
   using ParticleBase<Tdim>::neighbours_;
-  //! Volumetric mass density (mass / volume)
-  double mass_density_{0.};
-  //! Mass
-  double mass_{0.};
-  //! Volume
-  double volume_{0.};
```
kks32, line 337: I thought we were going to keep volume inside particle, no?

bodhinandach: I thought there might be a chance to have `liquid_volume` too in the future, so thinking to have them inside `properties`.

tests/particle_test.cc
```diff
@@ -145,10 +145,23 @@ TEST_CASE("Particle is checked for 1D case", "[particle][1D]") {
         std::make_shared<mpm::Particle<Dim>>(id, coords);
 
     // Check mass
-    REQUIRE(particle->mass() == Approx(0.0).epsilon(Tolerance));
-    double mass = 100.5;
+    REQUIRE(particle->mass() == Approx(0.).epsilon(Tolerance));
+    double mass = 201.0;
```
ezrayst: Any thoughts why change the value of the test?

bodhinandach: That's just to check the return and assign values to mass, no different. I will revert to minimize the change. Thanks for pointing that out.

tests/particle_test.cc
```diff
@@ -678,8 +711,6 @@ TEST_CASE("Particle is checked for 2D case", "[particle][2D]") {
     // TODO Assert: REQUIRE_NOTHROW(particle->compute_updated_position(dt,
     // true)); Compute volume
     // TODO Assert: REQUIRE(particle->compute_volume() == false);
-    // Update volume should fail
-    // TODO Assert: REQUIRE(particle->update_volume() == false);
```
ezrayst, line 682: Why is this deleted?

tests/particle_test.cc
```diff
@@ -737,13 +765,16 @@ TEST_CASE("Particle is checked for 2D case", "[particle][2D]") {
     REQUIRE_NOTHROW(particle->compute_volume());
 
     // Compute mass
-    REQUIRE_NOTHROW(particle->compute_mass());
+    REQUIRE_NOTHROW(mpm::particle::compute_mass<Dim>(particle));
```
ezrayst, line 775: There is no `try` in `compute_mass` actually. Should we still require this?

tests/particle_test.cc
```diff
@@ -737,13 +765,16 @@ TEST_CASE("Particle is checked for 2D case", "[particle][2D]") {
...
     // Mass
     REQUIRE(particle->mass() == Approx(1000.).epsilon(Tolerance));
+    // Compute mass function
+    REQUIRE_NOTHROW(mpm::particle::compute_mass(particle));
```
ezrayst, line 772: Same as above

tests/particle_test.cc
ezrayst: Sorry, there are `try` within the functions in this function. If any would `throw`, it would be caught here right?

tests/particle_test.cc
```diff
@@ -1223,10 +1253,23 @@ TEST_CASE("Particle is checked for 2D case", "[particle][2D]") {
         std::make_shared<mpm::Particle<Dim>>(id, coords);
 
     // Check mass
-    REQUIRE(particle->mass() == Approx(0.0).epsilon(Tolerance));
-    double mass = 100.5;
+    REQUIRE(particle->mass() == Approx(0.).epsilon(Tolerance));
+    double mass = 201.0;
```
ezrayst: Same as the other one, but it's really up to you, I thought there are more fundamental reasons to change

tests/particle_test.cc
bodhinandach: That's commented test. Tbh, there are many of them found all related to assert. I would prefer to clean them up. What do you think @kks32?

kks32: I think catch has a `REQUIRE_THROW` with a specific exception, we could test against those

bodhinandach:
Dear all, I think this refactoring branch is ready for review. Could you all please some time to take a look at it? We can discuss it for sure in the following developer meeting. Here are some notes to ease the checking process:
- The main changes are located mainly in `particle` and `node` class with the addition of `mpm_properties` and `particle_functions`. 
- Some functions in `particle` are moved out to `particle_functions`.
- Assign (or update) and return variables in nodes and particles are kept as before.
- In particle, tensorial properties are still kept as before, we will refactor these in another PR, if necessary.
- After a discussion with @thiagordonho, the contact functions in particle class are still kept as it is in this PR. We need to refactor how nodal_properties are stored and hopefully coming with similar approach to this. This will also be in a separate PR.
- Please refer to issue #637 for further details. 

As this change is breaking for all the development, could you please test it to run some of your simulations and let us know if you encounter any issue. Many many thanks in advance for the help. Any comments and questions are appreciated.

tests/particle_test.cc
bodhinandach: yes

tests/particle_test.cc
bodhinandach: @kks32 Should I revert the deleted comment and change to `REQUIRE_THROW`? Even though it is not tested here, I think it has been called in some other part of the test with `REQUIRE_NOTHROW`.

.circleci/config.yml
```diff
@@ -127,7 +127,7 @@ workflows:
     version: 2
     build:
         jobs:
-            - gcc
+            - benchmarks
```
bodhinandach: A small reminder to revert this back before merging.

tianchiTJ:
Why `initialise_particle(const HDF5Particle& particle, const std::shared_ptr<Material<Tdim>>& material)` is defined as virtual function in `particle.h`?

bodhinandach:
> Why `initialise_particle(const HDF5Particle& particle, const std::shared_ptr<Material<Tdim>>& material)` is defined as virtual function in `particle.h`?

Good point. I will remove it, we should either use `virtual` or `override`, but not both in one go.

tianchiTJ:
> > Why `initialise_particle(const HDF5Particle& particle, const std::shared_ptr<Material<Tdim>>& material)` is defined as virtual function in `particle.h`?
> 
> Good point. I will remove it, we should either use `virtual` or `override`, but not both in one go.

This one also exists in the `develop` branch

bodhinandach:
@tianchiTJ the most recent commit should fix, thanks

include/particles/particle.h
```diff
@@ -117,26 +120,17 @@ class Particle : public ParticleBase<Tdim> {
   bool assign_volume(double volume) override;
 
   //! Return volume
-  double volume() const override { return volume_; }
+  double volume() const override {
+    return this->scalar_property(mpm::properties::Scalar::Volume);
+  }
```
tianchiTJ, line 127: The current version define all simple return value functions in the derive class. How can I define such function for two phase? Such as `liquid_mass()` `porosity()` `liquid_material()`

bodhinandach: If it is inside your `twophase_particle`, you can create a private function to ease the calling process, for instance:
```
private:
double pore_pressure() const {
    return this->scalar_property(mpm::properties::Scalar::PorePressure);
}
```
However, to call it from the `twophase_particle_functions`, testings, or outside files, you can use:
```
particle->scalar_property(mpm::properties::Scalar::PorePressure);
```

jgiven100:
I'm getting matching outputs running a PEA simulation on `develop` and `refactor/particles` 👍

include/particles/particle.h
tianchiTJ: Yes, I need it in the `twophase_particle_functions`. I can do this way, but it will be not identical to the single phase format.

bodhinandach: I understand. That should be ok.

bodhinandach: If you use the variable multiple times, maybe you want to define it in the beginning of the function, such that:
`const auto pore_pressure = particle->scalar_property(mpm::properties::Scalar::PorePressure);`
and you can use it multiple times.

tianchiTJ: This also means a particle pointer should be set as an argument of the functions.

bodhinandach: Yes, you are right, the same as other functions in `particle_functions`.

include/particles/particle.h
```diff
@@ -329,18 +334,12 @@ class Particle : public ParticleBase<Tdim> {
...
-  //! Velocity
-  Eigen::Matrix<double, Tdim, 1> velocity_;
-  //! Displacement
-  Eigen::Matrix<double, Tdim, 1> displacement_;
   //! Particle velocity constraints
   std::map<unsigned, double> particle_velocity_constraints_;
   //! Set traction
   bool set_traction_{false};
   //! Surface Traction (given as a stress; force/area)
   Eigen::Matrix<double, Tdim, 1> traction_;
```
tianchiTJ, line 364: Why not put the `traction_` into the `vector_properties_`?

include/particles/particle_functions.tcc
```diff
@@ -0,0 +1,121 @@
...
+void compute_mass(std::shared_ptr<mpm::ParticleBase<Tdim>> particle) noexcept {
+  // Check if particle volume is set and material ptr is valid
+  assert(particle->volume() != std::numeric_limits<double>::max() &&
+         particle->material() != nullptr);
+
+  // Mass = volume of particle * mass_density
+  particle->update_scalar_property(
+      mpm::properties::Scalar::MassDensity, false,
+      particle->material()->template property<double>(std::string("density")));
+
```
tianchiTJ, line 16: Could we move this assign density function into the `assign_material`? Because for the twophase, there is no pointer for the `liquid_material` in the `twophase_particle_functions.tcc`

include/particles/particle_functions.tcc
```diff
@@ -0,0 +1,121 @@
...
+//! Map particle mass and momentum to nodes
+template <unsigned Tdim>
+void map_mass_momentum_to_nodes(
+    std::shared_ptr<mpm::ParticleBase<Tdim>> particle) noexcept {
+  // Check if particle mass is set
+  assert(particle->mass() != std::numeric_limits<double>::max());
+
+  // Map mass and momentum to nodes
+  particle->map_scalar_property_nodes(mpm::properties::Scalar::Mass, true,
+                                      mpm::ParticlePhase::Solid);
```
tianchiTJ: I think the `map_scalar_property_nodes` has a problem to be used for two phase. The liquid mass is called `LiquidMass` in the particle properties, but `Mass` at liquid_phase in the node properties. That's conflict.

bodhinandach: Use the overload function with input value as your `LiquidMass`
https://github.com/cb-geo/mpm/blob/164491fd7f5c7ef09536405aeb2ae0d5fd8947ea/include/particles/particle_base.tcc#L50-L53

include/node.h
```diff
@@ -65,6 +65,33 @@ class Node : public NodeBase<Tdim> {
...
 
+  //! Update scalar property at the nodes
+  //! \param[in] property Name of the property to update
+  //! \param[in] update A boolean to update (true) or assign (false)
+  //! \param[in] phase Index corresponding to the phase
+  //! \param[in] value Property value from the particles in a cell
+  void update_scalar_property(mpm::properties::Scalar property, bool update,
+                              unsigned phase, double value) noexcept override;
+
+  //! Return property at a given node for a given phase
```
thiagordonho, line 76: `//! \param[in] property Name of the property to return`

include/node_base.h
```diff
@@ -70,6 +72,34 @@ class NodeBase {
...
+  //! Update vector property at the nodes
+  //! \param[in] property Name of the property to update
+  //! \param[in] update A boolean to update (true) or assign (false)
+  //! \param[in] phase Index corresponding to the phase
+  //! \param[in] value Property value from the particles in a cell
+  virtual void update_vector_property(
+      mpm::properties::Vector property, bool update, unsigned phase,
+      const Eigen::Matrix<double, Tdim, 1>& value) noexcept = 0;
+
+  //! Return property at a given node for a given phase
```
thiagordonho, line 111:
```suggestion
  //! Return property at a given node for a given phase
  //! \param[in] property Name of the property to return
```

include/node_base.h
```diff
@@ -70,6 +72,34 @@ class NodeBase {
...
+  //! Update scalar property at the nodes
+  //! \param[in] property Name of the property to update
+  //! \param[in] update A boolean to update (true) or assign (false)
+  //! \param[in] phase Index corresponding to the phase
+  //! \param[in] value Property value from the particles in a cell
+  virtual void update_scalar_property(mpm::properties::Scalar property,
+                                      bool update, unsigned phase,
+                                      double value) noexcept = 0;
+
+  //! Return property at a given node for a given phase
```
thiagordonho, line 95:
```suggestion
  //! Return property at a given node for a given phase
  //! \param[in] property Name of the property to return
```

include/node.h
```diff
@@ -65,6 +65,33 @@ class Node : public NodeBase<Tdim> {
...
+  //! Update vector property at the nodes
+  //! \param[in] property Name of the property to update
+  //! \param[in] update A boolean to update (true) or assign (false)
+  //! \param[in] phase Index corresponding to the phase
+  //! \param[in] value Property value from the particles in a cell
+  virtual void update_vector_property(
+      mpm::properties::Vector property, bool update, unsigned phase,
+      const Eigen::Matrix<double, Tdim, 1>& value) noexcept override;
+
+  //! Return property at a given node for a given phase
```
thiagordonho, line 104:
```suggestion
  //! Return property at a given node for a given phase
  //! \param[in] property Name of the property to return
```

include/node.h
thiagordonho, line 88:
```suggestion
  //! Return property at a given node for a given phase
  //! \param[in] property Name of the property to return
```

include/node.tcc
```diff
@@ -108,71 +192,61 @@ template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
...
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+void mpm::Node<Tdim, Tdof, Tnphases>::compute_pressure() {
   const double tolerance = 1.E-16;
-  // Compute pressure from mass*pressure
-  if (mass_(phase) > tolerance) {
-    std::lock_guard<std::mutex> guard(node_mutex_);
-    pressure_(phase) += mass_pressure / mass_(phase);
+  for (unsigned phase = 0; phase < Tnphases; ++phase) {
+    if (this->mass(phase) > tolerance) {
+      scalar_properties_.at(mpm::properties::Scalar::Pressure)(phase) =
```
kks32, line 268: Don't we have pressure as a state variable?

bodhinandach: Yes, that's on `particle`, but on `node` we store it as `mpm::properties`.

thiagordonho (dismissed):

include/particles/particle_functions.tcc
bodhinandach: Thanks, I will address it in a future PR.

include/particles/particle.h
bodhinandach: As discussed, BC will also be in a future PR.

tianchiTJ (commented):
How could I call the two phase variable, such as `liquid_material` in the `twophase_particle_functions.tcc`, which is neither a scalar nor a vector?

tianchiTJ (dismissed):
Problem with `liquid_material` will be addressed in a follow up PR @bodhinandach

bodhinandach (dismissed):

ezrayst:
Anyways, how is the performance of using ordered map on different scalar and vector properties? Any comparison done yet?

kks32:
The 3D benchmarks of hydrostatic column shows this branch is about 1.5 - 2x slower. 

develop: 615931 ms
refactor-particles: 980017 ms

TACC develop: 742655 ms
TACC refactor-particles: 1280757 ms

kks32:
> Anyways, how is the performance of using ordered map on different scalar and vector properties? Any comparison done yet?

We have to look at that now separately to see if that gives us any benefit.

bodhinandach:
@kks32 I created a partially revert branch `refactor/particles_2` to move back the functions in `particle_functions.tcc` back to `particles`. Just want to see whether the slowdown is caused by restructuring the properties or moving the function out to a separate file with passing pointer argument. I am re-running the same 3D hydrostatic benchmark locally.

kks32:
Looks like the map function is the reason

| Branch                   | Local runtime (s) | TACC runtime (s) |
|--------------------------|-------------------|------------------|
| develop                  | 615.9             | 742.6            |
| particle (prop map + free fn) | 980               | 1280             |
| particle2 (only prop map)     | 872.9             | 1250             |

kks32:
image
Definitely the map lookup is the reason for the slowness

bodhinandach:
@kks32 Thanks for checking! I was worried about this before. Any thoughts on changing the data structure?

kks32:
I'm trying out a few different data structures, I'll let you know once I have the results.

kks32:
I tried a few different structures:


| Branch                   | TACC runtime (s) |
|--------------------------|-------------------|
| develop                  | 742.6            |
| prop map + free fn | 1280             |
| only prop map       | 1250             |
| only robin map       | 1012            |
| only vector       | 850             |

The vector implementation seems close to develop. Let me try vector implementation with free functions

kks32:
> I tried a few different structures:
> Branch 	TACC runtime (s)
> develop 	742.6
> prop map + free fn 	1280
> only prop map 	1250
> only robin map 	1012
> only vector 	850
> 
> The vector implementation seems close to develop. Let me try vector implementation with free functions

Vector with free functions is 865s.

bodhinandach:
Thanks @kks32, that makes it 1.16 times slower than the original. Are you checking the OpenMP/MPI performance as well? 
As the current refactor is created purposely for fluid and two-phase implementation, do you want me to check the vector one with the navier-stokes branch too?

kks32:
Vector implementation profiling

profile

include/particles/particle.tcc
```diff
@@ -224,20 +230,29 @@ mpm::HDF5Particle mpm::Particle<Tdim>::hdf5() const {
...
-  velocity_.setZero();
-  volume_ = std::numeric_limits<double>::max();
   volumetric_strain_centroid_ = 0.;
 
+  // Initialize scalar properties
+  scalar_properties_.resize(3);
+  scalar_properties_.at(mpm::properties::Scalar::Mass) = 0.;
+  scalar_properties_.at(mpm::properties::Scalar::MassDensity) = 0.;
+  scalar_properties_.at(mpm::properties::Scalar::Volume) =
+      std::numeric_limits<double>::max();
```
bodhinandach: @kks32 I think this could be a problem. Currently, in `mpm_properties`, `Mass`, `Volume`, and `MassDensity` are defined with index 0, 1, and 2, which is safe for this case. But this will be an error if we want to add another variable which enum associated index is larger than the `size` that you specified in `scalar_properties_.resize(size);`. It will be a memory access error.

include/particles/particle.tcc
```diff
@@ -224,20 +230,29 @@ mpm::HDF5Particle mpm::Particle<Tdim>::hdf5() const {
...
+
+  // Initialize vector properties
+  vector_properties_.resize(2);
+  vector_properties_.at(mpm::properties::Vector::Displacement) =
+      VectorDim::Zero();
+  vector_properties_.at(mpm::properties::Vector::Velocity) = VectorDim::Zero();
```
bodhinandach: The same here with `Displacement` and `Velocity`...

bodhinandach (changes requested):

include/node.tcc
```diff
@@ -17,23 +17,80 @@ mpm::Node<Tdim, Tdof, Tnphases>::Node(
...
+      Eigen::Matrix<double, 1, Tnphases>::Zero();
+  scalar_properties_.at(mpm::properties::Scalar::MassPressure) =
+      Eigen::Matrix<double, 1, Tnphases>::Zero();
+  scalar_properties_.at(mpm::properties::Scalar::Pressure) =
+      Eigen::Matrix<double, 1, Tnphases>::Zero();
+
+  // Initialise nodal vector properties
+  vector_properties_.resize(6);
+  vector_properties_.at(mpm::properties::Vector::Displacement) =
+      Eigen::Matrix<double, Tdim, Tnphases>::Zero();
```
bodhinandach: `Displacement` is not needed in node, but I know why we need to append it here to avoid access error. Does that means for every variable added in `mpm_properties`, we need to arrange and allocate all other particles? I don't think that's efficient storage-wise.

kks32:
Profiling of develop

image

kks32:
The problem with vector is like you said, we may have to fill everything otherwise we have an access issue.

kks32:
Profiling for robin hasmap

image

kks32:
| Branch                   | TACC runtime (s) |
|--------------------------|-------------------|
| develop                  | 742.6            |
| prop map + free fn | 1280             |
| vector + free fn map       | 865             |
| AssocVec + free fn       | 972            |

Profiling of develop
image

Profiling of robinhashmap
image

Profiling of vector
image

Profiling of AssocVector
image

kks32:
Profiling of AssocVector

image

kks32:
After considering the speed-ups and different approaches, the use of scalar and vector properties still is 30% slower than using private variables. The advantage of scalar/vector properties is we'll have fewer functions and less repetition of code. On the other hand, defining private variables and deriving particle classes has more function definitions and repeating code, but is faster. I'd like to propose a vote, use emojis to vote between different options, please use comments below if you like to elaborate on your thoughts. @bodhinandach @tianchiTJ @thiagordonho @cw646 @ezrayst @jgiven100 @WeijianLIANG

1. Use derived particle types with private variables (:rocket:)
2. Solid particles will remain as is, while we add scalar and vector properties to derived particle types (:smile:) PR #673 
3. Cleaner but slower design with free functions and scalar/vector properties for all particle types (:heart:) PR #654
4. Go back to the drawing board (:eyes:)

WeijianLiang:
Thanks Krishna, I would prefer the use of scalar and vector properties which has a more clear structure.

kks32:
@WeijianLIANG is that option 2 or 3? Could you please vote with a reaction emoji :smile:  for option 2 and :heart:  for option 3. Thanks!

bodhinandach:
@kks32 I am voting for 3 in the long term as we found other methods to improve the speed, but as for now, I think option 2 is the best.

yliang-sn:
> After considering the speed-ups and different approaches, the use of scalar and vector properties still is 30% slower than using private variables. The advantage of scalar/vector properties is we'll have fewer functions and less repetition of code. On the other hand, defining private variables and deriving particle classes has more function definitions and repeating code, but is faster. I'd like to propose a vote, use emojis to vote between different options, please use comments below if you like to elaborate on your thoughts. @bodhinandach @tianchiTJ @thiagordonho @cw646 @ezrayst @jgiven100 @WeijianLIANG
> 
> 1. Use derived particle types with private variables (🚀)
> 2. Solid particles will remain as is, while we add scalar and vector properties to derived particle types (😄)
> 3. Cleaner but slower design with free functions and scalar/vector properties for all particle types (❤️)
> 4. Go back to the drawing board (👀)

For 2, are we just dealing with the scalar and vector properties to the derived particle types? If we have other types of quantities, do we still use the private variables?

kks32:
@yliang-sn everything will be held within scalar/vector/tensor/boolean

yliang-sn:
> @yliang-sn everything will be held within scalar/vector/tensor/boolean

Can I vote, too? If the private variables have the best efficiency, I prefer to 1. The structure is also clear enough to be read and extended. However, 2 is also fine for me.

kks32:
@yliang-sn yes, I couldn't find your handle, please vote.

kks32:
@thiagordonho @tianchiTJ @cw646 @WeijianLIANG could you please use emoji reactions to vote on this comment please: https://github.com/cb-geo/mpm/pull/654#issuecomment-662429917

ezrayst:
Both @kks32 and @bodhinandach have been working hard on this issue so appreciate your work. I think it has been about 2 months now. 

However, I feel that we need a unanimous decision on this refactor, or at least a 2/3 decision. The reason being is I don't want us to be in this position within 6 months where we are thinking to refactor again. I think the timeline on this issue has shown that this is a lot of work and everyone's progress is halted. *Personally, if we are not even sure on (2) or (3), or anyone still has reservation on them, wouldn't it better to keep it as is (1)?* Yes, `particle_base` will be messy because of the many functions from `two_phase` particles, but at least we get the speedup. Yes we might need to refactor in the future, but isn't that the assumption that we will refactor again? Lastly, if you are sure for (2), might as well go for (3) so we don't come back to it again later. So my vote will be (3) now!

WeijianLiang:
@kks32  Thanks, I also vote for (2)

bodhinandach:
After considering @ezrayst comment, I changed my vote to option (3). I know that the code will be slower, but the code is structurally way better and tidier. Also, if currently by moving to (3) causing us to be 30% slower than develop, I think we should rethink any other way to improve the speed. For instance, PR #677 has already improved the speed by 4% (thanks to @kks32). I think we can regain the current performance in short time by improving efficiency in other parts.

yliang-sn:
After talking with Ezra and Nanda, I think i will change to vote for 3 if we can improve the efficiency in other aspects.

kks32:
Other containers like Abseil, folly, vector of vectors, static_vector, and small_vector implementations all gave slower or almost identical performance as the flat_map. Due to the 30% reduction in performance we are closing this issue. https://trello.com/c/fGF3MrKg/110-plan-for-next-week-notes

PR #655, opened 2020-06-29
Feature/contact/normal vector

thiagordonho:
**Describe the PR**
Implementation to compute the normal unit vector at every node with more than one materials. In order to do this, a function in `Particle` class to map the domain gradients (gradient of volumes) is implemented. 
```
    const auto& gradient = volume_ * dn_dx_.row(i);
    nodes_[i]->update_property(true, "domain_gradients", gradient, material_id_,
                               Tdim);
  }
```

Then, the normal unit vector is calculated at every node by means of a function implemented in the `Node` class computing the unit vector as the domain gradient divided by its norm.
```
    const auto& normal_unit_vector =
        (1 / domain_gradient.norm()) * domain_gradient;
```

**Related Issues/PRs**

### DIFF
--- a/include/node.tcc
+++ b/include/node.tcc
@@ -608,10 +608,11 @@ void mpm::Node<Tdim, Tdof,
   std::lock_guard<std::mutex> guard(node_mutex_);
   for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
     // calculte the normal unit vector
-    const auto& domain_gradient =
+    VectorDim domain_gradient =
         property_handle_->property("domain_gradients", prop_id_, *mitr, Tdim);
-    const auto& normal_unit_vector =
-        (1 / domain_gradient.norm()) * domain_gradient;
+    VectorDim normal_unit_vector = VectorDim::Zero();
+    if (domain_gradient.norm() > std::numeric_limits<double>::epsilon())
+      normal_unit_vector = domain_gradient.normalized();
 
     // assign nodal-multimaterial normal unit vector to property pool
     property_handle_->assign_property("normal_unit_vectors", prop_id_, *mitr,
--- a/include/node.tcc
+++ b/include/node.tcc
@@ -599,3 +599,22 @@ void mpm::Node<Tdim, Tdof,
                                       separation_vector, Tdim);
   }
 }
+
+//! Compute multimaterial normal unit vector
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+void mpm::Node<Tdim, Tdof,
+               Tnphases>::compute_multimaterial_normal_unit_vector() {
+  // Iterate over all materials in the material_ids set
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    // calculte the normal unit vector
+    const auto& domain_gradient =
+        property_handle_->property("domain_gradients", prop_id_, *mitr, Tdim);
+    const auto& normal_unit_vector =
+        (1 / domain_gradient.norm()) * domain_gradient;
+
+    // assign nodal-multimaterial normal unit vector to property pool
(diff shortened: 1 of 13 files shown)


include/node.tcc
```diff
@@ -599,3 +599,22 @@ void mpm::Node<Tdim, Tdof,
...
+void mpm::Node<Tdim, Tdof,
+               Tnphases>::compute_multimaterial_normal_unit_vector() {
+  // Iterate over all materials in the material_ids set
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    // calculte the normal unit vector
+    const auto& domain_gradient =
+        property_handle_->property("domain_gradients", prop_id_, *mitr, Tdim);
+    const auto& normal_unit_vector =
+        (1 / domain_gradient.norm()) * domain_gradient;
```
bodhinandach: I would check if `domain_gradient.norm()` if not zero, and no need to add `&` in `auto`.
```suggestion
VectorDim normal;
if (domain_gradient.norm() > std::numeric_limits<double>::epsilon())
   normal = (1 / domain_gradient.norm()) * domain_gradient;
else
   normal.setZero();
```

ezrayst: Can we use `normal = domain_gradient.normalize() instead?

include/node.tcc
```diff
@@ -599,3 +599,22 @@ void mpm::Node<Tdim, Tdof,
...
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+void mpm::Node<Tdim, Tdof,
+               Tnphases>::compute_multimaterial_normal_unit_vector() {
+  // Iterate over all materials in the material_ids set
+  std::lock_guard<std::mutex> guard(node_mutex_);
+  for (auto mitr = material_ids_.begin(); mitr != material_ids_.end(); ++mitr) {
+    // calculte the normal unit vector
+    const auto& domain_gradient =
+        property_handle_->property("domain_gradients", prop_id_, *mitr, Tdim);
+    const auto& normal_unit_vector =
```
ezrayst:
```suggestion
    VectorDim normal_unit_vector = VectorDim::Zero();
    if (domain_gradient.norm() > std::numeric_limits<double>::epsilon())
      normal_unit_vector = domain_gradient.normalize();
```

kks32 (changes requested):
Please complete the requested changes.

include/node.tcc
thiagordonho: I did not add the else condition in my next commit because I initialised the normal vector with VectorDim::Zero() as @ezrayst suggested. Thank you for the other change requests!

bodhinandach (approved):

ezrayst (approved):
Thanks @thiagordonho, not sure whether we should use `.normalized()` or `.normalize()` - but I think the different is that it returns a copy for the former and modify the current variable for the latter. So I think you are right in using the former!

kks32 (approved):

include/particles/particle.tcc
```diff
@@ -536,6 +536,22 @@ void mpm::Particle<Tdim>::map_multimaterial_displacements_to_nodes() noexcept {
...
+template <unsigned Tdim>
+void mpm::Particle<
+    Tdim>::map_multimaterial_domain_gradients_to_nodes() noexcept {
+  // Check if particle volume is set
+  assert(volume_ != std::numeric_limits<double>::max());
+
+  // Map domain gradients to nodal property. The domain gradients is defined as
+  // the gradient of the particle volume
+  for (unsigned i = 0; i < nodes_.size(); ++i) {
+    const auto& gradient = volume_ * dn_dx_.row(i);
```
kks32:
```suggestion
    Eigen::Matrix<double, Tdim, 1> gradient;
    for (unsigned j = 0; j < Tdim; ++j)
        gradient[j] += dn_dx_[i, j] * volume;
```

PR #657, opened 2020-06-30
[Hotfix] Particle on edge error

### LINKED ISSUE #656: Error in stress computation if there is particle lying on edge
**Describe the bug**
The stress computation shows a wrong result if there is a particle lying on an element edge/boundary. This is perhaps one of the major causes where the stress and the strain in the simulation sometimes show a singularity error. This happens in all constitutive model.

image

The error is resulted due to the incapability of preserving the rigid-body motion zero energy modes when a particle is lying exactly on element edge. See the following math for a single particle moving in rigid-body motion with v=+1 m/s:

image (1)

The following simulation is produced for an elastic particle with E=1000, v=-1, dt=1:

test2

**To Reproduce**
1. Compile 'cmake -DCMAKE_CXX_COMPILER=g++ .. && make -j8'
2. Run on one node with './mpm/build/mpm -f \path\to\input\file -i input-file.json'
3. Run any simulation where a particle is lying exactly on edge.

**Expected behavior**
Rigid body motion should have no stress. The corrected version:

test_correct

 - Any OS
 - Branch develop

**Additional context**
The easiest way to fix this is to implement a nudge of `std::numeric_limits<double>::epsilon()` while computing reference location.

bodhinandach:
**Describe the PR**
A nudge is implemented inside `is_point_in_cell()`, shifting the relative coordinate using `std::numeric_limit<double>::epsilon()`, if a particle is lying exactly on element edge. 

**Related Issues/PRs**
Fix #656.

### DIFF
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -659,11 +659,12 @@ inline Eigen::Matrix<double, Tdim, 1>
   // Trial guess for NR
   Eigen::Matrix<double, Tdim, 1> nr_xi = geometry_xi;
   // Check if the first trial xi is just outside the box
+  const double nudge_tolerance = 1.E-12;
   for (unsigned i = 0; i < nr_xi.size(); ++i) {
     if (nr_xi(i) < -1. && nr_xi(i) > -1.001)
-      nr_xi(i) = -1. + std::numeric_limits<double>::epsilon();
+      nr_xi(i) = -1. + nudge_tolerance;
     else if (nr_xi(i) > 1. && nr_xi(i) < 1.001)
-      nr_xi(i) = 1. - std::numeric_limits<double>::epsilon();
+      nr_xi(i) = 1. - nudge_tolerance;
   }
 
   // Maximum iterations of newton raphson
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -661,9 +661,9 @@ inline Eigen::Matrix<double, Tdim, 1>
   // Check if the first trial xi is just outside the box
   for (unsigned i = 0; i < nr_xi.size(); ++i) {
     if (nr_xi(i) < -1. && nr_xi(i) > -1.001)
-      nr_xi(i) = -0.999999999999;
+      nr_xi(i) = -1. + std::numeric_limits<double>::epsilon();
     else if (nr_xi(i) > 1. && nr_xi(i) < 1.001)
-      nr_xi(i) = 0.999999999999;
+      nr_xi(i) = 1. - std::numeric_limits<double>::epsilon();
   }
 
   // Maximum iterations of newton raphson
--- a/include/cell.tcc
+++ b/include/cell.tcc
@@ -336,15 +336,28 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
 
   // Check if the transformed coordinate is within the unit cell:
   // between 0 and 1-xi(1-i) if the element is a triangle, and between
-  // -1 and 1 if otherwise
+  // -1 and 1 if otherwise. Also, check if the transformed coordinate lies
+  // exactly on cell edge.
+  const double tolerance = std::numeric_limits<double>::epsilon();
   if (this->element_->corner_indices().size() == 3) {
-    for (unsigned i = 0; i < (*xi).size(); ++i)
+    for (unsigned i = 0; i < (*xi).size(); ++i) {
       if ((*xi)(i) < 0. || (*xi)(i) > 1. - (*xi)(1 - i) || std::isnan((*xi)(i)))
         status = false;
+      else {
+        if ((*xi)(i) < tolerance) (*xi)(i) = tolerance;
+        if ((*xi)(i) > 1. - (*xi)(1 - i) - tolerance)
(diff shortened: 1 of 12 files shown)


include/cell.tcc
```diff
@@ -336,15 +336,28 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
 
   // Check if the transformed coordinate is within the unit cell:
   // between 0 and 1-xi(1-i) if the element is a triangle, and between
-  // -1 and 1 if otherwise
+  // -1 and 1 if otherwise. Also, check if the transformed coordinate lies
+  // exactly on cell edge.
+  const double tolerance = std::numeric_limits<double>::epsilon();
```
ezrayst, line 341: Not sure this would be enough to prevent the blowing up. Should we use a higher tolerance?

include/cell.tcc
```diff
@@ -336,15 +336,28 @@ inline bool mpm::Cell<Tdim>::is_point_in_cell(
...
   if (this->element_->corner_indices().size() == 3) {
```
ezrayst, line 342: @kks32 should we do inline for triangular and quadrilateral elements? Too many `if` statements here

ezrayst (commented):
Thanks for this. In the issue, can you please add the ones with many particles within the cell for comparison?

bodhinandach:
@ezrayst Is this what you want to see:

Without correction (develop):
many

With correction (this branch):
many_cor

include/cell.tcc
bodhinandach: It should be ok. I checked it.

tests/cell_test.cc
```diff
@@ -223,11 +223,15 @@ TEST_CASE("Cell is checked for 2D case", "[cell][2D]") {
         point << 0., 0.;
         REQUIRE(cell->point_in_cartesian_cell(point) == true);
         REQUIRE(cell->is_point_in_cell(point, &xi) == true);
+        REQUIRE(xi(0) == -1. + std::numeric_limits<double>::epsilon());
+        REQUIRE(xi(1) == -1. + std::numeric_limits<double>::epsilon());
```
kks32: Comparing doubles this way may not work all the time, we may have to do something like this?

```
template<class T>
typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type
    almost_equal(T x, T y, int ulp)
{
    // the machine epsilon has to be scaled to the magnitude of the values used
    // and multiplied by the desired precision in ULPs (units in the last place)
    return std::fabs(x-y) <= std::numeric_limits<T>::epsilon() * std::fabs(x+y) * ulp
        // unless the result is subnormal
        || std::fabs(x-y) < std::numeric_limits<T>::min();
}
```

or alternatively, we could use the `Approx` call in `REQUIRE` macro with `epsilon` to check if the results are correct.

kks32 (changes requested):
Just a couple of minor things.

Could we also fix these, while we are at it to be consistent and change to machepsilon?
https://github.com/cb-geo/mpm/blob/develop/include/cell.tcc#L650-L653

tests/cell_test.cc
bodhinandach: Thanks for the suggestion, I decided to use `Approx` with machepsilon. Though, this also allows the value to be machepsilon different from the provided value. In our example, this will return true if `xi(i) = -1`. That's the reason I initially did doubles comparison.

include/cell.tcc
```diff
@@ -648,9 +661,9 @@ inline Eigen::Matrix<double, Tdim, 1>
   // Check if the first trial xi is just outside the box
   for (unsigned i = 0; i < nr_xi.size(); ++i) {
     if (nr_xi(i) < -1. && nr_xi(i) > -1.001)
```
bodhinandach, line 664: @kks32 Do you want the if check to be modified to be -1 + machepsilon too?

ezrayst: I feel that we need to check for bigger actually, as the inaccuracy springs from affine transformation. However, I am not sure 1E-3 is the magic number.

ezrayst:
@bodhinandach, in the results you show with 3x3 material points, why does the error happen when the last column of the material points hit the edge (not the first column)? I feel that the problem happens if the material all the material points in that element lie on the edge - do you agree with me?

include/cell.tcc
ezrayst: I used 1E-3 because if the material point is really outside, NR would converge outside - so no problem. Should we have a variation on this?

kks32 (approved):

include/cell.tcc
```diff
@@ -648,9 +661,9 @@ inline Eigen::Matrix<double, Tdim, 1>
...
-      nr_xi(i) = -0.999999999999;
```
ezrayst, line 651: Can we not use machine epsilon here? Maybe keep it at 1E-12? Would that be OK? Because this is not checking whether the particle is at the edge, but just making sure that there is a good starting point for the NR iterations. So an arbitrary number that makes sense should be fine.

bodhinandach: Done

ezrayst (approved):

PR #660, opened 2020-07-02
Fix hardening in NorSand

ezrayst:
**Describe the PR**
This fix ensures axissymetric pdstrain as shown and expected.

image

### DIFF
--- a/include/materials/norsand.tcc
+++ b/include/materials/norsand.tcc
@@ -353,13 +353,12 @@ void mpm::NorSand<Tdim>::compute_plastic_tensor(const Vector6d& stress,
 
   const double dF_dsigma_v = (dF_dsigma(0) + dF_dsigma(1) + dF_dsigma(2)) / 3;
   const double dF_dsigma_deviatoric =
-      std::sqrt(2. / 3.) *
-      std::sqrt(std::pow(dF_dsigma(0) - dF_dsigma_v, 2) +
-                std::pow(dF_dsigma(1) - dF_dsigma_v, 2) +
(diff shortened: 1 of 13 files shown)


include/materials/norsand.tcc
```diff
@@ -357,8 +357,9 @@ void mpm::NorSand<Tdim>::compute_plastic_tensor(const Vector6d& stress,
       std::sqrt(std::pow(dF_dsigma(0) - dF_dsigma_v, 2) +
                 std::pow(dF_dsigma(1) - dF_dsigma_v, 2) +
                 std::pow(dF_dsigma(2) - dF_dsigma_v, 2) +
-                2 * std::pow(dF_dsigma(3), 2) + 2 * std::pow(dF_dsigma(4), 2) +
-                2 * std::pow(dF_dsigma(5), 2));
+                std::pow(dF_dsigma(3), 2) / 2. +
+                std::pow(dF_dsigma(4), 2) / 2. + std::pow(dF_dsigma(5), 2)) /
+      2.;
```
bodhinandach: I smell some typo here... Shouldn't it be: 
```suggestion
                std::pow(dF_dsigma(4), 2) / 2. + std::pow(dF_dsigma(5), 2) / 2.);
```
Why the `std::pow(dF_dsigma(5), 2) ` is not divided by 2? And why all the `std::sqrt()` is divided by 2. in the end?

ezrayst: Yes fixing... realized that and that is why results are a bit weird

bodhinandach (approved):

jgiven100:
This looks good to solve axisymmetric `pdstrain` issues previously observed :+1:

PR #661, opened 2020-07-04
Feature/parallel/openmp

kks32:
**Describe the PR**
This PR replaces existing TBB parallelization with OpenMP. OpenMP is found to consistently better, if not the same as TBB. As there is more support for OpenMP, there is a further possibility of improving the performance of the code. 

**Related Issues/PRs**

**Additional context**
OpenMP/TBB speed ratio is almost always greater than 1, showing OpenMP is faster than TBB.

image


include/mesh.tcc
```diff
@@ -1058,27 +1045,18 @@ template <typename Toper>
...
-        tbb::blocked_range<int>(size_t(0), size_t(set.size()), tbb_grain_size_),
-        [&](const tbb::blocked_range<int>& range) {
-          for (int i = range.begin(); i != range.end(); ++i) {
-            unsigned id = set[i];
-            if (map_particles_.find(id) != map_particles_.end())
-              oper(map_particles_[id]);
-          }
-        },
-        tbb::simple_partitioner());
+#ifdef _OPENMP
```
bodhinandach: Do you really need this ifdef? Shouldn't it be automatic?

bodhinandach:
This might be breaking for all the development using TBB, I think it's better to notify everyone before merging.

src/main.cc
```diff
@@ -32,11 +31,11 @@ int main(int argc, char** argv) {
...
-    // Set TBB threads
-    unsigned nthreads = tbb::task_scheduler_init::default_num_threads();
-    // If number of TBB threads are positive set to nthreads
-    if (io->nthreads() > 0) nthreads = io->nthreads();
-    tbb::task_scheduler_init init(nthreads);
+    // If number of threads are positive set to nthreads
+    unsigned nthreads = io->nthreads();
+#ifdef _OPENMP
+    omp_set_num_threads(nthreads > 0 ? nthreads : omp_get_max_threads());
+#endif
```
bodhinandach, line 38: Is `io->nthreads()` equal to `XX` set by `export OMP_NUM_THREADS=XX` ? Or the one you specified after `-p` or `--parallel` option?

include/mesh.h
```diff
@@ -14,10 +14,11 @@
 #ifdef USE_MPI
 #include "mpi.h"
 #endif
-// TBB
-#include <tbb/parallel_for.h>
-#include <tbb/parallel_for_each.h>
-
+// OpenMP
+#ifdef _OPENMP
```
bodhinandach, line 18: Is this ifdef added because of clang? Do we need to automatically activate OMP if clang is used? I mean by `-fopenmp=libomp`, see: https://releases.llvm.org/3.7.0/tools/clang/docs/ReleaseNotes.html#openmp-support

bodhinandach (commented):
Seems good to me! What kind of testing are you planning to have? Thanks @kks32 

**Update:** testings are automatically replacing TBB tests

src/main.cc
kks32: if number of threads are not specified, we use the maximum number of threads. If specified using `-p` or `--parallel` we use those instead.

include/mesh.h
kks32: This is added not specifically for clang, we have FindOpenMP so it will add the appropriate flag. Clang needs to be compiled with OpenMP separately. These include guards are there if OpenMP is not found, irrespective of the compiler.

src/main.cc
bodhinandach: But can one still set the thread by using `OMP_NUM_THREADS`?

kks32: Yes, that will affect the `omp_get_max_threads()` parameter and will have the same effect as `-p`

thiagordonho (approved):

src/main.cc
bodhinandach: Understood! Thanks for the clarification

bodhinandach (approved):
I am good with this. Thanks @kks32

include/loads_bcs/constraints.tcc
```diff
@@ -64,29 +55,21 @@ bool mpm::Constraints<Tdim>::assign_nodal_velocity_constraints(
 template <unsigned Tdim>
 bool mpm::Constraints<Tdim>::assign_nodal_frictional_constraint(
     int nset_id, const std::shared_ptr<mpm::FrictionConstraint>& fconstraint) {
-  bool status = false;
+  bool status = true;
```
ezrayst, line 58: Do we need `bool` for this function, or `void` would suffice?
Here is where it is used: https://github.com/cb-geo/mpm/blob/8f19a4e1fae9a8f1afef0e0e5f723881008e3560/include/solvers/mpm_base.tcc#L863

include/loads_bcs/constraints.tcc
```diff
@@ -36,7 +29,7 @@ template <unsigned Tdim>
 bool mpm::Constraints<Tdim>::assign_nodal_velocity_constraints(
     const std::vector<std::tuple<mpm::Index, unsigned, double>>&
         velocity_constraints) {
-  bool status = false;
+  bool status = true;
```
ezrayst, line 32: Same comment whether we need `bool` for this function

ezrayst (approved):
Thanks for this, @kks32. I wonder what causes the less performance of openmp with 1 and 32 nodes. It does not seem to make sense for me.

cw646 (approved):

PR #666, opened 2020-07-09
Remove fedora command in partio installation commands

ezrayst:
**Describe the PR**
Just cleaning up installation commands to exclude one fedora installation line in ubuntu installation guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,6 @@ sudo make install
 ### Partio for Houdini SFX Visualization
 
 ```shell
-sudo dnf install -y libnsl freeglut freeglut-devel
 mkdir -p ~/workspace && cd ~/workspace/ && git clone https://github.com/wdas/partio.git && \
     cd partio && cmake . && make
 ```


bodhinandach (commented):
Hmmm... I think we need those libraries though to run Partio?

ezrayst:
Not in the this section

kks32 (commented):
We need these libraries, and this section is not under ubuntu, This is a separate section for compiling with Partio. This has the same section hierarchy as Fedora / Ubuntu.

ezrayst:
Yes, but we have the libraries in the dependencies before right - like the line for Fedora? And if this is for both Ubuntu and Fedora, these lines cannot be used for Ubuntu anyways.

kks32:
If someone wants to install partio these are dependencies, so it's good to have them there. Also, our recommended OS is Fedora, so all installation instructions are with respect to Fedora.

ezrayst:
But isn't it a duplicate of what you have in the dependency installation?

bodhinandach:
@kks32 One question: these dependencies are not available in Ubuntu: `libnsl freeglut freeglut-devel`. Do we need anything to install partio in Ubuntu?

bodhinandach (approved):
I checked that the installation of Partio in Ubuntu does not require the specified libraries/dependencies. I am okay to remove them.

kks32 (approved):

PR #667, opened 2020-07-09
Add return value for nodal friction and velocity constraints

ezrayst:
**Describe the PR**
The current assign nodal constraints in `mpm_base` for both velocity and friction do not throw when the bool returns `false`. Therefore, this PR returns the bool and throw if `false`.

### DIFF
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -815,7 +815,7 @@ void mpm::MPMBase<Tdim>::nodal_velocity_constraints(
               nset_id, velocity_constraint);
           if (!velocity_constraints)
             throw std::runtime_error(
-                "Assign nodal velocity constraint is not properly assigned");
+                "Nodal velocity constraint is not properly assigned");
         }
       }
     } else
@@ -868,7 +868,7 @@ void mpm::MPMBase<Tdim>::nodal_frictional_constraints(
                   nset_id, friction_constraint);
           if (!friction_constraints)
             throw std::runtime_error(
-                "Assign nodal friction constraint is not properly assigned");
+                "Nodal friction constraint is not properly assigned");
         }
       }
     } else
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -811,7 +811,7 @@ void mpm::MPMBase<Tdim>::nodal_velocity_constraints(
           // Add velocity constraint to mesh
           auto velocity_constraint =
               std::make_shared<mpm::VelocityConstraint>(nset_id, dir, velocity);
-          velocity_constraints = constraints_->assign_nodal_velocity_constraint(
+          bool velocity_constraints = constraints_->assign_nodal_velocity_constraint(
               nset_id, velocity_constraint);
           if (!velocity_constraints)
             throw std::runtime_error(
@@ -863,7 +863,7 @@ void mpm::MPMBase<Tdim>::nodal_frictional_constraints(
           // Add friction constraint to mesh
           auto friction_constraint = std::make_shared<mpm::FrictionConstraint>(
               nset_id, dir, sign_n, friction);
-          friction_constraints =
+          bool friction_constraints =
               constraints_->assign_nodal_frictional_constraint(
                   nset_id, friction_constraint);
           if (!friction_constraints)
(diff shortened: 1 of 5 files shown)


bodhinandach (dismissed):

include/solvers/mpm_base.tcc
```diff
@@ -811,8 +811,11 @@ void mpm::MPMBase<Tdim>::nodal_velocity_constraints(
           // Add velocity constraint to mesh
           auto velocity_constraint =
               std::make_shared<mpm::VelocityConstraint>(nset_id, dir, velocity);
-          constraints_->assign_nodal_velocity_constraint(nset_id,
-                                                         velocity_constraint);
+          bool velocity_constraints = constraints_->assign_nodal_velocity_constraint(
+              nset_id, velocity_constraint);
+          if (!velocity_constraints)
+            throw std::runtime_error(
+                "Assign nodal velocity constraint is not properly assigned");
```
```suggestion
                "Nodal velocity constraint is not properly assigned");
```
otherwise the sentence is confusing.

include/solvers/mpm_base.tcc
```diff
@@ -860,8 +863,12 @@ void mpm::MPMBase<Tdim>::nodal_frictional_constraints(
...
           auto friction_constraint = std::make_shared<mpm::FrictionConstraint>(
               nset_id, dir, sign_n, friction);
-          constraints_->assign_nodal_frictional_constraint(nset_id,
-                                                           friction_constraint);
+          bool friction_constraints =
+              constraints_->assign_nodal_frictional_constraint(
+                  nset_id, friction_constraint);
+          if (!friction_constraints)
+            throw std::runtime_error(
+                "Assign nodal friction constraint is not properly assigned");
```
```suggestion
                "Nodal friction constraint is not properly assigned");
```

kks32 (changes requested):

kks32:
@ezrayst Thanks. Could you please test the exception: https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#exceptions

kks32 (approved):

kks32 (commented):
Thanks Ezra!

ezrayst:
I want to discuss one more thing regarding this, let's use velocity constraints as example:
- `mpm_explicit` calls `bool` `initialise_mesh()` ([link](https://github.com/cb-geo/mpm/blob/a03af6cc5afa14c1b9e832b741e78a62eac89e46/include/solvers/mpm_base.tcc#L115))
- which in `mpm_base` calls `void` `nodal_velocity_constraints` ([link](https://github.com/cb-geo/mpm/blob/a03af6cc5afa14c1b9e832b741e78a62eac89e46/include/solvers/mpm_base.tcc#L182)) (I think the idea is to check for `NOTHROW` here right?)
- which in `constraints` calls `assign_nodal_velocity_constraint` ([link](https://github.com/cb-geo/mpm/blob/9ed027de4fb58f4edf5866b1f25d08a4fbae887a/include/loads_bcs/constraints.tcc#L3)). This function already has a throw that is tested. 

I do not think we need to check for `NOTHROW` for `nodal_velocity_constraints` in `mpm_base` anymore. I would leave it as is.

Further, this calls to change this function to a `void` instead of `bool`. What do you think?

thiagordonho (approved):

kks32:
> I want to discuss one more thing regarding this, let's use velocity constraints as example:
> 
>     * `mpm_explicit` calls `bool` `initialise_mesh()` ([link](https://github.com/cb-geo/mpm/blob/a03af6cc5afa14c1b9e832b741e78a62eac89e46/include/solvers/mpm_base.tcc#L115))
> 
>     * which in `mpm_base` calls `void` `nodal_velocity_constraints` ([link](https://github.com/cb-geo/mpm/blob/a03af6cc5afa14c1b9e832b741e78a62eac89e46/include/solvers/mpm_base.tcc#L182)) (I think the idea is to check for `NOTHROW` here right?)
> 
>     * which in `constraints` calls `assign_nodal_velocity_constraint` ([link](https://github.com/cb-geo/mpm/blob/9ed027de4fb58f4edf5866b1f25d08a4fbae887a/include/loads_bcs/constraints.tcc#L3)). This function already has a throw that is tested.
> 
> 
> I do not think we need to check for `NOTHROW` for `nodal_velocity_constraints` in `mpm_base` anymore. I would leave it as is.
> 
> Further, this calls to change this function to a `void` instead of `bool`. What do you think?

If we want to address this issue, these functions should throw and abort on MPM main. There is no reason to continue when things fail.

bodhinandach (approved):
I think I am good with this. Unless @kks32 has any further request. Thanks @ezrayst

PR #674, opened 2020-07-22
Feature/material_id

### LINKED ISSUE #671: Add "material_sets" to entity_set.json
# Add `"material_sets"` to `entity_set.json`

## Summary

Adding `"material_sets"` to the `entity_set.json` will eventually be used to define multiple material properties for particles from a single particle input file (`particles.txt`). Using a flag to indicate when to use the `"material_sets"` listed within the `entity_sets.json` allows for either  manual particle definition **or** multiple particle generators (current implementation). As a result user will have 2 viable methods for developing a model with more than one material. 

## Motivation

* Including `"material_sets"` in the `entity_set.json` will eventually be used to define mutliple materials for particles from a single particle input file (`particles.txt`).
* Currently, `"material_id"` is defined by the particle generator. If `n` materials are desired, multiple particle input files (`particles_0.txt`, `...` , `particles_n.txt`) are necessary.
* For some geometries (e.g., axisymmetric), a single particle input file (`particles.txt`) and location file (`particles_cells.txt`) is easier to produce although multiple materials are necessary. 

## Design Detail

The existing format of `"particle_sets"` will be adopted for `"material_sets"`. The `"set"` will be defined by `particle_id`:
```
{
    "material_sets": [
        {
        "id": 0,
        "set": [p0_0, ... , p0_n]
        },
        {
        "id": 1,
        "set": [p1_0, ... , p1_n]
        }
    ]
}

p0_i is the id of a particle within the set with id 0.
```

The current method of using a different particle generator for each desired material is convinent if additioanl particles are added since there is no need to regenerate the previsouly determined `particle_sets`. Additionally, particle sets can be developed automatically by the generator. **The benefits of using multiple generators should remain.**

Therefore, the eventual implementation of using `"material_sets"` from the `entity_set.json` should rely on a flag in the particle generator such as:
```
"material_id": -1
```

If `-1`, the flag will initiate an if-statement at [L1785](https://github.com/cb-geo/mpm/blob/b14691d68285b3a4dd8f265e85cc471c5cffec5b/include/mesh.tcc#L1785) of `mesh.tcc`. Within this section of code will be a for-loop moving through all particles:
* The `material_id` will be defined from the `entity_set.json` instead of the scalar user input
* The function `create_particles` will be called at the end of each loop

> Further details of this implementation will covered in a subsequent RFC!

## Drawbacks

* Data type of `material_id` must be updated from `unsigned` to `int`
* `mesh.tcc` currently does not call the `entity_set.json` file
* Using a loop to define `material_id` will have increased cost

## Rationale and Alternatives

If `"material_sets"` is added to the `entity_set.json`, this leads to defining multiple `material_id` for any given particle input file (`particles.txt`). Using a flag to achieve this result in current input files remaining usable while offering increased options to the user. 

One alternative considered was to define the `"set"` by the `"pset_id"` instead of the `"particle_id"`:
```
{
    "material_sets": [
        {
        "id": 0,
        "set": [ps0_0, ... , p0_n]
        },
        {
        "id": 1,
        "set": [ps1_0, ... , p1_n]
        }
    ]
}

ps0_i is the id of a particle set within the set with id 0.
```
This was not selected since the particle generator currently defines the same value for `pset_id` for all particles from any given input file. Thus, if the `material_id` was defined by the `pset_id`, the `pset_id` would additionally need to be updated (i.e., 2 loops instead of 1). 

If not implement, users will not have the option to manually define material properties via the `particle_id`. In some instances where only a few particles are desired to be changed, it would be much quicker to define material through the `entity_set.json` instead of developing an additional input particle file. 

> Consider an axisymmetric model with a very large number of elements where failure is initialize through selecting specific elements to weaken. Loading all particles as a single input file and manually selecting the 10 elements to have different material properties is much quicker than developing 2 input files. If the model is split into 2 seperate input files, it is perhaps no longer axisymmetric and thus requires more pre-processing. Additionally, creating the the `particle_cells` file becomes more difficult as more input files are used.   

## Prior Art

The `"material_id": -1` flag is taken from the current `cset` implementation on [L482](https://github.com/cb-geo/mpm/blob/b14691d68285b3a4dd8f265e85cc471c5cffec5b/include/mesh.tcc#L482) of `mesh.tcc`.

## Unresolved questions

More details regarding the eventual implementation of using `"material_sets"` from the `entity_set.json` to define `material_id` will be in a future RFC. Reading comments about expanding the information included in the `entity_set.json` file seems like a necessary first step. 

## Changelog

>

jgiven100:
**Describe the PR**
Adding the option to update `material_id` based on `pset_id` for a single input file (i.e. `particles.txt`).

**Related Issues/PRs**
https://github.com/cb-geo/mpm/issues/671 

**Additional context**
Input `JSON` must include new section **"material_sets"**:
```
"material_sets":[
    {
      "material_id": 0,
      "pset_id": 0
    },
    {
      "material_id": 1,
      "pset_id": 1
    }
]
```
And the **"mesh:"** section must include:
```
"update_materials": true
```

### DIFF
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -330,19 +330,9 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
                        .count());
 
     // Material id update using particle sets
-    bool update_materials = false;
-    try {
-      update_materials = mesh_props["update_materials"].template get<bool>();
-    } catch (std::exception& exception) {
-      console_->warn(
-          "{} #{}: Update materials, not specified setting default as false",
-          __FILE__, __LINE__, exception.what());
-      update_materials = false;
-    }
-
     try {
       auto material_sets = io_->json_object("material_sets");
-      if (!material_sets.empty() && update_materials) {
+      if (!material_sets.empty()) {
         for (const auto& material_set : material_sets) {
           unsigned material_id =
               material_set["material_id"].template get<unsigned>();
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -269,6 +269,11 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
 
     // Get material sets
     auto material_sets = io_->json_object("material_sets");
+    
+    if(!material_sets.empty())
+      for(const auto& material_set : material_sets){
+
+      }
 
     auto particles_gen_end = std::chrono::steady_clock::now();
     console_->info("Rank {} Generate particles: {} ms", mpi_rank,
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -267,6 +267,9 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
       if (!gen_status) status = false;
     }
 
+    // Get material sets
+    auto material_sets = io_->json_object("material_sets");
+
     auto particles_gen_end = std::chrono::steady_clock::now();
     console_->info("Rank {} Generate particles: {} ms", mpi_rank,
                    std::chrono::duration_cast<std::chrono::milliseconds>(
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -267,19 +267,6 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
       if (!gen_status) status = false;
     }
 
-    // Get material sets
-    auto material_sets = io_->json_object("material_sets");
-    
-    if(!material_sets.empty() && mesh_props["update_materials"] == true) 
-      for(const auto& material_set : material_sets){
-        unsigned material_id = material_set["material_id"].template get<unsigned>();
-        unsigned pset_id = material_set["pset_id"].template get<unsigned>();
-        // Update material_id for particles in each pset
-        mesh_->iterate_over_particle_set(pset_id,
-          std::bind(&mpm::ParticleBase<Tdim>::assign_material,
-          std::placeholders::_1, materials_.at(material_id)));
-      }
-
     auto particles_gen_end = std::chrono::steady_clock::now();
     console_->info("Rank {} Generate particles: {} ms", mpi_rank,
                    std::chrono::duration_cast<std::chrono::milliseconds>(
@@ -341,6 +328,36 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
                    std::chrono::duration_cast<std::chrono::milliseconds>(
                        particles_volume_end - particles_volume_begin)
                        .count());
+
+    // Get material sets
+    bool update_materials = false;
+    try {
+      update_materials = mesh_props["update_materials"].template get<bool>();
+    } catch (std::exception& exception) {
+      console_->warn(
+          "{} #{}: Update materials, not specified setting default as false",
+          __FILE__, __LINE__, exception.what());
+      update_materials = false;
+    }
+
+    try {
+      auto material_sets = io_->json_object("material_sets");
+      if (!material_sets.empty() && update_materials) {
+        for (const auto& material_set : material_sets) {
+          unsigned material_id =
+              material_set["material_id"].template get<unsigned>();
+          unsigned pset_id = material_set["pset_id"].template get<unsigned>();
+          // Update material_id for particles in each pset
+          mesh_->iterate_over_particle_set(
+              pset_id,
+              std::bind(&mpm::ParticleBase<Tdim>::assign_material,
+                        std::placeholders::_1, materials_.at(material_id)));
+        }
+      }
+    } catch (std::exception& exception) {
+      console_->warn("{} #{}: Material sets, not specified", __FILE__, __LINE__,
+                     exception.what());
+    }
 
   } catch (std::exception& exception) {
     console_->error("#{}: MPM Base generating particles: {}", __LINE__,
@@ -824,8 +841,9 @@ void mpm::MPMBase<Tdim>::nodal_velocity_constraints(
           // Add velocity constraint to mesh
           auto velocity_constraint =
               std::make_shared<mpm::VelocityConstraint>(nset_id, dir, velocity);
-          bool velocity_constraints = constraints_->assign_nodal_velocity_constraint(
-              nset_id, velocity_constraint);
+          bool velocity_constraints =
+              constraints_->assign_nodal_velocity_constraint(
+                  nset_id, velocity_constraint);
           if (!velocity_constraints)
             throw std::runtime_error(
                 "Nodal velocity constraint is not properly assigned");
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -272,7 +272,10 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
     
     if(!material_sets.empty())
       for(const auto& material_set : material_sets){
-
+        // Update material_id for particles in each pset
+        mesh_->iterate_over_particle_set(material_set["pset_id"], 
+          std::bind(&mpm::ParticleBase<Tdim>::assign_material,
+          std::placeholders::_1, materials_.at(material_set["material_id"])));
       }
 
     auto particles_gen_end = std::chrono::steady_clock::now();
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -272,10 +272,12 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
     
     if(!material_sets.empty() && mesh_props["update_materials"] == true) 
       for(const auto& material_set : material_sets){
+        unsigned material_id = material_set["material_id"].template get<unsigned>();
+        unsigned pset_id = material_set["pset_id"].template get<unsigned>();
         // Update material_id for particles in each pset
-        mesh_->iterate_over_particle_set(material_set["pset_id"], 
+        mesh_->iterate_over_particle_set(pset_id,
           std::bind(&mpm::ParticleBase<Tdim>::assign_material,
-          std::placeholders::_1, materials_.at(material_set["material_id"])));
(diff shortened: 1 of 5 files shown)


ezrayst:
@jgiven100 thanks for this. A few things from me:
- The circleci build passes, this means everything is in order including the formatting.
- A few test functions would be good to convince everyone that the code is working as it is. We can discuss in which file to do so.
- You might need to update the input file in our [documentation](https://github.com/cb-geo/mpm-doc). It would be the same process of making PR just like this, but no testing.

include/solvers/mpm_base.tcc
```diff
@@ -329,6 +329,36 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
                        particles_volume_end - particles_volume_begin)
                        .count());
 
+    // Get material sets
```
ezrayst: Can you change the comment to an appropriate one?

ezrayst (changes requested):
@jgiven100 thanks for this. A few things from me:
- The circleci build passes, this means everything is in order including the formatting.
- A few test functions would be good to convince everyone that the code is working as it is. We can discuss in which file to do so.
- You might need to update the input file in our [documentation](https://github.com/cb-geo/mpm-doc). It would be the same process of making PR just like this, but no testing.

include/solvers/mpm_base.tcc
```diff
@@ -329,6 +329,36 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
...
+    bool update_materials = false;
+    try {
+      update_materials = mesh_props["update_materials"].template get<bool>();
```
kks32: Why do we need a separate `update_materials` flag? Can we not just use the presence of `material_sets` to overwrite assigned materials?

include/solvers/mpm_base.tcc
```diff
@@ -329,6 +329,36 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
...
+          unsigned pset_id = material_set["pset_id"].template get<unsigned>();
+          // Update material_id for particles in each pset
+          mesh_->iterate_over_particle_set(
+              pset_id,
+              std::bind(&mpm::ParticleBase<Tdim>::assign_material,
+                        std::placeholders::_1, materials_.at(material_id)));
+        }
+      }
+    } catch (std::exception& exception) {
+      console_->warn("{} #{}: Material sets, not specified", __FILE__, __LINE__,
```
```suggestion
      console_->warn("{} #{}: Material sets are not specified", __FILE__, __LINE__,
```

kks32 (changes requested):
Thanks @jgiven100 I don't think we need the boolean flag.

include/solvers/mpm_base.tcc
jgiven100: updated to:
```
// Material id update using particle sets
```

jgiven100: commit https://github.com/cb-geo/mpm/pull/674/commits/150a22ff9d328c9eeb7bce0a808324145b116746

include/solvers/mpm_base.tcc
```diff
@@ -329,6 +329,26 @@ bool mpm::MPMBase<Tdim>::initialise_particles() {
                        particles_volume_end - particles_volume_begin)
                        .count());
 
+    // Material id update using particle sets
+    try {
+      auto material_sets = io_->json_object("material_sets");
```
kks32, line 334: We need to check if the `material_sets` is available. This should be an optional argument

kks32 (changes requested):

include/solvers/mpm_base.tcc
jgiven100: How do we check availability for `material_sets`?

`mpm` successfully runs even if `material_sets` is not provided in input `JSON`.  So I think this already implemented as an optional argument? 

image

include/solvers/mpm_base.tcc
jgiven100: Removed `bool` flag

include/solvers/mpm_base.tcc
kks32: Oh you are using `io_->json_object()` function call, which does that check for you, so you are right, no need to do it again. 

```
  if (json_.find(key) != json_.end()) {
    return json_.at(key);
  } else {
    throw std::runtime_error("No object found, returning an empty object");
    return empty;
  }
```

kks32 (commented):
Have you tried with `material_sets`? does it work with material_sets and the particles are assigned correctly?

jgiven100:
@kks32 Yes, works with material_sets and all particles are correctly updated.

@ezrayst and I quickly changed the `assign_particles` function to include an output that showed us the original and updated `material_id`.

* all `material_id`s were intially asigned `unsigned_max`
* all `material_id`s updated to `material_id` provided in `particles:"generator"`
* based on `pset_id`, some particles updated to new `material_id` provided in `material_sets`

Unfortunately didn't grab any screenshots.

kks32 (approved):

bodhinandach (commented):
Thanks @jgiven100 for your first PR! :tada: Before merging this, could you please add a test?

kks32:
For doing the test we also need `entity_sets.json` file

kks32 (changes requested):
Just to confirm have you tested it on resume to see if it still works?

ezrayst:
@kks32 for `resume` function, don't we get the material from the current `material_id` stored in `.h5` file for each particle? This means we would not pass this function anymore right?

jgiven100:
I tested a simple case using `resume` and stress &displacement outputs matched non-resume outputs

ezrayst (approved):
@jgiven100 maybe if you are interested, you could make test cases for this function (with `entity_sets.json`). But I am good with this PR!

kks32:
@jgiven100 thanks for testing resume. For testing you can add it to `tests/io/write_mesh_particles.cc` JSON object and write to a file.

jgiven100:
@kks32 within `tests/io/write_mesh_particles.cc` are you thinking:
* assign wrong `"material_id": 0` within the initial particle generator
* use `material_sets` to update to correct `"material_id": 1` 
* `ALL TESTS PASS` remains for `mpmtest`

kks32:
@jgiven100 I was thinking writing an entitiy_sets.json file, if we don't already do that and see if we can assign material ids to particle sets.

jgiven100:
the current implementation of particle generator is creating `pset_id: 0` so the procedure I've described is assigning `material_id: 1` based on a particle set without having to include an `entity_sets.json`
```
"material_sets": [
    {
      "material_id": 1,
      "pset_id": 0
    }
  ],
```

kks32:
Could you please fix clangfomat issue: https://app.circleci.com/pipelines/github/cb-geo/mpm/1009/workflows/d64c9659-c6c3-4bd2-b72b-60b008c8772f/jobs/5597

jgiven100:
@kks32 & @ezrayst 
The `material_id` is now updated via `material_sets` based particle set defined in `entity_sets.json` instead of `pset_id` created by particle generator

tests/io/write_mesh_particles.cc
```diff
@@ -110,6 +112,23 @@ bool write_json(unsigned dim, bool resume, const std::string& analysis,
...
+       {{{"id", 2},
+         {"set", {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}}}}}};
+
+  // Dump JSON as an input file to be read
+  std::ofstream file;
+  file.open("entity_sets_0.json");
+  file << json_file.dump(2);
+  file.close();
+
+  return true;
```
ezrayst, line 129: Since this is a `bool`, what is the case when it gives `false`?

tests/io/write_mesh_particles.cc
```diff
@@ -15,7 +15,7 @@ bool write_json(unsigned dim, bool resume, const std::string& analysis,
   auto io_type = "Ascii2D";
   std::string material = "LinearElastic2D";
   std::vector<double> gravity{{0., -9.81}};
-  unsigned material_id = 1;
+  unsigned material_id = 0;
```
ezrayst, line 18: So the plan is to initialize with `material_id` = 0, then update it with the `material_set` to make it 1?

tests/io/write_mesh_particles.cc
jgiven100: It doesn't seem like any of the functions within `write_mesh_particles` ever return `false` so I matched the existing structure

tests/io/write_mesh_particles.cc
jgiven100: Correct. If `material_id: 1` then `all tests pass`.

tests/io/write_mesh_particles.cc
ezrayst: Alright, I am OK with this!

tests/io/write_mesh_particles.cc
```diff
@@ -111,6 +112,23 @@ bool write_json(unsigned dim, bool resume, const std::string& analysis,
...
+bool write_entity_set() {
+  // JSON Entity Sets
+  Json json_file = {
+      {"particle_sets",
+       {{{"id", 2},
+         {"set", {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}}}}}};
+
+  // Dump JSON as an input file to be read
+  std::ofstream file;
+  file.open("entity_sets_0.json");
```
kks32, line 125: Is there already an `entity_sets.json`?

kks32 (changes requested):

tests/io/write_mesh_particles.cc
jgiven100: yes, it consists of 2 node sets. I couldn't find where it was being generated from.

kks32 (approved):

kks32:
@bodhinandach are you happy, could you approve before I merge?

jgiven100:
@kks32 oops I got excited and already merged. Sorry!

Issue #633, opened 2020-05-13
Create two phase solver for saturated soil with explicit method


tianchiTJ:
# Create two phase solver for saturated soil with explicit method

## Summary
Create `ExplicitTwoPhase` solver with fully saturated condition to carry out the hydraulic coupled computation. A one-point two phase method based on volume fractions is used to represent the solid phase and water phase on the same particle representively.

## Motivation
Use it to compute the hydraulic coupled problem. It's the foundation of following implementation of `SemiImplicitTwophase` solver and even `ThreePhase`  solvers. 

## Design Detail
1. Node base design
The computation is mainly based on solving the momentum balance equations of `mixture` and `water phase` on nodes. (There are three different sections to represent the domain, solid phase, water phase and mixture). As solving two of them is enough to obtain the results, we solve the `mixture` first and then the `water phase`. Therefore, in the `node` class, all variables, like `velocity`, are designed and stored as 0 for `mixture` and 1 for `water phase`.

2. Particle class design
Different with the `node` class, all variables in the `particle` class are designed and stored as 0 for `solid` and 1 for `water phase`, which is convenient to represent the effective stress and other variables. A "TwoPhaseParticle" namespace is created to include the functions and variables used.
> 2.1. New variables for two phase particle
Scalar type
  `porosity`, `liquid_material_id`, `liquid_mass`, `liquid_mass_density`, 
  `pore_pressure`, `pore_pressure_constraint`
Vector type
  `liquid_velocity`, `liquid_strain`, `permeability` 
Map type
  `liquid_velocity_constraints`
Other type
  `liquid_material`
> 2.2. Current function need to be modified
`initialise_particle`,  `hdf5`, `compute_mass`
> 2.3. New functions for two phase
`compute_pore_pressure`, `update_permeability`

3. Solver class design
 A `mpm-explicit-twophase` is created in the solver class. The main loop is quite similar with the `mpm-explicit` solver, but the mapping functions are used for each phase separately. `compute_pore_pressure` is an important additional function in the `TwoPhase` solver based on the Darcy law. The position update is based on the solid phase velocity.

3. Based on porous medium theory:
The `volume_fractions_` is defined in the `particle` class based on the porous medium theory and is used to compute the variables in each phase.

4. Fully saturated condition:
The sum of `volume_fractions_` should be equal to 1 and updated in each step.

5. An option is considered for the choice of formulation type (u-p,u-v-p,u-w-p)

## Drawbacks
[...]

PR #675, opened 2020-07-22
Feature/multiphase material

bodhinandach:
**Describe the PR**
This PR added an enhancement to deal with multiphase material initialization from file, which is essential for `two_phase_particle` either in explicit or implicit solver, or later on `three_phase_particle`. The current proposal will work ok with MPI transfer or read from hdf5 for the single-phase `particle` as I am using a default phase argument to be solid. However, hdf5 initialization and MPI transfer process for two-phase is not yet implemented. This should be integrated after we find an approach to deal with multiphase hdf5 output (@kks32 is investigating that).

**Related Issues/PRs**
#633

**Additional context**
I also added a small modification of pressure smoothing. @tianchiTJ @WeijianLIANG Can we store `pore_pressure` as a state variable in the fluid material? By doing so, we can use the same smoothing routine as we use in the single-phase `pressure`.


include/generators/injection.h
```diff
@@ -7,7 +7,7 @@ struct Injection {
   // Particle type
   std::string particle_type;
   // Material id
-  unsigned material_id;
+  std::vector<unsigned> material_id;
```
kks32:
```suggestion
  std::vector<unsigned> material_ids;
```

include/mesh.h
```diff
@@ -195,8 +195,8 @@ class Mesh {
   //! \retval status Create particle status
   bool create_particles(const std::string& particle_type,
                         const std::vector<VectorDim>& coordinates,
-                        unsigned material_id, unsigned pset_id,
-                        bool check_duplicates = true);
+                        const std::vector<unsigned>& material_id,
```
kks32:
```suggestion
                        const std::vector<unsigned>& material_ids,
```

include/mesh.h
```diff
@@ -341,8 +341,8 @@ class Mesh {
   //! \retval point Material point coordinates
   bool generate_material_points(unsigned nquadratures,
                                 const std::string& particle_type,
-                                unsigned material_id, int cset_id,
-                                unsigned pset_id);
+                                const std::vector<unsigned>& material_id,
```
kks32:
```suggestion
                                const std::vector<unsigned>& material_ids,
```

include/mesh.tcc
```diff
@@ -465,10 +465,9 @@ mpm::Index mpm::Mesh<Tdim>::nnodes_rank() {
 
 //! Create cells from node lists
 template <unsigned Tdim>
-bool mpm::Mesh<Tdim>::generate_material_points(unsigned nquadratures,
-                                               const std::string& particle_type,
-                                               unsigned material_id,
-                                               int cset_id, unsigned pset_id) {
+bool mpm::Mesh<Tdim>::generate_material_points(
+    unsigned nquadratures, const std::string& particle_type,
+    const std::vector<unsigned>& material_id, int cset_id, unsigned pset_id) {
```
```suggestion
    const std::vector<unsigned>& material_ids, int cset_id, unsigned pset_id) {
```

include/mesh.tcc
```diff
@@ -477,7 +476,9 @@ bool mpm::Mesh<Tdim>::generate_material_points(unsigned nquadratures,
       unsigned before_generation = this->nparticles();
       bool checks = false;
       // Get material
-      auto material = materials_.at(material_id);
+      std::vector<std::shared_ptr<mpm::Material<Tdim>>> material;
```
```suggestion
      std::vector<std::shared_ptr<mpm::Material<Tdim>>> materials;
```

include/mesh.tcc
```diff
@@ -501,7 +502,8 @@ bool mpm::Mesh<Tdim>::generate_material_points(unsigned nquadratures,
           status = this->add_particle(particle, checks);
           if (status) {
             map_particles_[pid]->assign_cell(*citr);
-            map_particles_[pid]->assign_material(material);
+            for (unsigned phase = 0; phase < material.size(); phase++)
+              map_particles_[pid]->assign_material(material[phase], phase);
```
kks32: how can we check if the particle can take more than 1 material?

include/mesh.tcc
```diff
@@ -477,7 +476,9 @@ bool mpm::Mesh<Tdim>::generate_material_points(unsigned nquadratures,
       unsigned before_generation = this->nparticles();
       bool checks = false;
       // Get material
-      auto material = materials_.at(material_id);
+      std::vector<std::shared_ptr<mpm::Material<Tdim>>> material;
+      for (const auto m_id : material_id)
```
kks32: We can drop the const, it's unsigned

kks32 (changes requested):
I think we should set the materials_ size in the particle constructor and not allow to add more than 1 material, if the particle type doesn't support it.

include/mesh.tcc
bodhinandach: That's a good point, will that be better to add an `unsigned` return function in particle which denotes the `phase_size` it contains? I will then use `map_particles_[pid]->phase_size();` instead of `material.size()`.

kks32: We could resize the materials in the particle constructor and assign material call will pass the material and phase and we use an index-based approach to adding material. This will throw if someone tries to add material to an invalid phase.

bodhinandach: I am not sure what do you mean by an index-based approach of adding material?

bodhinandach: So should I store an internal variable to denote the `phase_size`? There, I can assert if the `phase` passed in `assign_material` is larger or equal to the `phase_size` similar to the one in node.

include/particles/particle.tcc
```diff
@@ -22,7 +24,9 @@ mpm::Particle<Tdim>::Particle(Index id, const VectorDim& coord, bool status)
   this->initialise();
   cell_ = nullptr;
   nodes_.clear();
-  material_ = nullptr;
+  material_.emplace_back(nullptr);
+  material_id_.emplace_back(std::numeric_limits<unsigned>::max());
```
ezrayst: Why do we need this? If we want to loop, we could use `material_` and get the `id` from them right?

include/generators/injection.h
```diff
@@ -7,7 +7,7 @@ struct Injection {
   // Particle type
   std::string particle_type;
   // Material id
-  unsigned material_id;
+  std::vector<unsigned> material_ids;
```
ezrayst, line 10: Any thought of using a vector instead of map?

include/mesh.tcc
```diff
@@ -1644,13 +1648,19 @@ bool mpm::Mesh<Tdim>::generate_particles(const std::shared_ptr<mpm::IO>& io,
       auto particle_type =
           generator["particle_type"].template get<std::string>();
       // Material id
-      unsigned material_id = generator["material_id"].template get<unsigned>();
+      std::vector<unsigned> material_ids;
+      if (generator.at("material_id").is_array())
```
ezrayst, line 1653: If the input json is `[0, 1]` then we know it's an array. But if it is `[0]` would it it counted as array or scalar?

include/particles/particle.tcc
```diff
@@ -108,17 +112,18 @@ bool mpm::Particle<Tdim>::initialise_particle(
     const std::shared_ptr<mpm::Material<Tdim>>& material) {
   bool status = this->initialise_particle(particle);
   if (material != nullptr) {
-    if (this->material_id_ == material->id() ||
-        this->material_id_ == std::numeric_limits<unsigned>::max()) {
+    if (this->material_id() == material->id() ||
```
ezrayst, line 112: As mentioned in the description, this is still using one-phase with `hdf5` for multiphase developed later right?

include/particles/particle.tcc
bodhinandach: Previously the vector is empty. Need to resize them, I am following the previous initialization of `material_id`.

include/generators/injection.h
bodhinandach: Faster to retrieve

include/mesh.tcc
bodhinandach: array

ezrayst: The input file needs to be `0` instead of `[0]` which is not quite intuitive right?

bodhinandach: You can have `0` too following the else clause if `is_array()` is false, this won't break the current `.json` file.

include/generators/injection.h
ezrayst: I guess so with small data like this

include/particles/particle.tcc
ezrayst: The `material_id` could be retrieved easily right and can be obtained from `material_`? Would it be worth it to store a vector for something we have?

include/mesh.tcc
ezrayst: But instead of having an `unsigned` you will have a vector of unsigned, which is not your design right?

include/mesh.tcc
kks32: In the constructor can we do:
```
// for 2 phase
materials_.resize(2);
std::fill(materials_.begin(), materials_.end(), nullptr);
```
and then later in the assign_material we just do
```
materials_.at(phase) = material;
```
this way if we access an invalid phase, c++ `.at()` will throw.

include/mesh.tcc
kks32: Even a single element like `[0]` should behave like an array.

include/particles/particle.tcc
bodhinandach: We call material_id a couple of time in particle, I think saving it may worth if we call it multiple times. Retrieving it from the material class has a lookup cost as well.

include/mesh.tcc
bodhinandach: > But instead of having an `unsigned` you will have a vector of unsigned, which is not your design right?

That's the proposed implementation. User can either use a scalar `0` (the current interface) or an array `[0, 1]` as an input. The latter one is needed for two-phase or three-phase particles.

include/particles/particle.tcc
bodhinandach: Yes. The two-phase version of hdf5 will be implemented later. This PR is just to enhance the input material

include/mesh.tcc
bodhinandach: Instead of putting 2 or 1 in resize, I think it will be better to call it from a function `phase_size()` or a member variable `phase_size_`, there we don't have to write the same line of code again.

include/mesh.tcc
kks32: Thanks @bodhinandach , yes, so keeping scalar ensures backward compatibility

include/mesh.tcc
kks32: That would require an extra variable and a function. While setting materials size in constructor does the same thing.

bodhinandach: Okay, as for now, I think your idea works better. I was initially thinking to reuse `phase_size()` for two-phase two-points coupled but if I need it, I will modify it later.

bodhinandach: @kks32 I decided to write a protected function to resize. This can be used for the derived particle classes too and it is less messy.

tests/mesh_test_2d.cc
```diff
@@ -578,7 +578,7 @@ TEST_CASE("Mesh is checked for 2D case", "[mesh][2D]") {
         // Gauss point generation
         Json jgen;
         jgen["type"] = "gauss";
-        jgen["material_id"] = mid;
+        jgen["material_id"] = {0};
```
kks32, line 581: We can no longer have a scalar value?

include/particles/particle.tcc
ezrayst: I am OK with this then, thanks for the explanation.

tests/mesh_test_2d.cc
bodhinandach: Sure, can! The `mesh_test_3d.cc` test is still in scalar, I just add a variation for testing purposes.

kks32 (approved):

bodhinandach:
@kks32 I added another enhancement in the `material_sets` which has been recently merged so that user can specify the `phase_id` for assigning material using material set. If they don't specify the `phase_id`, `mpm::ParticlePhase::Solid` is used as default.

tianchiTJ (approved):

PR #676, opened 2020-07-22
Improve lock performance with SpinMutex and OpenMP static scheduling

kks32:
**Describe the PR**
Mutex locks take a significant amount of time in OpenMP parallel versions of the code. This update is to reduce the lock by trying a spinlock and reduce the wait times across threads. 

Enable setting the chunk size in OpenMP schedule using
```
export OMP_SCHEDULE="static, 4"
```
This allows for a finer control without having to recompile for different chunk sizes when running different problems. If the variable is not set, the code will still run.

**Additional context**
Spinlocks vs mutexes are a computational bottleneck. However, we cannot use std::atomic, so this is a good workaround the lack of std::atomic support for Eigen and Vector containers. The speed is improved by about 4 - 5 % for the 3D hydrostatic column.

| **OMP Schedule static chunk size** | **Time (s)** | **Speedup** |
|--------------------------------|----------|---------|
| no schedule                    | 742      | 1       |
| 4                              | 725.3    | 1.023   |
| **16**                             | 721.5    | **1.028**   |
| 128                            | 735.2    | 1.009   |
| 512                            | 736.7    | 1.007   |
| 1000                           | 760      | 0.976   |
| **SpinMutex with Chunk-4**         | 711      | **1.044**   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ The CB-Geo MPM code uses a `JSON` file for input configuration. To run the mpm c
 For example:
 
 ```
+export OMP_SCHEDULE="static, 4"
 ./mpm -f /path/to/input-dir/ -i mpm-usf-3d.json -p 8
 ```
 
--- a/include/mutex.h
+++ b/include/mutex.h
@@ -1,0 +1,32 @@
+#ifndef _MUTEX_
+#define _MUTEX_
+
+#include <atomic>
+
+class SpinMutex {
+ private:
+  std::atomic_flag _lock = ATOMIC_FLAG_INIT;
+  std::atomic<std::size_t> _spin_pred{0};
+
+ public:
+  bool try_lock() { return !_lock.test_and_set(std::memory_order_acquire); }
+
+  void lock() {
+    std::size_t spin_count{0};
+    while (!try_lock()) {
+      ++spin_count;
+
+      if (spin_count < _spin_pred * 2) continue;
+
+      // REVISIT (fbrereto) : Iff sleep_for is guaranteed to block even
+      // when the duration is 0, then std::chrono::nanoseconds(0) will
+      // suffice here.
+      std::this_thread::sleep_for(std::chrono::nanoseconds(1));
+    }
+
+    _spin_pred += (spin_count - _spin_pred) / 8;
+  }
+
+  void unlock() { _lock.clear(std::memory_order_release); }
+};
+#endif
--- a/include/mutex.h
+++ b/include/mutex.h
@@ -3,30 +3,35 @@
 
 #include <atomic>
 
+//! Hybrid SpinMutex class
+//! \brief SpinMutex class that offers a fast locking and unlocking mechanism
 class SpinMutex {
- private:
-  std::atomic_flag _lock = ATOMIC_FLAG_INIT;
-  std::atomic<std::size_t> _spin_pred{0};
 
  public:
-  bool try_lock() { return !_lock.test_and_set(std::memory_order_acquire); }
+  //! Attempt locking
+  bool try_lock() { return !lock_.test_and_set(std::memory_order_acquire); }
 
+  //! Call lock 
   void lock() {
     std::size_t spin_count{0};
     while (!try_lock()) {
       ++spin_count;
 
-      if (spin_count < _spin_pred * 2) continue;
+      if (spin_count < spin_pred_ * 2) continue;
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -178,6 +178,7 @@ The CB-Geo MPM code uses a `JSON` file for input configuration. To run the mpm c
 For example:
 
 ```
+export OMP_SCHEDULE="static, 4"
```
bodhinandach, line 181: @kks32 I got a short question: here, the chunk_size is set to be 4, whereas the `-p 8` option is used. In this case, how many threads will be used to run the `parallel for`, with and without the `schedule(runtime)`?

kks32: chunk size and number of threads are different. http://jakascorner.com/blog/2016/06/omp-for-scheduling.html `-p 8` sets the number of threads, not the chunk size.

ezrayst: I think it would still be 8 threads as there will be block cycling scheduling or like "round robin".

include/mutex.h
```diff
@@ -0,0 +1,37 @@
...
+    std::size_t spin_count{0};
+    while (!try_lock()) {
+      ++spin_count;
+
+      if (spin_count < spin_pred_ * 2) continue;
+
+      std::this_thread::sleep_for(std::chrono::nanoseconds(1));
+    }
+
+    spin_pred_ += (spin_count - spin_pred_) / 8;
```
ezrayst, line 26: I guess this is a predictor from measured value from the literature? I am just not sure where number 8 is from.

include/mutex.h
```diff
@@ -0,0 +1,37 @@
...
+class SpinMutex {
+
+ public:
+  //! Attempt locking
+  bool try_lock() { return !lock_.test_and_set(std::memory_order_acquire); }
+
+  //! Call lock
+  void lock() {
+    std::size_t spin_count{0};
+    while (!try_lock()) {
```
ezrayst, line 18: I wonder the double negative here, first in line 12 which returns `!` and then here. Is there any reason to do so?

include/mutex.h
```diff
@@ -0,0 +1,37 @@
+#ifndef _MUTEX_
+#define _MUTEX_
+
+#include <atomic>
+
+//! Hybrid SpinMutex class
+//! \brief SpinMutex class that offers a fast locking and unlocking mechanism
+class SpinMutex {
```
ezrayst, line 9: Do you want to call it Hybrid or Adaptive Mutex instead? I think it's clearer, but I am also OK with SpinMutex

ezrayst (commented):
Thank for this @kks32! I leave my comments but I don't seem to understand your table. What are the number represent (I thought it's how you divide your chunks but apparently not because of the last row)? 

So if you divide into chunks of 4, you get 4-5% speedup, but did you try with 2 or 8? Any thoughts on using "dynamic" or "auto"?

include/mutex.h
kks32: It's a spin lock and a Mutex, hence SpinMutex

kks32:
> Thank for this @kks32! I leave my comments but I don't seem to understand your table. What are the number represent (I thought it's how you divide your chunks but apparently not because of the last row)?
> 
> So if you divide into chunks of 4, you get 4-5% speedup, but did you try with 2 or 8? Any thoughts on using "dynamic" or "auto"?

There are two things on the table. 1. OpenMP performance with different chunk sizes and 2. SpinMutex implementation with a certain number of chunk size. If you let OpenMP decide, that's our baseline (742s). Dynamic will be slower. The best number of chunk-sizes and threads must be decided by the user running to code. These are only for the 3D hydrostatic column.

include/mutex.h
kks32: Because of the operation it is trying to do, we are trying to lock with the bool call. However, when we call the `try_lock` we are seeing if it can lock or not. We can't rename it as `try_unlock` because it won't be representative of what it means.

bodhinandach (approved):
I am ok with this. Thanks @kks32. Do you know though why the CI doesn't let this pass?

kks32:
https://app.circleci.com/pipelines/github/kks32/mpm?branch=performance%2Flock CI passed.

include/mutex.h
ezrayst: Understood

include/mutex.h
ezrayst: Understood, thanks for the explanation

ezrayst (approved):
I am OK too. Thanks @kks32

PR #677, opened 2020-07-24
Performance/openmp schedule mutex

kks32:
**Describe the PR**
Duplicate of #676 
Mutex locks take a significant amount of time in OpenMP parallel versions of the code. This update is to reduce the lock by trying a spinlock and reduce the wait times across threads. 

Enable setting the chunk size in OpenMP schedule using
```
export OMP_SCHEDULE="static, 4"
```
This allows for a finer control without having to recompile for different chunk sizes when running different problems. If the variable is not set, the code will still run.

**Additional context**
Spinlocks vs mutexes are a computational bottleneck. However, we cannot use std::atomic, so this is a good workaround the lack of std::atomic support for Eigen and Vector containers. The speed is improved by about 4 - 5 % for the 3D hydrostatic column.

| **OMP Schedule static chunk size** | **Time (s)** | **Speedup** |
|--------------------------------|----------|---------|
| no schedule                    | 742      | 1       |
| 4                              | 725.3    | 1.023   |
| **16**                             | 721.5    | **1.028**   |
| 128                            | 735.2    | 1.009   |
| 512                            | 736.7    | 1.007   |
| 1000                           | 760      | 0.976   |
| **SpinMutex with Chunk-4**         | 711      | **1.044**   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ The CB-Geo MPM code uses a `JSON` file for input configuration. To run the mpm c
 For example:
 
 ```
+export OMP_SCHEDULE="static, 4"
 ./mpm -f /path/to/input-dir/ -i mpm-usf-3d.json -p 8
 ```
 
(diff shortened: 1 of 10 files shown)


kks32:
This is just a duplicate of #676  as I couldn't get the CI to work

bodhinandach (approved):
Thanks @kks32!

ezrayst:
Alright, thanks @kks32! LGTM!

ezrayst (approved):

PR #681, opened 2020-07-29
[Solver] Semi-implicit Navier-Stokes solver

### LINKED ISSUE #634: Semi-Implicit Navier-Stokes solver for incompressible flow
## Summary

This RFC is to propose the Navier-Stokes solver in branch `solver/navier-stokes`. The NS solver utilized semi-implicit Chorin's projection scheme which is mainly to be used for modeling incompressible fluids.

## Motivation

To extend the capability of treating internal incompressibility constraint more accurately particularly for simulations of fluid flows. The proposed method has been commonly studied in CFD community and proven to be one of the simplest yet efficient.

## Design Detail

The following classes and functionality were implemented:
1. `MPMSemiImplicitNavierStokes` solver class (this at the moment only works with point 2.)
2. A `FluidParticle` derived from `Particle` with some specialized class derivation. (Will be refactor following #637, #654, #673)
3. Linear solvers and assembler classes based on Eigen.

a. The assembler class is made to assemble the global system of equations. First, all nodes will be assigned an `active_id` to specify the index of each node in the global systems of equations; this id is changing every time step as the particles move over the background grids. Then, when the explicit step is completed, a laplacian and RHS matrix are constructed in all cells. We store this local matrix with size of (nnodes x nnodes) as Eigen matrices in `Cell` along with the correction_matrix as:
```
  //! Local laplacian matrix
  Eigen::MatrixXd laplacian_matrix_;
  //! Local poisson RHS matrix
  Eigen::MatrixXd poisson_right_matrix_;
  //! Local correction RHS matrix
  Eigen::MatrixXd correction_matrix_;
```
These local matrices are by default has zero sizes, and only initialized when the `initialise_element_matrix` function is called by the solver. Each time set, the assembler job is to organize all the cell's local stiffness matrix to a global matrix and RHS vector and to apply constraints to the assembled systems through row-column modification.

b. The linear solver solves the given assembled coefficient matrix and RHS vector, and it is independent of dimension. The `solver_type` should be specified in the input .json.
```
virtual Eigen::VectorXd solve(const Eigen::SparseMatrix<double>& A,
                                const Eigen::VectorXd& b,
                                std::string solver_type) = 0;
```
c. Parallel solver capability is also added to improve the solver scalability and efficiency as previously noted in RFC #635.

4. Free-surface detection algorithms (both cheap and expensive detection were implemented) in `Mesh` class.
5. Some boundary conditions were implemented, e.g. pressure BC.
6. Some work on level-set signed distance computation if accurate free-surface detection method is considered.
7. Few testings have been added and few more will be added.

## Drawbacks

No drawbacks in performance at the moment. It will enrich the capability of the software in many ways. However, the implementation might cause the function derivatives to be a bit messy, particularly in `Cell` class where we need to store and construct some local element matrices. Further thoughts and suggestions are welcome.

## Rationale and Alternatives

After the merge of the NS solver, we can also extend the solver to be used for any incompressible Solid with minimum modification. Some further thoughts on including purely Implicit solver class is also possible for future development. They utilize very much the same structure and the capability can be extended from the proposed work. 

## Prior Art
Referring a lot to the work of @srhgk2:
1. Kularathna, S., & Soga, K. (2017). Implicit formulation of material point method for analysis of incompressible materials. Computer Methods in Applied Mechanics and Engineering, 313, 673-686.

## Changelog
1. Add details of linear solver and assembler class structure.
2. Add capability of parallel solver and other MPI processes in fluid particle.

bodhinandach:
**Describe the PR**
A draft PR to ease to check the solver structure.


include/cell.h
```diff
@@ -214,6 +214,66 @@ class Cell {
   //! Return rank
   unsigned rank() const;
 
+  //! Assign free surface
+  //! \param[in] free_surface boolean indicating free surface cell
+  void assign_free_surface(bool free_surface) { free_surface_ = free_surface; };
+
+  //! Return free surface bool
+  //! \retval free_surface_ indicating free surface cell
+  bool free_surface() { return free_surface_; };
```
kks32, line 223: Could you make all unaltered functions as const and add inline when appropriate?

include/cell.tcc
```diff
@@ -844,3 +844,107 @@ template <unsigned Tdim>
 inline unsigned mpm::Cell<Tdim>::previous_mpirank() const {
   return this->previous_mpirank_;
 }
+
+//! Map cell volume to nodes
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::map_cell_volume_to_nodes(unsigned phase) {
+  bool status = true;
+  try {
```
kks32, line 852: We don't need to use exception handling here, instead, use assert. A failure here will give incorrect results, so this should fail.

include/cell.tcc
```diff
@@ -844,3 +844,107 @@ template <unsigned Tdim>
...
+
+//! Return local node indices
+template <unsigned Tdim>
+Eigen::VectorXi mpm::Cell<Tdim>::local_node_indices() {
+  Eigen::VectorXi indices;
+  try {
+    indices.resize(nodes_.size());
+    indices.setZero();
+    unsigned node_idx = 0;
+    for (auto node_itr = nodes_.cbegin(); node_itr != nodes_.cend();
```
kks32, line 876: Would this be the same for both standard linear and GIMP elements?

include/cell.tcc
```diff
@@ -844,3 +844,107 @@ template <unsigned Tdim>
...
+  return status;
+}
+
+//! Compute local matrix of laplacian
+template <unsigned Tdim>
+void mpm::Cell<Tdim>::compute_local_laplacian(
+    const Eigen::MatrixXd& grad_shapefn, double pvolume,
+    double multiplier) noexcept {
+
+  std::lock_guard<std::mutex> guard(cell_mutex_);
```
kks32, line 919: Use mutex lock and unlock

include/cell.tcc
```diff
@@ -844,3 +844,107 @@ template <unsigned Tdim>
...
+  }
+  return indices;
+}
+
+//! Initialise element matrix
+template <unsigned Tdim>
+bool mpm::Cell<Tdim>::initialise_element_matrix() {
+  bool status = true;
+  if (this->status()) {
+    try {
```
kks32, line 892: We don't need an exception block

kks32 (changes requested):
Remove exception handling in core computational functions, if those block enter invalid regions, the code should crash.

[Partial review]

PR #682, opened 2020-08-03
[Solver] Explicit Modified Update Stress Last solver

bodhinandach:
**Describe the PR**
This PR quickly adds the explicit MUSL solver capability in our `mpm_explicit` solver. I thought it will be useful as the `mpm::StressUpdate::MUSL` is available, though never be used. Due to its considerably slower performance, if you think the MUSL option is not necessary, let's remove the MUSL option.

**Additional context**
The MUSL option is basically similar to USL stress update, though require one more time momentum update right before recomputing strain_rate and strain for stress update. I ran the benchmark tests in both serial and parallel and both are ok! The necessity to do `map_momentum_to_nodes` and the `compute_velocity` one more time add a bit more computation time, so I am not sure if it is so useful. Here is the time comparison for the 3D hydrostatic column test in our benchmark test run in my local machine.

| Stress Update                   | Runtime (s) |
|--------------------------|-------------------|
| USF                  | 512.5            |
| USL | 505.4             |
| MUSL       | 560.2             |

### DIFF
--- a/include/solvers/mpm_explicit.tcc
+++ b/include/solvers/mpm_explicit.tcc
@@ -288,11 +288,8 @@ bool mpm::MPMExplicit<Tdim>::solve() {
     // Apply particle velocity constraints
     mesh_->apply_particle_velocity_constraints();
 
-    // Update Stress Last
-    if (this->stress_update_ == mpm::StressUpdate::USL)
-      this->compute_stress_strain(phase);
     // Modified Update Stress Last
-    else if (this->stress_update_ == mpm::StressUpdate::MUSL) {
+    if (this->stress_update_ == mpm::StressUpdate::MUSL) {
 
 #pragma omp parallel sections
       {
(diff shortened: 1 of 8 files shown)


kks32 (changes requested):
Could we have separate functions inside MPMExplicit for USF, USL, MUSL. There will be some repetition, at least it would be clearer and we'll have fewer branching conditions

bodhinandach:
@kks32 I am okay. How you want to design or categorize them? I was thinking to call `initialise` node, `compute_mass_momentum`, and `compute_velocity` so that it was a set as the one done initially. However, the computational time becomes even slower as you do more mappings. Please feel free to commit your change @kks32.

include/solvers/mpm_explicit.tcc
```diff
@@ -288,9 +288,54 @@ bool mpm::MPMExplicit<Tdim>::solve() {
     // Apply particle velocity constraints
     mesh_->apply_particle_velocity_constraints();
 
+    // Modified Update Stress Last
+    if (this->stress_update_ == mpm::StressUpdate::MUSL) {
+
+#pragma omp parallel sections
```
kks32, line 294: Why do we need parallel sections here?

kks32 (changes requested):
Update to the newer format to include MUSL #685

bodhinandach:
I am closing this PR and will reconsider the framework after looking at #685

PR #680, opened 2020-07-29
[Solver] Two-Phase One-Point Explicit

tianchiTJ:
**Describe the PR**
This PR is to implement the two-phase one-layer explicit solver for fully saturated soil simulation. One particle includes both solid phase and liquid phase, which are defined by the corresponding volume fractions (`n_s` and `n_w`). The theory of porous media is adopted to set up the governing equations (momentum balance and mass balance). The pore pressure is updated based on the strain of the solid skeleton, and a weakly compressible fluid assumption is assumed to solve the fluid and mixture balance equations. All implementations work coherently with the current MPI and OpenMP parallelization features.

**New implementation**
***TwoPhase Particle***
1. New class `ParticleTwoPhase` derived from `Particle`, which include all necessary functions for two-phase computation and new variables for two-phase (etc. pore pressure, porosity)
2. New hdf5 structure for `ParticleTwoPhase` (also includes refactoring of base hdf5 class)

***MPMExplicitTwoPhase***
1. New class `MPMExplicitTwoPhase` derived from `MPMBase`, which includes the main loop for two-phase computation.
2. New input functions in `MPMBase` to read the new types of input files for two-phase.

***TwoPhase functions for nodes***
1. New functions for the two-phase are implemented in the current `Node` class directly (see `node_multiphase.tcc`)

***Free surface detection functions in mesh***
1. New functions to detect the changing surface are implemented in the `Mesh` class (see `mesh_multiphase.tcc`)

***Unit tests and benchmarks***
All new implementations are validated by corresponding test functions, and the unit-test for two-phase computation is also included. Related 2D and 3D one-dimensional consolidation tests can be found here: https://github.com/cb-geo/mpm-benchmarks/pull/19.

PR #683, opened 2020-08-05
[Feature] State variable vtk output for different phases

bodhinandach:
**Describe the PR**
This PR enhances the capability to output vtk files for two-phase particles following the material container refactoring done in PR #675 and #678.

**Related Issues/PRs**
PR #675 and #678 as well as PR #680 and Issue #633.

**Additional context**
This will break the existing input configuration. User should change the current `vtk_statevars` arrangement:
```
"vtk_statevars": ["pressure", "pdstrain"]
```
to the new one with `phase_id`:
```
"vtk_statevars": [
	{
		"phase_id" : 0,
		"statevars": ["pressure", "pdstrain"]
	},
	{
		"phase_id" : 1,
		"statevars": ["pressure"]
	}
]
```
If the user doesn't specify the `phase_id`, it is assumed that the `phase_id` is 0 by default.
```
"vtk_statevars": [
	{
		"statevars": ["pressure", "pdstrain"]
	}
]
```
There is a slight change also in the outputted file name. Instead of producing `pressureXYZ.vtp` or `pressureXYZ.pvtp` as output, the modified code will produce `P0pressureXYZ.vtp` and `P0pressureXYZ.pvtp` or `P1pressureXYZ.vtp` and `P1pressureXYZ.pvtp` for the solid and liquid phase, respectively. I am okay with any other idea of naming the output file, though we need to make it different, as otherwise, the second phase will overwrite the first one.

### DIFF
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -266,6 +266,7 @@ class Mesh {
 
   //! Return particles state variable data
   //! \param[in] attribute Name of the state variable attribute
+  //! \param[in] phase Index corresponding to the phase
   std::vector<double> particles_statevars_data(
       const std::string& attribute, unsigned phase = mpm::ParticlePhase::Solid);
 
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -264,8 +264,8 @@ class Mesh {
   std::vector<Eigen::Matrix<double, Tsize, 1>> particles_tensor_data(
       const std::string& attribute);
 
-  //! Return particles scalar data
-  //! \param[in] attribute Name of the scalar data attribute
+  //! Return particles state variable data
+  //! \param[in] attribute Name of the state variable attribute
   std::vector<double> particles_statevars_data(
       const std::string& attribute, unsigned phase = mpm::ParticlePhase::Solid);
 
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -266,7 +266,8 @@ class Mesh {
 
   //! Return particles scalar data
   //! \param[in] attribute Name of the scalar data attribute
-  std::vector<double> particles_statevars_data(const std::string& attribute);
+  std::vector<double> particles_statevars_data(
+      const std::string& attribute, unsigned phase = mpm::ParticlePhase::Solid);
 
   //! Compute and assign rotation matrix to nodes
   //! \param[in] euler_angles Map of node number and respective euler_angles
--- a/include/solvers/mpm_base.h
+++ b/include/solvers/mpm_base.h
@@ -193,7 +193,7 @@ class MPMBase : public MPM {
   //! Mathematical functions
   std::map<unsigned, std::shared_ptr<mpm::FunctionBase>> math_functions_;
   //! VTK state variables
-  std::vector<std::string> vtk_statevars_;
+  std::map<unsigned, std::vector<std::string>> vtk_statevars_;
   //! Set node concentrated force
   bool set_node_concentrated_force_{false};
   //! Damping type
--- a/include/solvers/mpm_base.h
+++ b/include/solvers/mpm_base.h
@@ -193,7 +193,7 @@ class MPMBase : public MPM {
   //! Mathematical functions
   std::map<unsigned, std::shared_ptr<mpm::FunctionBase>> math_functions_;
   //! VTK state variables
-  std::map<unsigned, std::vector<std::string>> vtk_statevars_;
+  tsl::robin_map<unsigned, std::vector<std::string>> vtk_statevars_;
   //! Set node concentrated force
   bool set_node_concentrated_force_{false};
   //! Damping type
--- a/include/solvers/mpm_base.tcc
+++ b/include/solvers/mpm_base.tcc
@@ -98,42 +98,27 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
         post_process_.at("vtk_statevars").size() > 0) {
       // Iterate over state_vars
       for (const auto& svars : post_process_["vtk_statevars"]) {
-        // Initial phase id and state vars
+        // Phase id
         unsigned phase_id = 0;
+        if (svars.contains("phase_id"))
+          phase_id = svars.at("phase_id").template get<unsigned>();
+
+        // State variables
         std::vector<std::string> state_var;
-
-        // If state variables are arranged as string
-        if (svars.is_string()) {
-          std::string attribute = svars.template get<std::string>();
-          state_var.emplace_back(attribute);
-        }
-        // If state variables are arranged as json object
-        else if (svars.is_object()) {
-          // Phase id
-          if (svars.contains("phase_id"))
-            phase_id = svars.at("phase_id").template get<unsigned>();
-
-          // State variables
-          if (svars.at("statevars").is_array() &&
-              svars.at("statevars").size() > 0) {
-            for (unsigned i = 0; i < svars.at("statevars").size(); ++i) {
-              std::string attribute =
-                  svars["statevars"][i].template get<std::string>();
-              state_var.emplace_back(attribute);
-            }
-          } else {
-            throw std::runtime_error(
-                "No VTK statevariable were specified, none will be generated");
+        if (svars.at("statevars").is_array() &&
+            svars.at("statevars").size() > 0) {
+          for (unsigned i = 0; i < svars.at("statevars").size(); ++i) {
+            std::string attribute =
+                svars["statevars"][i].template get<std::string>();
+            state_var.emplace_back(attribute);
           }
+        } else {
+          throw std::runtime_error(
+              "No VTK statevariable were specified, none will be generated");
         }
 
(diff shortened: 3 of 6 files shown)


include/solvers/mpm_base.tcc
```diff
@@ -96,10 +96,44 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+        if (svars.is_string()) {
+          std::string attribute = svars.template get<std::string>();
+          state_var.emplace_back(attribute);
+        }
+        // If state variables are arranged as json object
+        else if (svars.is_object()) {
+          // Phase id
+          if (svars.contains("phase_id"))
+            phase_id = svars.at("phase_id").template get<unsigned>();
+
```
tianchiTJ: Should we add a check for the valid input of phase_id?

tianchiTJ (approved):

include/solvers/mpm_base.tcc
bodhinandach: Good point. I think this part is defined globally for simulating single-phase and two-phase, so checking `phase_id` might be a bit difficult as we need to go to particle level initially unless we specify it somewhere. Nevertheless, if you look at these lines:
https://github.com/cb-geo/mpm/blob/3b20da5fcace5ed390c8ff58b08f0040fa27d51e/include/particles/particle.h#L241-L247
Maybe we can use `state_variables_.at(phase)` instead of `state_variables_[phase]` for all the above to assert if the given phase is not allocated. Just that it will be a slower call. What do you think @kks32?

include/solvers/mpm_base.tcc
```diff
@@ -554,23 +588,29 @@ void mpm::MPMBase<Tdim>::write_vtk(mpm::Index step, mpm::Index max_steps) {
...
-    // Write state variables
-    auto file =
-        io_->output_file(attribute, extension, uuid_, step, max_steps).string();
-    vtk_writer->write_scalar_point_data(
-        file, mesh_->particles_statevars_data(attribute), attribute);
-    // Write a parallel MPI VTK container file
+  for (auto const& vtk_statevar : vtk_statevars_) {
+    unsigned phase_id = vtk_statevar.first;
+    for (const auto& attribute : vtk_statevar.second) {
+      std::string phase_attribute = "P" + std::to_string(phase_id) + attribute;
```
jgiven100: Maybe just use `"phase" + phase_id` instead of `"P" + phase_id`. I feel there are many `"P"s` within the code (e.g. particles, pressure, phase) so why not be specific? 
```suggestion
      std::string phase_attribute = "phase" + std::to_string(phase_id) + attribute;
```

include/solvers/mpm_base.h
```diff
@@ -193,7 +193,7 @@ class MPMBase : public MPM {
   //! Mathematical functions
   std::map<unsigned, std::shared_ptr<mpm::FunctionBase>> math_functions_;
   //! VTK state variables
-  std::vector<std::string> vtk_statevars_;
+  std::map<unsigned, std::vector<std::string>> vtk_statevars_;
```
kks32: I'd rather have `tsl::robinmap` as it is amortized

include/solvers/mpm_base.tcc
```diff
@@ -96,10 +96,44 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
-        std::string attribute =
-            post_process_["vtk_statevars"][i].template get<std::string>();
-        vtk_statevars_.emplace_back(attribute);
+      // Iterate over state_vars
+      for (const auto& svars : post_process_["vtk_statevars"]) {
+        // Initial phase id and state vars
+        unsigned phase_id = 0;
+        std::vector<std::string> state_var;
+
+        // If state variables are arranged as string
```
kks32: Why not just set input JSON to a unified format, where they have to specify phase_id and variables?

include/solvers/mpm_base.tcc
```diff
@@ -96,10 +96,44 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+                  svars["statevars"][i].template get<std::string>();
+              state_var.emplace_back(attribute);
+            }
+          } else {
+            throw std::runtime_error(
+                "No VTK statevariable were specified, none will be generated");
+          }
+        }
+
+        // Check if vtk_statevars_ has the designated phase_id
```
kks32: I don't follow what's happening here. Why can't we check before we add to `state_var`?

kks32 (changes requested):

kks32:
We can use the updated alternative, no reason for us to be backward compatible in this case. 

```
"vtk_statevars": [
	{
		"phase_id" : 0,
		"statevars": ["pressure", "pdstrain"]
	},
	{
		"phase_id" : 1,
		"statevars": ["pressure"]
	}
]
```

bodhinandach:
It means then we need to modify all mpm.json, are you ok with that? Can we at least make it deprecated for some time and output a deprecated message to let the user change their format?

kks32:
Yes, otherwise it's too much of unnecessary branching conditions. No, we create a breaking change, no need for deprecated message or sometime to change. If they use the old format, the code will throw an error, that should be sufficient.

bodhinandach:
@kks32 I added the breaking change. Ping @cb-geo/mpm for info. I will make a new pull request to the benchmark repo to change the input json there as soon as this is merged.

include/solvers/mpm_base.tcc
bodhinandach: Any further thoughts on this?

kks32: Doesn't this guarantee that the variable is available: `state_variables_[phase].find(var) != state_variables_[phase].end()`

include/solvers/mpm_base.tcc
```diff
@@ -96,10 +96,29 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+        std::vector<std::string> state_var;
+        if (svars.at("statevars").is_array() &&
+            svars.at("statevars").size() > 0) {
+          for (unsigned i = 0; i < svars.at("statevars").size(); ++i) {
+            std::string attribute =
+                svars["statevars"][i].template get<std::string>();
+            state_var.emplace_back(attribute);
+          }
+        } else {
+          throw std::runtime_error(
```
kks32: Change to `console_->warn`

include/solvers/mpm_base.tcc
```diff
@@ -96,10 +96,29 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+      // Iterate over state_vars
+      for (const auto& svars : post_process_["vtk_statevars"]) {
+        // Phase id
+        unsigned phase_id = 0;
+        if (svars.contains("phase_id"))
+          phase_id = svars.at("phase_id").template get<unsigned>();
+
+        // State variables
+        std::vector<std::string> state_var;
+        if (svars.at("statevars").is_array() &&
```
kks32: `std::vector<std::string> state_var = svars["statevars"];`

kks32: We don't have to iterate

kks32 (changes requested):
We never check if the state variable is valid, do we?

include/solvers/mpm_base.tcc
bodhinandach: Yes, if the `state_variables_` have the appropriate `phase`. If the passed `phase` is equal or larger than `state_variables_.size()`, then it will cause a segmentation fault. For instance, if I am accessing index `phase=1` in single-phase particle, it will be seg-fault

include/solvers/mpm_base.tcc
bodhinandach: Fixed

include/solvers/mpm_base.tcc
bodhinandach: Fixed

include/mesh.h
```diff
@@ -264,9 +264,10 @@ class Mesh {
   std::vector<Eigen::Matrix<double, Tsize, 1>> particles_tensor_data(
       const std::string& attribute);
 
-  //! Return particles scalar data
-  //! \param[in] attribute Name of the scalar data attribute
-  std::vector<double> particles_statevars_data(const std::string& attribute);
+  //! Return particles state variable data
+  //! \param[in] attribute Name of the state variable attribute
```
kks32, line 268: We need \param[in] phase

tianchiTJ (approved):

kks32 (approved):

include/solvers/mpm_base.tcc
bodhinandach: @kks32 Are we good with this? The current implementation has no `phase_id` check. So if input `phase_id` is not available, we will have seg fault.

PR #684, opened 2020-08-05
[Feature] VTK output vars

### LINKED ISSUE #594: Output variable libraries
**Describe the feature**
Generalization of particle output variables other than `state_vars`

**Additional context**
I am thinking we should have a "vtk" input parameter back in the JSON file to output necessary variables. It doesn’t work at the moment after we change to `particle_tensor_data`, i.e. PR #583. Also, the only way to write an output implementation is to write it inside `state_variables`. This mess up the initial definition of `state_variables`, so I don’t think it is good. We can hardcoded it to `mpm_base`, but that's not the best way to do it again and again.

Instead, we should create a list (or library) of output variables, which can be simply a string, or some other advance class type if necessary, and should be associated with the type of the same variable. This can be scalar (including bool, int, unsigned, and double), vector, or tensor (2nd order tensor). This works just like a register.

There, when we have the list of output variables read from JSON, we just need to first validate, with respect to the default parameter, i.e. displacements, velocities. Stresses and Strains can be set as non-default as the automatic print now slows down the output as it results in 9 component prints, instead of 3 previously. After validation to remove repetition, we pass the validated string list into the library we defined to get the data type associated with each output variable. 

The next process is straight forward, i.e. to pass this to the `property_data` in particle, which I think should be a function with a template of the data type, or at least an unsigned of size n={1, 3, or 9}. We can even generalize this by returning Eigen vector data type of size n. There, we can implement which function associated to return the output variables at the particle level (the same way or similar to the current code).

The current output routines can be used to produce the VTK file, or at least with minimum modification.

bodhinandach:
**Describe the PR**
This PR solves issue #594 to output variables from JSON with unknown type. A new map to link the string specified in JSON and a new `VariableType` class is created. Also, this removes the default output variables of displacements, velocities, stresses, and strains.

Every time a user wants to output a variable, just add the string and type to the list specified in `src/mpm.cc` and at the `initialise() `function in `particle.tcc` for which variable to return using lambda function.

The JSON list "vtk" is ready back to be used.

**Related Issues/PRs**
Issue #594

**Additional context**
1. The current design work only for vector and tensor variables in particle, the scalar part is still ongoing development. (**Done** :heavy_check_mark:)
2. @kks32 Maybe we might want to use the flat map that you test to have a fast retrieval. (**Done using `tsl_map`** :heavy_check_mark:)

### DIFF
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -258,6 +258,10 @@ class Mesh {
   //! Return coordinates of particles
   std::vector<Eigen::Matrix<double, 3, 1>> particle_coordinates();
 
+  //! Return particles scalar data
+  //! \param[in] attribute Name of the scalar data attribute
+  std::vector<double> particles_scalar_data(const std::string& attribute);
+
   //! Return particles tensor data
   //! \param[in] attribute Name of the tensor data attribute
   template <unsigned Tsize>
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -261,20 +261,20 @@ class Mesh {
   //! Return particles scalar data
   //! \param[in] attribute Name of the scalar data attribute
   //! \retval scalar_data Vector containing scalar properties from particles
-  std::vector<double> particles_scalar_data(const std::string& attribute);
+  std::vector<double> particles_scalar_data(const std::string& attribute) const;
 
   //! Return particles vector data
   //! \param[in] attribute Name of the tensor data attribute
   //! \retval vector_data Vector containing vector properties from particles
   std::vector<Eigen::Matrix<double, 3, 1>> particles_vector_data(
-      const std::string& attribute);
+      const std::string& attribute) const;
 
   //! Return particles tensor data
   //! \param[in] attribute Name of the tensor data attribute
   //! \retval tensor_data Vector containing tensor properties from particles
   template <unsigned Tsize>
   std::vector<Eigen::Matrix<double, Tsize, 1>> particles_tensor_data(
-      const std::string& attribute);
+      const std::string& attribute) const;
 
   //! Return particles state variable data
   //! \param[in] attribute Name of the state variable attribute
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -268,9 +268,11 @@ class Mesh {
   std::vector<Eigen::Matrix<double, Tsize, 1>> particles_tensor_data(
       const std::string& attribute);
 
-  //! Return particles scalar data
-  //! \param[in] attribute Name of the scalar data attribute
-  std::vector<double> particles_statevars_data(const std::string& attribute);
+  //! Return particles state variable data
+  //! \param[in] attribute Name of the state variable attribute
+  //! \param[in] phase Index corresponding to the phase
+  std::vector<double> particles_statevars_data(
+      const std::string& attribute, unsigned phase = mpm::ParticlePhase::Solid);
 
   //! Compute and assign rotation matrix to nodes
   //! \param[in] euler_angles Map of node number and respective euler_angles
--- a/include/mesh.h
+++ b/include/mesh.h
@@ -260,10 +260,18 @@ class Mesh {
 
   //! Return particles scalar data
   //! \param[in] attribute Name of the scalar data attribute
+  //! \retval scalar_data Vector containing scalar properties from particles
   std::vector<double> particles_scalar_data(const std::string& attribute);
+
+  //! Return particles vector data
+  //! \param[in] attribute Name of the tensor data attribute
+  //! \retval vector_data Vector containing vector properties from particles
+  std::vector<Eigen::Matrix<double, 3, 1>> particles_vector_data(
+      const std::string& attribute);
 
   //! Return particles tensor data
   //! \param[in] attribute Name of the tensor data attribute
+  //! \retval tensor_data Vector containing tensor properties from particles
   template <unsigned Tsize>
   std::vector<Eigen::Matrix<double, Tsize, 1>> particles_tensor_data(
       const std::string& attribute);
@@ -271,6 +279,7 @@ class Mesh {
   //! Return particles state variable data
   //! \param[in] attribute Name of the state variable attribute
   //! \param[in] phase Index corresponding to the phase
+  //! \retval statevars_data Vector containing state variable from particles
   std::vector<double> particles_statevars_data(
       const std::string& attribute, unsigned phase = mpm::ParticlePhase::Solid);
 
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1078,7 +1078,7 @@ std::vector<Eigen::Matrix<double, 3, 1>>
 //! Return particle scalar data
 template <unsigned Tdim>
 std::vector<double> mpm::Mesh<Tdim>::particles_scalar_data(
-    const std::string& attribute) {
+    const std::string& attribute) const {
   std::vector<double> scalar_data;
   scalar_data.reserve(particles_.size());
   // Iterate over particles and add scalar value to data
@@ -1090,7 +1090,7 @@ std::vector<double> mpm::Mesh<Tdim>::particles_scalar_data(
 //! Return particle vector data
 template <unsigned Tdim>
 std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::particles_vector_data(
-    const std::string& attribute) {
+    const std::string& attribute) const {
   std::vector<Eigen::Matrix<double, 3, 1>> vector_data;
   // Iterate over particles
   for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr) {
@@ -1110,7 +1110,7 @@ std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::particles_vector_data(
 template <unsigned Tdim>
 template <unsigned Tsize>
 std::vector<Eigen::Matrix<double, Tsize, 1>>
-    mpm::Mesh<Tdim>::particles_tensor_data(const std::string& attribute) {
+    mpm::Mesh<Tdim>::particles_tensor_data(const std::string& attribute) const {
   std::vector<Eigen::Matrix<double, Tsize, 1>> tensor_data;
   // Iterate over particles
   for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr) {
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1087,6 +1087,31 @@ std::vector<double> mpm::Mesh<Tdim>::particles_scalar_data(
   return scalar_data;
 }
 
+//! Return particle vector data
+template <unsigned Tdim>
+std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::particles_vector_data(
+    const std::string& attribute) {
+  std::vector<Eigen::Matrix<double, 3, 1>> vector_data;
+  try {
+    // Iterate over particles
+    for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr) {
+      Eigen::Matrix<double, 3, 1> data;
+      data.setZero();
+      auto pdata = (*pitr)->vector_data(attribute);
+      // Fill vector_data to the size of dimensions
+      for (unsigned i = 0; i < pdata.size(); ++i) data(i) = pdata(i);
+
+      // Add to a tensor of data
+      vector_data.emplace_back(data);
+    }
+  } catch (std::exception& exception) {
+    console_->error("{} #{}: {} {}\n", __FILE__, __LINE__, exception.what(),
+                    attribute);
+    vector_data.clear();
+  }
+  return vector_data;
+}
+
 //! Return particle tensor data
 template <unsigned Tdim>
 template <unsigned Tsize>
@@ -1099,7 +1124,7 @@ std::vector<Eigen::Matrix<double, Tsize, 1>>
       Eigen::Matrix<double, Tsize, 1> data;
       data.setZero();
       auto pdata = (*pitr)->tensor_data(attribute);
-      // Fill stresses to the size of dimensions
+      // Fill tensor_data to the size of dimensions
       for (unsigned i = 0; i < pdata.size(); ++i) data(i) = pdata(i);
 
       // Add to a tensor of data
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1075,6 +1075,18 @@ std::vector<Eigen::Matrix<double, 3, 1>>
   return particle_coordinates;
 }
 
+//! Return particle scalar data
+template <unsigned Tdim>
+std::vector<double> mpm::Mesh<Tdim>::particles_scalar_data(
+    const std::string& attribute) {
+  std::vector<double> scalar_data;
+  scalar_data.reserve(particles_.size());
+  // Iterate over particles and add scalar value to data
+  for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr)
+    scalar_data.emplace_back((*pitr)->scalar_data(attribute));
+  return scalar_data;
+}
+
 //! Return particle tensor data
 template <unsigned Tdim>
 template <unsigned Tsize>
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -1092,22 +1092,16 @@ template <unsigned Tdim>
 std::vector<Eigen::Matrix<double, 3, 1>> mpm::Mesh<Tdim>::particles_vector_data(
     const std::string& attribute) {
   std::vector<Eigen::Matrix<double, 3, 1>> vector_data;
-  try {
-    // Iterate over particles
-    for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr) {
-      Eigen::Matrix<double, 3, 1> data;
-      data.setZero();
-      auto pdata = (*pitr)->vector_data(attribute);
-      // Fill vector_data to the size of dimensions
-      for (unsigned i = 0; i < pdata.size(); ++i) data(i) = pdata(i);
+  // Iterate over particles
+  for (auto pitr = particles_.cbegin(); pitr != particles_.cend(); ++pitr) {
+    Eigen::Matrix<double, 3, 1> data;
+    data.setZero();
+    auto pdata = (*pitr)->vector_data(attribute);
+    // Fill vector_data to the size of dimensions
+    for (unsigned i = 0; i < pdata.size(); ++i) data(i) = pdata(i);
 
-      // Add to a tensor of data
-      vector_data.emplace_back(data);
-    }
-  } catch (std::exception& exception) {
-    console_->error("{} #{}: {} {}\n", __FILE__, __LINE__, exception.what(),
-                    attribute);
-    vector_data.clear();
+    // Add to a tensor of data
+    vector_data.emplace_back(data);
   }
   return vector_data;
 }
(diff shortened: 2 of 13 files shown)


include/solvers/mpm_base.h
```diff
@@ -24,6 +24,13 @@
 
 namespace mpm {
 
+//! Variable type
+//! Scalar: boolean, unsigned, int, double
+//! Vector: Vector of size 3
+//! Tensor: Symmetric tensor arranged in voigt notation
+enum class VariableType { Scalar, Vector, Tensor };
+extern std::map<std::string, VariableType> variables;
```
kks32: What is the reasoning behind defining the variable outside the class?

include/solvers/mpm_base.h
```diff
@@ -192,6 +199,8 @@ class MPMBase : public MPM {
   std::map<unsigned, std::shared_ptr<mpm::Material<Tdim>>> materials_;
   //! Mathematical functions
   std::map<unsigned, std::shared_ptr<mpm::FunctionBase>> math_functions_;
+  //! VTK particle variables
+  std::map<mpm::VariableType, std::vector<std::string>> vtk_vars_;
```
kks32: map is slow, use unordered_map instead or tsl::robin_map

include/solvers/mpm_base.h
kks32: Defining inside the class gives better control of what variable can be listed here

include/solvers/mpm_base.tcc
```diff
@@ -92,6 +92,39 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+        post_process_.at("vtk").size() > 0) {
+      // Iterate over vtk
+      for (unsigned i = 0; i < post_process_.at("vtk").size(); ++i) {
+        std::string attribute =
+            post_process_["vtk"][i].template get<std::string>();
+        if (mpm::variables.find(attribute) != mpm::variables.end())
+          vtk_vars_[mpm::variables.at(attribute)].emplace_back(attribute);
+        else {
+          console_->warn(
+              "{} #{}: VTK variable \"{}\" was specified, but is not available "
```
```suggestion
              "{} #{}: VTK variable {} was specified, but is not available "
```

include/solvers/mpm_base.tcc
```diff
@@ -92,6 +92,39 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+              "in variable list",
+              __FILE__, __LINE__, attribute);
+        }
+      }
+    } else {
```
kks32: There is no reason this should be a runtime error. Use console warn instead.

kks32 (changes requested):

include/solvers/mpm_base.h
bodhinandach: I tried doing that, but the test won't compile for multiple definitions.
image

include/solvers/mpm_base.tcc
bodhinandach: I just followed the one in `state_var` and other input functions with `try` and `catch` just to make it the same.

kks32: Could you change it to console->warn, it shound't throw.

bodhinandach: Should I change the other part too in this PR or the next one?

kks32: Both please

bodhinandach: @kks32 Just to confirm, you want me to change all the catch and throw here too?
https://github.com/cb-geo/mpm/blob/2481028eb09f0cfd7da91bd27dfb860dc5dff272/include/solvers/mpm_base.tcc#L24-L93

kks32: No need to change those. Just change vtk_statevariable

include/solvers/mpm_base.h
```diff
@@ -24,6 +24,13 @@
 
 namespace mpm {
 
+//! Variable type
+//! Scalar: boolean, unsigned, int, double
+//! Vector: Vector of size 3
+//! Tensor: Symmetric tensor arranged in voigt notation
+enum class VariableType { Scalar, Vector, Tensor };
+extern tsl::robin_map<std::string, VariableType> variables;
```
kks32: I was thinking of moving it inside the class similar to `vtk_vars_`

include/solvers/mpm_base.tcc
bodhinandach: Okay, I will make the change for the vtk and the vtk_statevars in PR #683 so they are not conflicting

include/solvers/mpm_base.h
bodhinandach: Which class do you mean here?

kks32: In to MPMBase, so we can derive them in MPMExplicit. This would allow a certain degree of control on what variables are available as output under different solvers.

include/particles/particle.tcc
```diff
@@ -852,10 +855,16 @@ void mpm::Particle<Tdim>::apply_particle_velocity_constraints(unsigned dir,
   this->velocity_(dir) = velocity;
 }
 
+//! Return particle scalar data
+template <unsigned Tdim>
+double mpm::Particle<Tdim>::scalar_data(const std::string& property) {
+  return this->scalar_properties_.at(property)();
```
kks32: We need to return a `nan` if we can't find the scalar property, otherwise if we have multiple materials in the mesh and we write vtk file, where one material has a property that the other doesn't, then the code will crash. We should do this for state variables too.

kks32 (changes requested):

include/particles/particle.tcc
bodhinandach: I agree with you. I will change the one in `state_vars` too. How about `tensor_properties` though? We might have `liquid_velocities` or `permeability_` for instance. They are dimensional vector.

kks32: We return nans for scalar, vector and tensorial properties. Do we not have vector_properties anymore?

bodhinandach: Nope, it was combined before in your `tensor_data()` function as `properties_`. I changed it to `tensor_properties_` after having `scalar_properties_`. I can separate them to make it coherence and clearer with the VariableType naming. Will that be good?

include/particles/particle.h
```diff
@@ -241,8 +241,11 @@ class Particle : public ParticleBase<Tdim> {
...
       const std::string& var,
       unsigned phase = mpm::ParticlePhase::Solid) const override {
-    return (state_variables_[phase].find(var) != state_variables_[phase].end())
-               ? state_variables_[phase].at(var)
+    return (phase < state_variables_.size())
+               ? (state_variables_[phase].find(var) !=
+                  state_variables_[phase].end())
+                     ? state_variables_[phase].at(var)
+                     : std::numeric_limits<double>::quiet_NaN()
                : std::numeric_limits<double>::quiet_NaN();
```
kks32:
```suggestion
    return (phase < state_variables_.size() && ((state_variables_[phase].find(var) !=
                  state_variables_[phase].end())
                     ? state_variables_[phase].at(var)
                     : std::numeric_limits<double>::quiet_NaN();
```

include/particles/particle.h
```diff
@@ -258,12 +261,14 @@ class Particle : public ParticleBase<Tdim> {
...
-            state_variables_[phase].end())
-               ? state_variables_[phase].at("pressure")
-               : std::numeric_limits<double>::quiet_NaN();
+    return this->state_variable("pressure", phase);
   }
 
+  //! Return scalar data of particles
+  //! \param[in] property Property string
+  //! \retval vecdata Scalar data of particle property
+  double scalar_data(const std::string& property) override;
```
kks32: Could you inline these functions and const them?

include/particles/particle.tcc
```diff
@@ -243,11 +243,14 @@ void mpm::Particle<Tdim>::initialise() {
...
-  this->properties_["velocities"] = [&]() { return velocity(); };
-  this->properties_["displacements"] = [&]() { return displacement(); };
+  // Initialize scalar and tensor data properties
+  this->scalar_properties_["mass"] = [&]() { return mass(); };
+  this->scalar_properties_["volume"] = [&]() { return volume(); };
+  this->scalar_properties_["mass_density"] = [&]() { return mass_density(); };
+  this->tensor_properties_["stresses"] = [&]() { return stress(); };
+  this->tensor_properties_["strains"] = [&]() { return strain(); };
+  this->tensor_properties_["velocities"] = [&]() { return velocity(); };
+  this->tensor_properties_["displacements"] = [&]() { return displacement(); };
```
kks32: Does this affect vtk vector?

kks32: why do we have vector variables in MPM but not here?

kks32 (changes requested):

include/particles/particle.tcc
bodhinandach: > why do we have vector variables in MPM but not here?

Initially, they are generally called as `properties_`, you call it through the `tensor_data()` function which returns an `Eigen::VectorXd`. When passing to the io, we are calling the `particles_tensor_data()` function in mesh class with templated size, i.e. 3 for vector and 6 for tensor. See this: https://github.com/cb-geo/mpm/blob/a1023baa55cf387275a274aed6b9b955f432d2ad/include/solvers/mpm_base.tcc#L572-L575
and
https://github.com/cb-geo/mpm/blob/a1023baa55cf387275a274aed6b9b955f432d2ad/include/solvers/mpm_base.tcc#L593-L596
I am currently following your previous implementation. If you want to distinguish them to `vector_variables_`, we need to change the `particles_tensor_data()` function in `mesh.tcc` and create `vector_data()` function in particle. I am fine with staying as it is or changing it to `vector_data`, both will work.

> Does this affect vtk vector?

It does not. As I mentioned above, the template size allocates the correct size of Eigen matrix for the io.

include/particles/particle.tcc
kks32: I think having `vector_properties_` is good. Wouldn't not having that impact how they are written in vtk? as in would they still be written as vtkvectors? Either way, it's good to have a separate `vector_properties`

kks32: OK. I'm happy not needing a separate `vector_properties` in here.

include/particles/particle.tcc
```diff
@@ -852,10 +855,21 @@ void mpm::Particle<Tdim>::apply_particle_velocity_constraints(unsigned dir,
...
+             : std::numeric_limits<double>::quiet_NaN();
+}
+
 //! Return particle tensor data
 template <unsigned Tdim>
-Eigen::VectorXd mpm::Particle<Tdim>::tensor_data(const std::string& property) {
-  return this->properties_.at(property)();
+inline Eigen::VectorXd mpm::Particle<Tdim>::tensor_data(
+    const std::string& property) const {
+  return this->tensor_properties_.at(property)();
```
kks32: Shouldn't we return `nans` for tensorial quantities too? This may need us to split vector and tensorial properties.

include/solvers/mpm_base.tcc
```diff
@@ -92,6 +92,36 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+  // VTK particle variables
+  // Initialise container with empty vector
+  vtk_vars_.insert(
+      std::make_pair(mpm::VariableType::Scalar, std::vector<std::string>()));
+  vtk_vars_.insert(
+      std::make_pair(mpm::VariableType::Vector, std::vector<std::string>()));
+  vtk_vars_.insert(
+      std::make_pair(mpm::VariableType::Tensor, std::vector<std::string>()));
+
+  if (post_process_.at("vtk").is_array() &&
```
kks32: Should have a similar structure to #687

kks32 (changes requested):

include/particles/particle.tcc
bodhinandach: Done. I still keep the template `Tsize` in `particle_tensor_data` function in `mesh`. Should I remove the template and change default to 6? I am leaving it there so that if one day there is someone wants to output a nonsymmetric tensor with more than 6 independent components, we can slightly modify it.

include/mesh.tcc
```diff
@@ -1075,28 +1075,53 @@ std::vector<Eigen::Matrix<double, 3, 1>>
...
-      auto pdata = (*pitr)->tensor_data(attribute);
-      // Fill stresses to the size of dimensions
-      for (unsigned i = 0; i < pdata.size(); ++i) data(i) = pdata(i);
-
-      // Add to a tensor of data
-      tensor_data.emplace_back(data);
-    }
-  } catch (std::exception& exception) {
-    console_->error("{} #{}: {} {}\n", __FILE__, __LINE__, exception.what(),
-                    attribute);
```
jgiven100, line 1098: I don't quite follow the benefit of removing this `catch` and subsequent `error` message @bodhinandach

tianchiTJ (approved):

include/mesh.h
```diff
@@ -258,15 +258,28 @@ class Mesh {
   //! Return coordinates of particles
   std::vector<Eigen::Matrix<double, 3, 1>> particle_coordinates();
 
+  //! Return particles scalar data
+  //! \param[in] attribute Name of the scalar data attribute
+  //! \retval scalar_data Vector containing scalar properties from particles
+  std::vector<double> particles_scalar_data(const std::string& attribute);
```
kks32: Could we have these functions as `const`?

include/solvers/mpm_base.h
```diff
@@ -156,6 +162,20 @@ class MPMBase : public MPM {
   //! \param[in] damping_props Damping properties
   bool initialise_damping(const Json& damping_props);
 
+ protected:
+  // Variable list
+  tsl::robin_map<std::string, VariableType> variables_ = {
```
kks32: Could we move this to constructor?

include/solvers/mpm_base.h
```diff
@@ -156,6 +162,20 @@ class MPMBase : public MPM {
...
+      {"volume", VariableType::Scalar},
+      {"mass_density", VariableType::Scalar},
+      // Vector variables
+      {"displacements", VariableType::Vector},
+      {"velocities", VariableType::Vector},
+      // Tensor variables
+      {"strains", VariableType::Tensor},
+      {"stresses", VariableType::Tensor}};
+
  protected:
```
kks32: Protected is repeated

include/solvers/mpm_base.tcc
```diff
@@ -92,8 +92,39 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
     abort();
   }
 
-  bool vtk_statevar = false;
+  // VTK particle variables
+  // Initialise container with empty vector
+  vtk_vars_.insert(
```
kks32, line 110: These 3 lines can be in the constructor?

include/solvers/mpm_base.tcc
```diff
@@ -92,8 +92,39 @@ mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) {
...
+      post_process_.at("vtk").is_array() &&
+      post_process_.at("vtk").size() > 0) {
+    // Iterate over vtk
+    for (unsigned i = 0; i < post_process_.at("vtk").size(); ++i) {
+      std::string attribute =
+          post_process_["vtk"][i].template get<std::string>();
+      if (variables_.find(attribute) != variables_.end())
+        vtk_vars_[variables_.at(attribute)].emplace_back(attribute);
+      else {
+        console_->warn(
```
kks32, line 127: Use a bool and avoid repetition

src/mpm.cc
```diff
@@ -6,6 +6,7 @@
 #include "mpm_explicit.h"
 
 namespace mpm {
+
```
kks32: let's remove this?

kks32 (changes requested):

include/solvers/mpm_base.tcc
bodhinandach: I think having repetition is okay. This is just to note which variables are not available in one go. For instance for the following "vtk" input:
image
We know which one are not available in one go:
image

include/solvers/mpm_base.tcc
bodhinandach: Aren't these lines located in the constructor of MPMBase now?

include/solvers/mpm_base.h
bodhinandach: I am not sure which constructor do you mean here, @kks32

include/particles/particle.tcc
```diff
@@ -243,11 +243,14 @@ void mpm::Particle<Tdim>::initialise() {
   volume_ = std::numeric_limits<double>::max();
   volumetric_strain_centroid_ = 0.;
 
-  // Initialize vector data properties
-  this->properties_["stresses"] = [&]() { return stress(); };
-  this->properties_["strains"] = [&]() { return strain(); };
-  this->properties_["velocities"] = [&]() { return velocity(); };
-  this->properties_["displacements"] = [&]() { return displacement(); };
+  // Initialize scalar and tensor data properties
```
kks32:
```suggestion
  // Initialize scalar, vector and tensor data properties
```

include/solvers/mpm_base.h
kks32: `MPMBase` constructor

include/solvers/mpm_base.tcc
kks32: If they are in the constructor, why do we need it again here?

kks32 (changes requested):

include/solvers/mpm_base.tcc
bodhinandach: What I mean is that these lines are written in the MPMBase constructor, see: https://github.com/cb-geo/mpm/blob/da3ba554b18d4d1a7e3757a67be28cab31e24884/include/solvers/mpm_base.tcc#L1-L103

include/solvers/mpm_base.h
bodhinandach: Any thoughts why moving it to MPMBase constructor? Wouldn't it be just the same?

kks32: Will be cleaner to add more variables in the constructor, especially since we have more than 1 variable

include/solvers/mpm_base.tcc
kks32: My question is why are reinitializing?

bodhinandach: `vtk_vars_` was initially empty, it is a map of `VariableType` and `std::vector<std::string>`. We are not reinitializing. We just insert an empty `std::vector<std::string>()` associated to the `VariableType` key. By doing so, the output can loop over those keys as:
https://github.com/cb-geo/mpm/blob/da3ba554b18d4d1a7e3757a67be28cab31e24884/include/solvers/mpm_base.tcc#L553
or:
https://github.com/cb-geo/mpm/blob/da3ba554b18d4d1a7e3757a67be28cab31e24884/include/solvers/mpm_base.tcc#L574
Without those three lines, `vtl_vars_.at(VariableType)` cant be called as it has not been initialized.

kks32: My question is why do we have to do it again here if it's already in the constructor?
```
 // VTK particle variables 
   // Initialise container with empty vector 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Scalar, std::vector<std::string>())); 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Vector, std::vector<std::string>())); 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Tensor, std::vector<std::string>())); 
```

bodhinandach: Sorry @kks32, I don't quite get what you are intending to say. These lines are written in the constructor. Line 97-102 are within the constructor of `MPMBase`, we only do those lines once:
```
//! Constructor 
 template <unsigned Tdim> 
 mpm::MPMBase<Tdim>::MPMBase(const std::shared_ptr<IO>& io) : mpm::MPM(io) { 

...

 // VTK particle variables 
   // Initialise container with empty vector 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Scalar, std::vector<std::string>())); 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Vector, std::vector<std::string>())); 
   vtk_vars_.insert( 
       std::make_pair(mpm::VariableType::Tensor, std::vector<std::string>())); 

...

 }
```

kks32 (approved):

PR #685, opened 2020-08-06
Refactor/mpm solvers to handle different stress update schemes

### LINKED ISSUE #686: Refactor MPM Solver for stress update schemes
# Refactor MPM solver to handle different stress update schemes

## Summary

Replacing multiple if-else branching conditions for solver types and interface with a new StressUpdate and Interface class. This design will use the Factory pattern to handle different StressUpdate types and Interface.

## Motivation

Adding multiple USF/USL/MUSL schemes pollute the MPMExpliciti class with a lot of branching conditions. This new design changes how MPM solvers are written.

## Design Detail

The current design has a lot of if-else conditions for checking stress update schemes. USF needs to compute stress first. So it's structure looks like:
```
    // Update stress first
    if (this->stress_update_ == mpm::StressUpdate::USF)
      this->compute_stress_strain(phase);
```
Current solver supports `usf` and `usl`. When adding new schemes like `USAVG` and `MUSL`. This leads to multiple checks in each iteration of the loop. This RFC proposes deriving a new StressUpdate class and create derived classes for different StressUpdate schemes. StressUpdate will have the following design. 

```
template <unsigned Tdim>
class StressUpdate {
 public:
  //! Default constructor with mesh class
  StressUpdate(const std::shared_ptr<mpm::Mesh<Tdim>>& mesh, double dt);

  //! Intialize
  void initialise();

  //! Map mass and momentum to nodes
  void momentum_nodes(unsigned phase);

  //! Compute stress and strain
  //! \param[in] phase Phase to smooth pressure
  //! \param[in] pressure_smoothing Enable or disable pressure smoothing
  void compute_stress_strain(unsigned phase, bool pressure_smoothing);

  //! Precompute stress and strain (empty call)
  //! \param[in] phase Phase to smooth pressure
  //! \param[in] pressure_smoothing Enable or disable pressure smoothing
  virtual void precompute_stress_strain(unsigned phase,
                                        bool pressure_smoothing) = 0;

  //! Postcompute stress and strain (empty call)
  //! \param[in] phase Phase to smooth postssure
  //! \param[in] postssure_smoothing Enable or disable postssure smoothing
  virtual void postcompute_stress_strain(unsigned phase,
                                         bool pressure_smoothing) = 0;

  //! Pressure smoothing
  //! \param[in] phase Phase to smooth pressure
  //! \param[in] pressure_smoothing Enable or disable pressure smoothing
  void pressure_smoothing(unsigned phase);

  //! Compute forces
  //! \param[in] gravity Acceleration due to gravity
  //! \param[in] step Number of step in solver
  //! \param[in] concentrated_nodal_forces Boolean for if a concentrated force
  //! is applied or not
  void compute_forces(const Eigen::Matrix<double, Tdim, 1>& gravity,
                      unsigned phase, unsigned step,
                      bool concentrated_nodal_forces);

  //! Compute acceleration velocity position
  //! \param[in] velocity_update Velocity or acceleration update flag
  //! \param[in] phase Phase of particle
  //! \param[in] damping_type Type of damping
  //! \param[in] damping_factor Value of critical damping
  void compute_particle_kinematics(bool velocity_update, unsigned phase,
                                   const std::string& damping_type,
                                   double damping_factor);

  //! Compute particle location
  //! \param[in] locate_particles Flag to enable locate particles, if set to
  //! false, unlocated particles will be removed
  void locate_particles(bool locate_particles);

  //! Stress update scheme
  //! \retval scheme Stress update scheme
  virtual std::string scheme() const = 0;
```

The `precompute_stress_strain` and `post_compute_stress_strain` will be modified in USF and USL methods. In USF for example, the stress update and scheme output will look like:

```
//! Precompute stresses and strains
template <unsigned Tdim>
void mpm::StressUpdateUSF<Tdim>::precompute_stress_strain(
    unsigned phase, bool pressure_smoothing) {
  this->compute_stress_strain(phase, pressure_smoothing);
}

//! Postcompute stresses and strains
template <unsigned Tdim>
void mpm::StressUpdateUSF<Tdim>::postcompute_stress_strain(
    unsigned phase, bool pressure_smoothing) {}

//! Stress update scheme
template <unsigned Tdim>
std::string mpm::StressUpdateUSF<Tdim>::scheme() const {
  return "USF";
}
```

Similar to the `StressUpdate` design, we will also alter the `Interface` and `InterfaceContact` class. `InterfaceContact` needs a more appropriate name - TBD. Unlike the `StressUpdate`, the `Interface` class will have empty void functions, so if no interface model is defined, nothing will be performed.

```
  //! Intialize
  virtual void initialise(){};

  //! Compute contact forces
  virtual void compute_contact_forces(){};
```

A factory register will be used to create both StressUpdate scheme and Interface model

## Drawbacks

* The performance of this design needs to be checked against the basic version of if-else

* Some additional refactoring may be necessary in the future to split certain function calls.

## Rationale and Alternatives

> Why is this design the best in the space of possible designs?
This design avoids branching, unlike a standard if-else and offers a cleaner interface to add different StressUpdate schemes. 

> What other designs have been considered and what is the rationale for not choosing them?
If-else conditions - ugly and unwieldy to maintain as more stress update schemes are added.

Maintaining the MPM Solver becomes tedious.

## Unresolved questions

Check the performance of the code with this new design. 

> What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?

A draft PR is at: https://github.com/cb-geo/mpm/pull/685

kks32:
**Describe the PR**
Replacing multiple if-else branching conditions for solver types and interface with a new StressUpdate and Interface class. This design will use the Factory pattern to handle different StressUpdate types and Interface.

### Performance measurement on Stampede2
Performance tests were conducted for 3D hydrostatic column with USF 10,000 steps
Refactored: 731639 ms 
Dev version: 731874 ms
There is no perceivable difference between the two versions in terms of performance. 

This PR is a work in progress that addresses #686 

### Additional context
- [x] Validation of the MPM code with new design
- [x] Performance benchmark
- [x] Refactor `ContactInterface` to a different name
- [x] Unit tests

**Related Issues/PRs**
An RFC is available at #686


kks32:
Generic factory design isn't efficient as it involves separately initializing 2D and 3D `StressUpdate` classes. The current if-else in the constructor is sufficient.

include/solvers/mpm_explicit.tcc
```diff
@@ -52,41 +63,23 @@ bool mpm::MPMExplicit<Tdim>::solve() {
...
   if (analysis_.find("interface") != analysis_.end())
     interface_ = analysis_.at("interface").template get<bool>();
 
   // Initialise material
-  bool mat_status = this->initialise_materials();
-  if (!mat_status) {
-    status = false;
-    throw std::runtime_error("Initialisation of materials failed");
-  }
+  this->initialise_materials();
```
jgiven100, line 72: This is much easier to read & follow. Nice. :+1:

tests/mpm_explicit_usf_test.cc
```diff
@@ -45,17 +45,17 @@ TEST_CASE("MPM 2D Explicit implementation is checked",
     auto mpm = std::make_unique<mpm::MPMExplicit<Dim>>(std::move(io));
 
     // Initialise materials
-    REQUIRE(mpm->initialise_materials() == true);
+    REQUIRE_NOTHROW(mpm->initialise_materials());
```
jgiven100, line 47: Does `REQUIRE_NOTHROW` indicate which function (e.g. `initialise_materials()`) failed since we removed the `runtime_error` above?

kks32: `REQUIRE_NOTHROW` checks this function `initialise_materials()` doesn't fail. If there is a problem with the function, this condition will be false and the unit test will fail. 

To answer your question, since we are throwing a runtime error, I have added the function name in the error message, so you'll know which function failed if a function fails. We are not catching it locally, but is caught at a higher level.

include/contacts/contact.tcc
```diff
@@ -0,0 +1,16 @@
+//! Constructor of contact with mesh
+template <unsigned Tdim>
+mpm::Contact<Tdim>::Contact(const std::shared_ptr<mpm::Mesh<Tdim>>& mesh) {
+  // Assign mesh
+  mesh_ = mesh;
+  // Initialise MPI rank and size
+  int mpi_rank_ = 0;
+  int mpi_size_ = 1;
```
bodhinandach: Why we need to store it as protected variables? Couldn't we just obtain it if we need? I don't see any call of these variables either.

include/contacts/contact.h
```diff
@@ -0,0 +1,35 @@
...
+  //! Compute contact forces
+  virtual inline void compute_contact_forces(){};
+
+ protected:
+  //! Mesh object
+  std::shared_ptr<mpm::Mesh<Tdim>> mesh_;
+  //! MPI Size
+  int mpi_size_ = 1;
+  //! MPI rank
+  int mpi_rank_ = 0;
```
bodhinandach: I would remove these

include/contacts/contact_friction.h
```diff
@@ -0,0 +1,35 @@
...
+  //! Compute contact forces
+  virtual inline void compute_contact_forces() override;
+
+ protected:
+  //! Mesh object
+  using mpm::Contact<Tdim>::mesh_;
+  //! MPI Size
+  using mpm::Contact<Tdim>::mpi_size_;
+  //! MPI rank
+  using mpm::Contact<Tdim>::mpi_rank_;
```
bodhinandach: These two too.

include/solvers/stress_update/stress_update.h
```diff
@@ -0,0 +1,91 @@
...
+class StressUpdate {
+ public:
+  //! Default constructor with mesh class
+  StressUpdate(const std::shared_ptr<mpm::Mesh<Tdim>>& mesh, double dt);
+
+  //! Intialize
+  virtual inline void initialise();
+
+  //! Map mass and momentum to nodes
+  virtual inline void momentum_nodes(unsigned phase);
```
bodhinandach: Maybe `compute_nodal_kinematics`? I think `momentum_nodes` is not so clear to me. Also, you need \param[in]

include/solvers/stress_update/stress_update.h
```diff
@@ -0,0 +1,91 @@
...
+
+#include "mesh.h"
+
+namespace mpm {
+
+//! StressUpdate class
+//! \brief StressUpdate base class to support different stress update schemes
+//! \tparam Tdim Dimension
+template <unsigned Tdim>
+class StressUpdate {
```
bodhinandach: Instead of `StressUpdate`, I would prefer a more general base class name like `Scheme`. It can be extended for many kind of integration scheme, both in stress and time, and in a different order. So can we have it stored as `scheme/scheme.h`.

include/solvers/mpm_explicit.tcc
```diff
@@ -52,41 +63,23 @@ bool mpm::MPMExplicit<Tdim>::solve() {
...
 
   // Pressure smoothing
-  if (analysis_.find("pressure_smoothing") != analysis_.end())
-    pressure_smoothing_ =
-        analysis_.at("pressure_smoothing").template get<bool>();
+  pressure_smoothing_ = io_->analysis_bool("pressure_smoothing");
 
   // Interface
   if (analysis_.find("interface") != analysis_.end())
     interface_ = analysis_.at("interface").template get<bool>();
```
bodhinandach:
```suggestion
  interface_ = io_->analysis_bool("interface");
```

include/solvers/mpm_base.h
```diff
@@ -178,8 +176,12 @@ class MPMBase : public MPM {
   //! Logger
   using mpm::MPM::console_;
 
-  //! Stress update method (default USF = 0, USL = 1, MUSL = 2)
-  mpm::StressUpdate stress_update_{mpm::StressUpdate::USF};
+  //! Stress update method
+  std::string stress_update_scheme_{"usf"};
+  //! Stress update scheme
+  std::shared_ptr<mpm::StressUpdate<Tdim>> stress_update_{nullptr};
```
bodhinandach: Could we change `stress_update_` into `scheme_` for naming? I think `stress_update_` is good for only explicit MPM, when we have semi-implicit, implicit, or nonlinear implicit solver, I think `scheme_` is more appropriate.

include/solvers/mpm_explicit.tcc
```diff
@@ -4,6 +4,17 @@ mpm::MPMExplicit<Tdim>::MPMExplicit(const std::shared_ptr<IO>& io)
     : mpm::MPMBase<Tdim>(io) {
   //! Logger
   console_ = spdlog::get("MPMExplicit");
+  //! Stress update
+  if (this->stress_update_scheme_ == "usl")
+    stress_update_ = std::make_shared<mpm::StressUpdateUSL<Tdim>>(mesh_, dt_);
+  else
+    stress_update_ = std::make_shared<mpm::StressUpdateUSF<Tdim>>(mesh_, dt_);
```
bodhinandach: Why we read `stres_update_scheme_` in `MPMBase` and create `stress_update_` here? Wouldn't it be better to make them together in `MPMBase`? By doing so, we don't have to store `stress_update_scheme_`.

bodhinandach (changes requested):
Thanks for the refactor @kks32, I generally like the idea and I think this is much cleaner in terms of structuring. I have some comments which you can consider. 

One small request from @tianchiTJ  and I, can we somehow have this PR merge after the two_phase and navier_stokes branch are merged? They have been excessive number of commits there and since this PR refactor the base classes, it will require us to refactor the solver again, i.e. `compute_stress_strains`, `pressure_smoothing`, etc. The refactoring is definitely a very nice idea, but I won't see that as something urgent at the moment.

Anyway, many thanks for the cool idea.

include/solvers/mpm_base.h
kks32: `scheme` may be way too generic, `interface` could have schemes.

include/solvers/mpm_explicit.tcc
kks32: No, we have to initialize them here, there may be methods which are specific to Two-phase or Semi-Implicit, which don't apply to other MPM methods

include/contacts/contact.h
kks32: We need these, as contacts would need mpi_size and rank when doing MPI part

include/contacts/contact_friction.h
kks32: We need these, as contacts would need mpi_size and rank when doing MPI part

include/contacts/contact.tcc
kks32: We need these, as contacts would need mpi_size and rank when doing MPI part

include/solvers/stress_update/stress_update.h
kks32: The functions derived here are specific to StressUpdate.

kks32:
This refactoring doesn't remove `pressure_smoothing` and `compute_stress_strain` in solver class, exactly for this reason that it doesn't break @tianchiTJ and @bodhinandach changes. This shouldn't affect your derived classes and can stay the same. The SemiImplicit and TwoPhase can be refactored later once they are merged to develop. Let me know if this is still an issue.

This change affects the Contact so @thiagordonho has to change the interface class

include/solvers/stress_update/stress_update.h
kks32: I see your thought on stresses and time, I was thinking of splitting them as separate classes. Maybe `MPMScheme`?

include/contacts/contact.tcc
bodhinandach: I see your point, though I am not sure where I can find the contact class is using it. Will that be better to remove it now and add it later? Btw, I think you don't have to put the `int` type here again.
```suggestion
  mpi_rank_ = 0;
  mpi_size_ = 1;
```

include/solvers/mpm_explicit.tcc
bodhinandach: That's a wonderful thought @kks32! Let's go with this.

include/solvers/stress_update/stress_update.h
bodhinandach: `MPMScheme` seems to be a nice one, but as you mention in the other PR, does that means that you are deriving from `MPM` class?

bodhinandach:
@kks32 `pressure_smoothing`, `compute_stress_strain`, and the prior `stress_update_` are the three things we need, I guess. The last part may be an issue for the `two_phase` branch.

tianchiTJ (approved):

include/solvers/stress_update/stress_update.h
kks32: `MPMScheme` is the base class, so it doesn't mean we are deriving from `MPM`. I do see your point, but it would be difficult to have class names which are similar.

kks32:
@tianchiTJ @bodhinandach  I have reverted `stress_update_` so you can use the variable as before. However, the type is a `string` and not an `enum`. So instead of `mpm::StressUpdate::USF` use `usf` and instead of `mpm::StressUpdate::USL` use `usl`. Thanks. Let me know if there are any issues.

bodhinandach (approved):
@kks32 Thanks for addressing all the comments! `MPMScheme` sounds legit to me :tada: :tada: :tada:

bodhinandach:
I would like to recommend everyone @cb-geo/mpm to look at this PR too. I guess it is a good direction for everyone. 
@kks32 Can we merge this after next week's group meeting if that's okay?

jgiven100 (approved):

include/solvers/mpm_scheme/mpm_scheme_usf.tcc
```diff
@@ -0,0 +1,26 @@
+//! Constructor
+template <unsigned Tdim>
+mpm::MPMSchemeUSF<Tdim>::MPMSchemeUSF(
+    const std::shared_ptr<mpm::Mesh<Tdim>>& mesh, double dt)
+    : mpm::MPMScheme<Tdim>(mesh, dt) {
+  mesh_ = mesh;
+  dt_ = dt;
```
yliang-sn: Could we remove these two lines? It's already performed in the basic class.

include/solvers/mpm_scheme/mpm_scheme_usl.tcc
```diff
@@ -0,0 +1,26 @@
+//! Constructor
+template <unsigned Tdim>
+mpm::MPMSchemeUSL<Tdim>::MPMSchemeUSL(
+    const std::shared_ptr<mpm::Mesh<Tdim>>& mesh, double dt)
+    : mpm::MPMScheme<Tdim>(mesh, dt) {
+  mesh_ = mesh;
+  dt_ = dt;
```
yliang-sn: Could we remove these two lines? It's already performed in the basic class.

yliang-sn (approved):
Thanks! @kks32. I feel it's much clearer than before.

thiagordonho (approved):

include/solvers/mpm_explicit.tcc
```diff
@@ -4,6 +4,17 @@ mpm::MPMExplicit<Tdim>::MPMExplicit(const std::shared_ptr<IO>& io)
     : mpm::MPMBase<Tdim>(io) {
   //! Logger
   console_ = spdlog::get("MPMExplicit");
+  //! Stress update
+  if (this->stress_update_ == "usl")
```
ezrayst, line 8: Do we want to include MUSL as well or should this be in subsequent PR?

ezrayst, line 8: How does this work for MUSL? Just another if statement inside this?

include/solvers/mpm_explicit.tcc
```diff
@@ -117,190 +109,36 @@ bool mpm::MPMExplicit<Tdim>::solve() {
...
-                    std::placeholders::_1));
-    }
+    contact_->initialise();
 
-    // Assign mass and momentum to nodes
-    mesh_->iterate_over_particles(
-        std::bind(&mpm::ParticleBase<Tdim>::map_mass_momentum_to_nodes,
-                  std::placeholders::_1));
+    // Mass momentum and compute velocity at nodes
+    mpm_scheme_->compute_nodal_kinematics(phase);
```
ezrayst, line 121: Nice idea!

include/solvers/mpm_explicit.tcc
```diff
@@ -117,190 +109,36 @@ bool mpm::MPMExplicit<Tdim>::solve() {
...
-#pragma omp section
-      {
-        // Spawn a task for internal force
-        // Iterate over each particle to compute nodal internal force
-        mesh_->iterate_over_particles(
-            std::bind(&mpm::ParticleBase<Tdim>::map_internal_force,
-                      std::placeholders::_1));
-      }
-    }  // Wait for tasks to finish
+    mpm_scheme_->precompute_stress_strain(phase, pressure_smoothing_);
```
ezrayst, line 127: Any reason why pressure_smoothing is included in the compute_stress_strain? Isn't supposed to be separate? I think we always need to separate the current stress state and the smoothed one because any plastic constitutive models (especially critical state model) are not happy when the stress state changes.

ezrayst (commented):
Thanks @kks32. In terms of big picture thought process then, we have the following:
- MPMExplicit as the MPM type that defines the solution scheme and the phase
- MPMScheme as the detail within the type which houses the stress update scheme, even what to do with throw elements - in the future I want to get into more integration schemes.
- Analysis would be separate from the MPMScheme then? How does the JSON look like?

include/solvers/mpm_explicit.tcc
bodhinandach: I can resume the PR #682 with the test after this PR is merged

kks32: Yes

include/solvers/mpm_explicit.tcc
kks32: This is just a flag to say you want to pressure smoothing or not

ezrayst (approved):
LGTM!

kks32:
> Thanks @kks32. In terms of big picture thought process then, we have the following:
>     * MPMExplicit as the MPM type that defines the solution scheme and the phase
> 
>     * MPMScheme as the detail within the type which houses the stress update scheme, even what to do with throw elements - in the future I want to get into more integration schemes.
> 
>     * Analysis would be separate from the MPMScheme then? How does the JSON look like?

Nothing changes in the JSON.

PR #691, opened 2020-08-17
[Solver] XMPM solver to handle predefined discontinuity

yliang-sn:
**Describe the PR**
A draft PR to ease to check the solver structure.


include/mesh.h
```diff
@@ -447,6 +448,33 @@ class Mesh {
...
   void initialise_nodal_properties();
 
+  //! Set particles lsm values
+  void assign_particle_levelset(std::vector<double>& phi_list);
+
+  // Create the nodal properties' map for discontinuity
+  void create_nodal_properties_discontinuity();
+
+  //! Initialise discontinuities
+  //! \param[in] discontinuities
```
kks32:
```suggestion
  //! \param[in] discontinuities List of discontinuities 
```

include/mesh.h
```diff
@@ -447,6 +448,33 @@ class Mesh {
...
+  void initialise_discontinuities(
+      const std::map<unsigned, std::shared_ptr<mpm::DiscontinuityBase<Tdim>>>&
+          discontinuities) {
+    discontinuities_ = discontinuities;
+  }
+
+  //! Locate points of discontinuity in a cell
+  void locate_discontinuity_mesh();
```
kks32:
```suggestion
  void locate_discontinuity();
```
Mesh would be redundant here

include/mesh.h
```diff
@@ -447,6 +448,33 @@ class Mesh {
   // Initialise the nodal properties' map
   void initialise_nodal_properties();
 
+  //! Set particles lsm values
+  void assign_particle_levelset(std::vector<double>& phi_list);
```
kks32: Pass as `const&`
```suggestion
  void assign_particle_levelset(const std::vector<double>& phi_list);
```

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
 template <unsigned Tdim>
 void mpm::Mesh<Tdim>::initialise_nodal_properties() {
   // Call initialise_properties function from the nodal properties
   nodal_properties_->initialise_nodal_properties();
 }
+
+//! Set particles lsm values
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::assign_particle_levelset(std::vector<double>& phi_list) {
+  for (mpm::Index i = 0; i < nparticles(); ++i)
```
kks32: We could make this parallel as it has no dependencies

include/node.tcc
```diff
@@ -34,6 +34,8 @@ void mpm::Node<Tdim, Tdof, Tnphases>::initialise() noexcept {
   acceleration_.setZero();
   status_ = false;
   material_ids_.clear();
+  //mark: need to check
+  discontinuity_enrich_ = true;
```
kks32, line 38: In the header, this is set to false.

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
   }
 }
 
+// Create the nodal properties' map for discontinuity
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::create_nodal_properties_discontinuity() {
+  // Initialise the shared pointer to nodal properties
+  if (nodal_properties_ == nullptr)
```
kks32, line 1963: We can move this to an assert condition

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
+                                       nrows, 1);
+    nodal_properties_->create_property("friction_coef",
+                                       nodes_.size(), 1);
+    // Iterate over all nodes to initialise the property handle in each node
+    // and assign its node id as the prop id in the nodal property data pool
+    for (auto nitr = nodes_.cbegin(); nitr != nodes_.cend(); ++nitr)
+      (*nitr)->initialise_discontinuity_property_handle((*nitr)->id(),
+                                                        nodal_properties_);
+  } else {
+    throw std::runtime_error("Number of nodes is zero");
```
kks32: I'll move this to assert as well, because we can't recover from this message.

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
+void mpm::Mesh<Tdim>::assign_particle_levelset(std::vector<double>& phi_list) {
+  for (mpm::Index i = 0; i < nparticles(); ++i)
+    particles_[i]->assign_levelsetphi(phi_list[i]);
+}
+
+//! Locate points in a cell
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::locate_discontinuity_mesh() {
+  for (unsigned i = 0; i < discontinuities_.size(); ++i) {
+    auto discontinuity = discontinuities_[i];
```
```suggestion
    discontinuities_[i]->locate_discontinuity_mesh(cells_, map_cells_);
```

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
+    discontinuity->compute_shapefn();
+  }
+}
+
+// compute the normal vector of enriched nodes at the discontinuity
+template <unsigned Tdim>
+void mpm::Mesh<Tdim>::compute_normal_vector_discontinuity() {
+  // just compute for one discontinuity for now
+  unsigned discontinuity_id = 0;
+  auto discontinuity = discontinuities_[discontinuity_id];
```
kks32: This will always use the id of 0?

include/node.h
```diff
@@ -264,13 +274,59 @@ class Node : public NodeBase<Tdim> {
...
+  void compute_normal_vector() noexcept override;
+
  private:
   //! Mutex
   SpinMutex node_mutex_;
   //! nodebase id
   Index id_{std::numeric_limits<Index>::max()};
   //! nodal property id
   unsigned prop_id_{std::numeric_limits<unsigned>::max()};
+  //! nodal property id
```
kks32: discontinuity property id

include/node_xmpm.tcc
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// Return data in the nodal properties map at a specific index
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+Eigen::MatrixXd mpm::Node<Tdim, Tdof, Tnphases>::discontinuity_property(
+    const std::string& property, unsigned nprops) noexcept {
+  // Const pointer to location of property: node_id * nprops x mat_id
+  auto property_value =
+      property_handle_->property(property, discontinuity_prop_id_, 0, nprops);
+  ;
```
kks32: extra ;

include/particles/particle_xmpm.h
```diff
@@ -0,0 +1,161 @@
...
+  ParticleXMPM(Index id, const VectorDim& coord);
+
+  //! Construct a particle with id, coordinates and status
+  //! \param[in] id Particle id
+  //! \param[in] coord coordinates of the particle
+  //! \param[in] status Particle status (active / inactive)
+  ParticleXMPM(Index id, const VectorDim& coord, bool status);
+
+  //! Destructor
+  ~ParticleXMPM() override{};
```
kks32:
```suggestion
  ~ParticleXMPM() override = default;
```

include/particles/particle_xmpm.h
```diff
@@ -0,0 +1,161 @@
...
+
+  //! Assign the level set function values
+  //! \param[in] phivalue The level set values
+  void assign_levelsetphi(const double phivalue) { levelset_phi_ = phivalue; };
+
+  //! Return 1 if x > 0, -1 if x < 0 and 0 if x = 0
+  //! \param[in] x double value
+  inline double sgn(double x) noexcept {
+    return (x > 0) ? 1. : ((x < 0) ? -1. : 0);
+  };
```
kks32: You can remove ; after }

include/particles/particle_xmpm.h
```diff
@@ -0,0 +1,161 @@
...
+  inline Eigen::Matrix<double, 6, 1> compute_strain_rate(
+      const Eigen::MatrixXd& dn_dx, unsigned phase) noexcept;
+
+  //! Assign the level set function values
+  //! \param[in] phivalue The level set values
+  void assign_levelsetphi(const double phivalue) { levelset_phi_ = phivalue; };
+
+  //! Return 1 if x > 0, -1 if x < 0 and 0 if x = 0
+  //! \param[in] x double value
+  inline double sgn(double x) noexcept {
```
kks32, line 76: This is a generic function, we could define it somewhere more centrally in `utilities`?

```
template <typename T> int sgn(T val) {
    return (T(0) < val) - (val < T(0));
}
```

include/solvers/xmpm_explicit.tcc
```diff
@@ -0,0 +1,463 @@
...
+  if (pressure_smoothing_) this->pressure_smoothing(phase);
+
+  // Iterate over each particle to compute stress
+  mesh_->iterate_over_particles(std::bind(
+      &mpm::ParticleBase<Tdim>::compute_stress, std::placeholders::_1));
+}
+
+//! MPM Explicit solver
+template <unsigned Tdim>
+bool mpm::XMPMExplicit<Tdim>::solve() {
```
kks32, line 41: Could you please use the updated structure similar to https://github.com/cb-geo/mpm/pull/685

include/solvers/xmpm_explicit.tcc
```diff
@@ -0,0 +1,463 @@
...
+                    exception.what());
+    status = false;
+  }
+
+  return status;
+}
+
+// Initialise particles
+template <unsigned Tdim>
+bool mpm::XMPMExplicit<Tdim>::initialise_levelset() {
```
kks32: We can make these functions void, there is no need to try-catch if we can't recover from the error

include/xmpm/discontinuity_base.h
```diff
@@ -0,0 +1,246 @@
...
+  //! self-contact
+  bool self_contact_{true};
+
+  //! friction coefficient
+  double friction_coef_;
+
+};  // DiscontinuityBase class
+
+
+//! struct of discontinuity point
```
kks32, line 110: Could you move all of the following classes to a separate file?

include/xmpm/discontinuity_base.h
```diff
@@ -0,0 +1,246 @@
...
+template <unsigned Tdim>
+struct discontinuity_point {
```
kks32, line 112: why a struct instead of a class?

include/xmpm/discontinuity_base.h
```diff
@@ -0,0 +1,246 @@
...
+
+  //! assign the surface center
+  //! \param[in] center coordinates of the surface center
+  inline void assign_center(VectorDim& center) { center_ = center; }
+
+  //! assign the surface normal vector
+  //! \param[in] normal normal vector of the surface
+  inline void assign_normal(VectorDim& normal) { normal_ = normal; }
+
+  //! Reture normal of the elements
```
kks32:
```suggestion
  //! Return normal of the elements
```

include/xmpm/discontinuity_base.h
```diff
@@ -0,0 +1,246 @@
...
+  //! \param[in]  coor coordinates
+  double vertical_distance(const VectorDim& coor) const {
+    return (coor[0] - center_[0]) * normal_[0] +
+           (coor[1] - center_[1]) * normal_[1] +
+           (coor[2] - center_[2]) * normal_[2];
+  };
+
+ private:
+  //! points indices
+  Eigen::Matrix<mpm::Index, 3, 1> points_;
```
kks32, line 234: Are these triangles?

include/xmpm/discontinuity_base.h
```diff
@@ -0,0 +1,246 @@
...
+//! struct of discontinuity surface: triangle
+template <unsigned Tdim>
+struct discontinuity_surface {
+ public:
+  //! Define a vector of size dimension
+  using VectorDim = Eigen::Matrix<double, Tdim, 1>;
+
+  //! construct with points indices
+  discontinuity_surface(const std::vector<mpm::Index>& points) {
+    for (int i = 0; i < points.size(); ++i) points_[i] = points[i];
```
kks32: Aren't point <3, 1> I don't follow this logic, could you please explain what's happening?

kks32 (changes requested):
[Partial review] This is great Yong. 

1. Could you please explain how you plan to handle surface/lines distributed across several MPI ranks?

2. Is the 3D surface always defined by triangular elements?

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
+  if (nodal_properties_ == nullptr)
+    nodal_properties_ = std::make_shared<mpm::NodalProperties>();
+
+  // Check if nodes_ is empty and throw runtime error if they are
+  if (nodes_.size() != 0) {
+    // Compute number of rows in nodal properties for vector entities
+    const unsigned nrows = nodes_.size() * Tdim;
+    // Create pool data for each property in the nodal properties struct
+    // object. Properties must be named in the plural form
+    nodal_properties_->create_property("mass_enrich", nodes_.size(), 1);
```
thiagordonho: I would change this to "masses_enrich" to keep all as plural

include/mesh.tcc
```diff
@@ -1885,9 +1885,92 @@ void mpm::Mesh<Tdim>::create_nodal_properties() {
...
+    nodal_properties_->create_property("momenta_enrich", nrows, 1);
+    nodal_properties_->create_property("internal_force_enrich", nrows, 1);
+    nodal_properties_->create_property("external_force_enrich", nrows, 1);
```
thiagordonho: Same thing here for the forces: "internal_forces_enrich" and "external_forces_enrich" in plural form

include/node.h
```diff
@@ -264,13 +274,59 @@ class Node : public NodeBase<Tdim> {
...
+  // Return data in the nodal discontinuity properties map at a specific index
+  // \param[in] property Property name
+  // \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  Eigen::MatrixXd discontinuity_property(const std::string& property,
+                                         unsigned nprops = 1) noexcept override;
+
+  //! Compute momentum
+  //! \param[in] phase Index corresponding to the phase
+  //! \param[in] dt Timestep in analysis
+  virtual bool intergrate_momentum_discontinuity(unsigned phase,
```
thiagordonho: What if we name this `compute_momentum_discontinuity`, so its name is comparable to other functions that also deal with computations in the node and particle classes.

include/node.h
```diff
@@ -264,13 +274,59 @@ class Node : public NodeBase<Tdim> {
...
+  //! \param[in] dt Time-step
+  void self_contact_discontinuity(double dt) noexcept override;
+
+  //! Return the discontinuity_prop_id
+  virtual unsigned discontinuity_prop_id() const noexcept override {
+    return discontinuity_prop_id_;
+  };
+
+  //! Compute normal direction for discontinuity
+  void compute_normal_vector() noexcept override;
```
thiagordonho: Perhaps change this to `compute_normal_vector_discontinuity` since there is also another function that computes normal vector but for a multimaterial interface.

include/node_base.h
```diff
@@ -253,6 +263,45 @@ class NodeBase {
...
+  //! \param[in] property Property name
+  //! \param[in] property_value Property quantity from the particles in the cell
+  //! \param[in] discontinuity_id Id of the material within the property data
+  //! \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  virtual void update_discontinuity_property(
+      bool update, const std::string& property,
+      const Eigen::MatrixXd& property_value, unsigned discontinuity_id,
+      unsigned nprops) noexcept = 0;
+
+  //! Compute momentum
```
thiagordonho:
```suggestion
  //! Compute momentum in discontinuity
```

include/node.h
```diff
@@ -264,13 +274,59 @@ class Node : public NodeBase<Tdim> {
...
+                                     unsigned discontinuity_id,
+                                     unsigned nprops) noexcept override;
+
+  // Return data in the nodal discontinuity properties map at a specific index
+  // \param[in] property Property name
+  // \param[in] nprops Dimension of property (1 if scalar, Tdim if vector)
+  Eigen::MatrixXd discontinuity_property(const std::string& property,
+                                         unsigned nprops = 1) noexcept override;
+
+  //! Compute momentum
```
thiagordonho:
```suggestion
  //! Compute momentum in discontinuity
```

include/node_base.h
```diff
@@ -253,6 +263,45 @@ class NodeBase {
...
+
+  //! Apply self-contact of the discontinuity
+  //! \param[in] dt Time-step
+  virtual void self_contact_discontinuity(double dt) noexcept = 0;
+
+  //! Return the discontinuity_prop_id
+  virtual unsigned discontinuity_prop_id() const noexcept = 0;
+
+  //! Compute normal direction for discontinuity
+  virtual void compute_normal_vector() noexcept = 0;
```
thiagordonho:
```suggestion
  virtual void compute_normal_vector_discontinuity() noexcept = 0;
```

include/node_xmpm.tcc
```diff
@@ -0,0 +1,206 @@
...
+  // mpm::MapProperty property_handle(position, nprops);
+  return property_value;
+}
+
+//! Compute acceleration and velocity with cundall damping factor
```
thiagordonho:
```suggestion
//! Compute momentum in discontinuity
```

include/node_xmpm.tcc
```diff
@@ -0,0 +1,206 @@
...
+      // Apply boundary condition in local coordinate
+      local_velocity(direction, phase) = constraint.second;
+      local_acceleration(direction, phase) = 0.;
+      // Transform back to global coordinate
+      this->velocity_ = rotation_matrix_ * local_velocity;
+      this->acceleration_ = rotation_matrix_ * local_acceleration;
+    }
+  }
+}
+//! Apply velocity constraints
```
thiagordonho:
```suggestion
//! Apply self-contact of the discontinuity
```

include/particles/particle_xmpm.tcc
```diff
@@ -0,0 +1,266 @@
...
+  this->initialise();
+  //! Logger
+  std::string logger =
+      "particlexmpm" + std::to_string(Tdim) + "d::" + std::to_string(id);
+  console_ = std::make_unique<spdlog::logger>(logger, mpm::stdout_sink);
+}
+
+// Initialise particle properties
+template <unsigned Tdim>
+void mpm::ParticleXMPM<Tdim>::initialise() {
```
thiagordonho: I believe this function is currently not needed since it is only used inside the constructor but `levelset_phi_` has already been assigned an initial value of 0. in the `particle_xmpm.h`

include/xmpm/discontinuity_3d.tcc
```diff
@@ -0,0 +1,169 @@
...
+  } catch (std::exception& exception) {
+    console_->error("{} #{}: {}\n", __FILE__, __LINE__, exception.what());
+    status = false;
+  }
+  return status;
+}
+
+// return the cross product of ab X bc
+template <unsigned Tdim>
+Eigen::Matrix<double, Tdim, 1> mpm::Discontinuity3D<Tdim>::three_cross_product(
```
thiagordonho, line 115: Should this function be in `utilities` as well? @kks32 , what do you think?

include/solvers/xmpm_explicit.tcc
```diff
@@ -0,0 +1,463 @@
...
+      mesh_->template nodal_halo_exchange<Eigen::Matrix<double, Tdim, 1>, Tdim>(
+          std::bind(&mpm::NodeBase<Tdim>::internal_force, std::placeholders::_1,
+                    phase),
+          std::bind(&mpm::NodeBase<Tdim>::update_internal_force,
+                    std::placeholders::_1, false, phase,
+                    std::placeholders::_2));
+    }
+#endif
+
+    // intergrate momentum Iterate over
```
thiagordonho:
```suggestion
    // integrate momentum by iterating over nodes
```

thiagordonho (changes requested):
Well done with this portion of the code so far. That is a lot of work done!
Here are a few overall comments:
1. I also agree with Krishna's suggestions
2. Why is the XMPM explicit solver in a separate file?
3. Perhaps it would be good to standardize the nomenclature of the discontinuity functions (e.g., all with discontinuity in the end, such as `compute_momentum_discontinuity`). This is something I also want to do for the Contact algorithm but I haven't done it yet. We can use the same structure to keep the code as similar as possible.

include/mesh.tcc
bodhinandach: Why don't we loop as before using:
```
for (unsigned i = 0; i < discontinuities_.size(); ++i) {
    auto discontinuity = discontinuities_[i];
    ...
    following operations
}
```

include/particles/particle_base.h
```diff
@@ -282,6 +282,9 @@ class ParticleBase {
   //! Return neighbour ids
   virtual std::vector<mpm::Index> neighbours() const = 0;
 
+  //! set the level set function values
```
bodhinandach, line 317:
```suggestion
  //! set the level set function values
  //! \param[in] phivalue Signed distance function
```

include/particles/particle_base.h
```diff
@@ -282,6 +282,9 @@ class ParticleBase {
...
+  virtual void assign_levelsetphi(const double phivalue){};
```
bodhinandach: no need to const

include/particles/particle_xmpm.tcc
```diff
@@ -0,0 +1,266 @@
...
+
+  // Compute at centroid
+  // Strain rate for reduced integration
+  const Eigen::Matrix<double, 6, 1> strain_rate_centroid =
+      this->compute_strain_rate(dn_dx_centroid_, mpm::ParticlePhase::Solid);
+
+  // Assign volumetric strain at centroid
+  dvolumetric_strain_ = dt * strain_rate_centroid.head(Tdim).sum();
+  volumetric_strain_centroid_ += dvolumetric_strain_;
+}
```
bodhinandach: Can we remove line 98-116, they are exactly the same as in Particle class?

include/particles/particle_xmpm.tcc
```diff
@@ -0,0 +1,266 @@
...
+                                               0, 1);
+      nodes_[i]->update_discontinuity_property(true, "momenta_enrich",
+                                               velocity_ * nodal_mass, 0, Tdim);
+    }
+  }
+}
+
+// Compute strain rate of the particle
+template <>
+inline Eigen::Matrix<double, 6, 1> mpm::ParticleXMPM<3>::compute_strain_rate(
```
bodhinandach, line 49: Don't we need to treat the 2D version of discontinuity too? In 2D `compute_strain_rate` will call the base class, which is the one written in `particle.tcc `and does not have the enrichment part.

include/particles/particle_xmpm.tcc
```diff
@@ -0,0 +1,266 @@
...
+    force[2] = dn_dx_(i, 2) * stress_[2] + dn_dx_(i, 1) * stress_[4] +
+               dn_dx_(i, 0) * stress_[5];
+
+    force *= -1. * this->volume_;
+
+    nodes_[i]->update_internal_force(true, mpm::ParticlePhase::Solid, force);
+    if (nodes_[i]->discontinuity_enrich())
+      nodes_[i]->update_discontinuity_property(
+          true, "internal_force_enrich", sgn(levelset_phi_) * force, 0, 3);
+  }
```
bodhinandach, line 161: Rather than rewriting the whole function, which is a code duplicacy, why don't we call the base class function `mpm::Particle<3>::map_internal_force()`?
```suggestion
  // Compute nodal internal forces
  mpm::Particle<3>::map_internal_force();
  // Compute nodal internal forces enrich
  for (unsigned i = 0; i < nodes_.size(); ++i) {
    if (nodes_[i]->discontinuity_enrich())
      nodes_[i]->update_discontinuity_property(
          true, "internal_force_enrich", sgn(levelset_phi_) * force, 0, 3);
  }
```

include/xmpm/discontinuity_3d.tcc
```diff
@@ -0,0 +1,169 @@
...
+  bool status = true;
+  try {
+    // Check if surfs is empty
+    if (surfs.empty()) throw std::runtime_error("List of surfaces is empty");
+    // Iterate over all surfaces
+    for (const auto& points : surfs) {
+
+      mpm::discontinuity_surface<Tdim> surf(points);
+
+      surfaces_.emplace_back(surf);  //
```
bodhinandach: cleanup //

include/xmpm/discontinuity_3d.tcc
```diff
@@ -0,0 +1,169 @@
...
+                                            const Json& discontinuity_props)
+    : DiscontinuityBase<Tdim>(id, discontinuity_props) {
+
+  try {
+    // assign friction_coef_ if it's given in input file
+    if (discontinuity_props.contains("friction_coefficient"))
+      friction_coef_ =
+          discontinuity_props.at("friction_coefficient").template get<double>();
+    else
+      friction_coef_ = 0;
```
bodhinandach:
```suggestion
    friction_coef_ = 0;
    if (discontinuity_props.contains("friction_coefficient"))
      friction_coef_ =
          discontinuity_props.at("friction_coefficient").template get<double>();
```

include/xmpm/discontinuity_3d.tcc
```diff
@@ -0,0 +1,169 @@
...
+    status = false;
+    throw std::runtime_error(
+        "Addition of surfaces in discontinuity to mesh failed");
+  }
+
+  bool normal_status = initialize_center_normal();
+  if (!normal_status) {
+    status = false;
+    throw std::runtime_error(
+        "initialization of the center and normal vector of the discontunity failed");
```
bodhinandach:
```suggestion
        "initialization of the center and the normal vector of the discontinuity failed");
```

include/xmpm/discontinuity_3d.tcc
bodhinandach: I agree with @thiagordonho, this can be put outside as generic function

include/xmpm/discontinuity_3d.tcc
```diff
@@ -0,0 +1,169 @@
...
+
+      surf.assign_center(center);
+
+      // the normal of the surfaces
+      normal = three_cross_product(points_[points[0]].coordinates(),
+                                   points_[points[1]].coordinates(),
+                                   points_[points[2]].coordinates());
+      double det = std::sqrt(normal[0] * normal[0] + normal[1] * normal[1] +
+                             normal[2] * normal[2]);
+      normal = 1.0 / det * normal;
```
bodhinandach:
```suggestion
if (normal.norm() > std::numeric_limits<double>::epsilon())
    normal.normalize();
else
    normal = VectorDim::Zero();
```

bodhinandach (changes requested):
Thanks @yliang-sn for the draft PR. I have some comments and generally agree with @thiagordonho and @kks32 comments earlier. Also, would you be able to run `sh ../clang-tools/format.sh` from your build folder before committing?

include/mesh.tcc
yliang-sn: Yes, I loop for each particle now including computing the level set value. Will do all the parallel for XMPM next.

include/node.tcc
yliang-sn: Yes. It should be. It will be false.

include/mesh.tcc
yliang-sn: > We can move this to an assert condition

It's not an assert condition. We use one `nodal_properties_` for contact and XMPM. I will initialize the point if it's nullptr.

include/mesh.tcc
yliang-sn: Loop now but cant handle multiple discontinuities now.

include/xmpm/discontinuity_base.h
yliang-sn: Because they are defined for discontinuity class. moving to another file will bring too many files: point line area. How about put them together?

include/xmpm/discontinuity_base.h
yliang-sn: No  particular reason. Do you think class will be better? I can switch to class.

include/xmpm/discontinuity_base.h
yliang-sn: Yes.

include/xmpm/discontinuity_base.h
yliang-sn: The discontinuity_surface (triangles) is constructed with three points index from the input file as we created cells.

include/mesh.tcc
yliang-sn: How about keep the same name with the quantities in` node.h`? Always use the singular form. which one will be better?

include/particles/particle_xmpm.tcc
yliang-sn: Finish 3D first and i will enrich the 2D part.

include/particles/particle_xmpm.tcc
yliang-sn: By doing that we need to compute the force twice.

include/xmpm/discontinuity_3d.tcc
yliang-sn: Why i write the equation?

yliang-sn:
@kks32 @thiagordonho @bodhinandach Thank you all. I fixed the code according to your suggestion. It is much better.

yliang-sn:
> [Partial review] This is great Yong.
> 
> 1. Could you please explain how you plan to handle surface/lines distributed across several MPI ranks?
> 2. Is the 3D surface always defined by triangular elements?

1. In 3D, the discontinuity will be a spatial surface and in 2D it's just a line. I prefer to copy it for each MPI ranks.
2. Yes.

yliang-sn:
> Well done with this portion of the code so far. That is a lot of work done!
> Here are a few overall comments:
> 
> 1. I also agree with Krishna's suggestions
> 2. Why is the XMPM explicit solver in a separate file?
> 3. Perhaps it would be good to standardize the nomenclature of the discontinuity functions (e.g., all with discontinuity in the end, such as `compute_momentum_discontinuity`). This is something I also want to do for the Contact algorithm but I haven't done it yet. We can use the same structure to keep the code as similar as possible.

I will do the initialization and propagation of the discontinuity then. It will be clear in a separate file.

Issue #696, opened 2020-09-22
Error in transfer_norank_particles


bodhinandach:
**Describe the bug**
I am noticing a problem with our dynamic load balancing feature or the resume feature which use the function `transfer_norank_particles` in `mesh.tcc `even in `develop`. The following screenshot is taken after running the benchmark 3D hydrostatic case with dynamic load balancing, or with resume, as they call `mpi_domain_decompose(false)` which further calls the `transfer_norank_particles`.

**To Reproduce**
1. Compile `develop` using MPI
2. Run with `mpirun -n xx mpm` any problem in benchmark
3. Activate resume to true or add `"nload_balance_steps": 10` in `mpm.json` to fasten the dynamic load balancing.
4. See the error.

**Expected behavior**
We should not have arbitrary number of ptype as otherwise we cant do resume nor dynamic load balancing feature.

**Screenshots**
![image](https://user-images.githubusercontent.com/37140224/93889203-2084c700-fd13-11ea-87a7-339c78ddf9a6.png)

As you can see above, even though that most of the `ptype` received is correct, equal to 1 for 3D particles, there are some time that it receive a weird number, in this case, 1456803152, and thus, we cant retrieve the appropriate type with `ParticleTypeName` map. As indicated, there is a PMIX ERROR while receiving the particle.

**Runtime environment (please complete the following information):**
 - OS/Docker image: Ubuntu 20.04
 - Branch: develop

**Additional context**
This is the same reason why our nightly fails since #689 is merged. From today's nightly build of running 2D sliding block:
![image](https://user-images.githubusercontent.com/37140224/93889918-fc75b580-fd13-11ea-8b03-53458f7dc4c1.png)

PR #697, opened 2020-09-23
Increase MPI buffer to avoid errors in sending

kks32:
**Describe the PR**

On domain decomposition, `transfer_nonrank_particles` gave an error when running on multiple MPI tasks. It resulted in corrupted particle data.

image

The buffer point in `MPI_Isend` ran out of scope/space. When using a thread-safe `MPI_Ibsend` instead of `MPI_Isend` the following error happened. 
```
Abort(67117313) on node 0 (rank 0 in comm 0): Fatal error in PMPI_Ibsend: Invalid buffer pointer, error stack:
```
Which clearly shows the error is because of running out of buffer space.

The MPI buffer space is increased in the main function to avoid this error.

**Related Issues/PRs**
#696

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,7 @@ workflows:
             - clang
             - cppcheck
             - codecov
+            - benchmarks
     nightly:
         jobs:
             - benchmarks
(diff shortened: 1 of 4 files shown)


bodhinandach (approved):
Checked and that works smoothly! Thanks @kks32

cbgeo (approved):

thiagordonho (approved):

PR #698, opened 2020-09-28
[Hotfix] Deserialize particle bug

bodhinandach:
**Describe the PR**
I found two major bugs in the new particle deserialization function. The first one is that `mass_density_` is not computed from mass and volume as in hdf5 initialization, and therefore equals `0`. The navier stokes and two phase solvers will have an error, while since it is not used in the explicit solver, we didn't realize this bug. The second one is, if material with state variables is used, the current version of deserialization cannot initialize the `svars` properly in the new particle. We need to add the [0] while unpacking the svars vector. The current test and benchmarks are using linear elastic, and therefore, this error was not detected, however, mohr-coulomb problems show a segmentation fault as attached below.

**Related Issues/PRs**
PR #689 

**Additional context**
See the screenshot before and after the modification of the `svars` below (at step 8975). This is a Mohr-Coulomb dam break problem.
Before:
image
After:
image

I added tests for both the mass density and the state variables, i.e. changing from LinearElastic3D to Newtonian3D that has pressure as the state variables.

### DIFF
--- a/include/particles/particle.tcc
+++ b/include/particles/particle.tcc
@@ -1082,6 +1082,9 @@ void mpm::Particle<Tdim>::deserialize(
   // volume
   MPI_Unpack(data_ptr, data.size(), &position, &volume_, 1, MPI_DOUBLE,
              MPI_COMM_WORLD);
+  // mass density
+  this->mass_density_ = mass_ / volume_;
+
   // pressure
   double pressure;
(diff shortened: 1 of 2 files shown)


kks32 (approved):

cbgeo (approved):

kks32:
Thanks @bodhinandach Great catch!

bodhinandach:
@kks32 I spent the whole weekend to debug this :sob: but I feel good after fixing it :smile: Thanks for the quick review @kks32

Issue #695, opened 2020-09-15
Compilation issue with KaHIP


tianchiTJ:
**Describe the bug**
The compile with MPI has an issue based on the current documentation. The `cmake ` is incompleted.

**To Reproduce**
1. Compile the current version 'KaHIP'

The command `sh ./compile_withcmake.sh` will give an error message as follows:
```
./compile_withcmake.sh: 33: [[: not found
./compile_withcmake.sh: 37: [[: not found
./compile_withcmake.sh: 41: [[: not found
```

It should be change to only `./compile_withcmake.sh`, and then the compile of 'KaHIP' will be OK.

2. Compile the  'MPM' with MPI

The command `cmake -DCMAKE_BUILD_TYPE=Release -DKAHIP_ROOT=~/workspace/KaHIP/ -DHALO_EXCHANGE=On ..` will give an error message as follows:

```
-- The CXX compiler identification is GNU 9.3.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake (found version "1.71.0") found components: filesystem system 
-- Found Eigen3: /usr/include/eigen3 (Required is at least version "2.91.0") 
-- Found MPI_CXX: /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi_cxx.so (found version "3.1") 
-- Found MPI: TRUE (found version "3.1")  
-- The C compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- HDF5: Using hdf5 compiler wrapper to determine CXX configuration
[...]

PR #700, opened 2020-10-07
:hammer: Fix libkahip location in cmake

kks32:
**Describe the PR**
Fix issue of unable to find KaHIP library using `FindKaHIP.cmake`

**Related Issues/PRs**
https://github.com/cb-geo/mpm/issues/695
https://github.com/KaHIP/KaHIP/issues/59

**Additional context**
`5594f1d` (copies `libinterface.so` to `deploy/libkahip.so`), causing find_library in your FindKaHIP.cmake to find the shared library rather than the static one, hence the test case cannot be run because the deploy/ directory is probably not in your `LD_LIBRARY_PATH`.

### DIFF
--- a/cmake/FindKaHIP.cmake
+++ b/cmake/FindKaHIP.cmake
@@ -46,12 +46,12 @@ if (MPI_CXX_FOUND)
     DOC "Directory where the KaHIP header files are located"
     )
 
-  find_library(KAHIP_LIBRARY kahip
+  find_library(KAHIP_LIBRARY libkahip.a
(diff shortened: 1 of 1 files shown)


bodhinandach:
Thanks @kks32 does this work only with the newest version of kahip, or also the old one? @tianchiTJ would you be able to check this?

kks32:
@bodhinandach It was moved in this version `5594f1d`. We aim to support the latest version.

bodhinandach:
Awesome! Does that mean that we have to update our kahip to the latest one @kks32?

kks32:
@bodhinandach works for both old and newer versions, so no need to update. Recommended but not required.

bodhinandach (approved):

cbgeo (approved):

Issue #692, opened 2020-08-20
Earthquake MPM


ezrayst:
# Earthquake MPM

## Summary
In light of doing earthquake simulations using the MPM, features such as absorbing boundary and input ground motion have to be implemented. Validations can be done using Linear Elastic model though in the future, more advanced cyclic constitutive models are hoped to be implemented within the CB-Geo code.

## Design Detail
There are three components to make an Earthquake MPM simulation:
- Input ground motion in terms of either acceleration-time or stress-time (Joyner and Chen, 1975; Graff 1991)
- Absorbing boundary (Lysmer and Kuhlemeyer, 1969)
![image](https://user-images.githubusercontent.com/22064721/90726943-5d265400-e277-11ea-94d0-0118fcfbd475.png)

### Input ground motion in terms of acceleration-time on the node
This will be the most straight forward of all the implementations. A discrete ground motion consisting of the time and the acceleration is saved within `constraints` such as follow:
```
  bool assign_nodal_acceleration_constraints(
      const std::vector<std::tuple<mpm::Index, unsigned, int, double>>&
          acceleration_constraints);
```
This will replace the computed acceleration through the first principle. A `time_history.h` class is derived from the abstract `function_base.h`. Usually this math function is used to yield a value between 0 and 1, which will be multiplied by the prescribed traction or velocity. However, here, the prescribed acceleration is 1 where the `time_history.h` defines the acceleration given the time. In addition, another alternative input file will be implemented to read a file containing time and value (in this case it is acceleration).

An alternative is to utilize the linear math function and define the acceleration accordingly. More specialized class to contain the ground motion acceleration time history is preferred.

### Input ground motion in terms of stress-time on the node
Joyner and Chen (1975)  proposed to have a flexible base model where only soil is modeled instead of rock, where earthquake initiates. This calls for a transformation from the acceleration or velocity (kinematics) in the rock to stress induced in the soil. The applied stresses are now governed by these equations which are pre-processed to only stresses.
![image](https://user-images.githubusercontent.com/22064721/90726003-dc1a8d00-e275-11ea-9b83-1cf589cea05d.png)

In terms of the implementation in the code, this is analogous to the traction boundary. The stresses will be translated into normal and shear forces on the nodes with `time_history.h` class defining the time history as discussed in the previous subsection.

### Absorbing boundary
[...]

PR #701, opened 2020-11-27
Add math function from file #692

ezrayst:
**Describe the PR**
Math function is modified to read acceleration time history of an earthquake, thus better to have it from a file instead of simply a json entry. 

**Related Issues/PRs**
Part 1 of #692 

**Explanation**
This design utilizes the available class structures already available. It mirrors the velocity boundary condition, with input being
```
node_id    dir    velocity_value / acceleration_value
``` 
with the variation over time defined in the `math_function`. With that, the acceleration will be kept at `1.0` in `node` and `math_function` defines the ground motion. The advantage is the robustness as we now have acceleration boundary condition as input, and it can be modified easily to allow for dynamic inputs. This method is preferred, but there may be extra double at each constrained node.

`math_function` can be defined through a csv file as shown here: https://ezrayst.github.io/mpm-doc/#/user/preprocess/input?id=math-functions

**Alternative**
To have another class, `timehistory.h` to define acceleration and stress over time. Those can have public functions such as `apply_acceleration_time()` that can be called in our solver (`mpm`) and applied directly instead of having a static value in `node`. This will mean that we have separate json entry called `dynamic_input` such as follow:
```
"dynamic_input": {
    "type": "rigid" (or "flexible"),
    "rigid_acceleration_time": "acc_time_filename.csv",
    "flexible_stress_time": "stress_time_filename.csv"
}
```


include/io/io_mesh_ascii.tcc
```diff
@@ -530,3 +530,47 @@ std::vector<std::tuple<mpm::Index, unsigned, double>>
...
+        boost::algorithm::trim(line);
+        std::istringstream istream(line);
+        // ignore comment lines (# or !) or blank lines
+        if ((line.find('#') == std::string::npos) &&
+            (line.find('!') == std::string::npos) && (line != "")) {
+          while (istream.good()) {
+            // x value
+            double x_value;
+            // fx value
+            double fx_value;
```
bodhinandach: You might want to put these two lines outside the while loop so that they are not generated and deleted multiple times.

bodhinandach (approved):
Looks okay to me, @ezrayst, one minor comment only. A small question: could you let me know which lines assign these math_functions to nodal variables acceleration/velocity/displacement? Or is it coming in the near future? Thanks btw!

ezrayst:
It should be coming next, but it is briefly described here #692, @bodhinandach

include/io/io_mesh_ascii.tcc
ezrayst: I prefer that with it being empty, because we are reading `istream`

jgiven100:
@ezrayst are acceleration time histories typically only 1d?

include/io/io_mesh_ascii.tcc
bodhinandach: I don't think we need them to be empty. I modified the other `istream` read in all the read `io_mesh_ascii` functions in PR #680 as suggested by @kks32.

ezrayst:
@jgiven100, in 2D we usually only use the horizontal acceleration (not vertical), so earthquake input will only be in one direction. In 3D, we have two lateral directions.

But maybe that is not your question (I might be reading too much into this). If we need to run horizontal and vertical acceleration time history, we will have math_function_0 for the x axis and math_function_1 for the y axis.

include/io/io_mesh_ascii.tcc
ezrayst: Fair enough! Done and thanks!

jgiven100:
>we will have math_function_0 for the x axis and math_function_1 for the y axis

@ezrayst Right, I was thinking about the 3D case. Seperate math functions for the 2 directions understood :+1:

jgiven100 (approved):

kks32 (changes requested):
Instead of writing a custom csv reader, it would be better to use https://github.com/ben-strasser/fast-cpp-csv-parser

Also, the input JSON file structure is needed. Please add more details to the PR describing the design and input file structure.

ezrayst:
Hi @kks32, two things:
- The input file and json input are available here: https://github.com/cb-geo/mpm-doc/pull/53, so please take a look. The overall design is described in #692, thus this PR is just the first point (make it easier and quicker for review). 
- For the parser, do you want to change it for the whole of our code? This PR is consistent currently with the rest of the code. If you want to change the `io` to use `ben-strasser`, I could do that too. Would that be a separate PR?

kks32:
@ezrayst 

* It's good to have PRs which are self-contained, so adding input details to the PR description helps to understand how the file would look like and the JSON configuration too since mpm-doc is a separate repository.  Additionally, RFC #692 is inconsistent with the PR: 
```
A time_history.h class is derived from the abstract function_base.h. Usually this math function is used to yield a value between 0 and 1, which will be multiplied by the prescribed traction or velocity. However, here, the prescribed acceleration is 1 where the time_history.h defines the acceleration given the time. In addition, another alternative input file will be implemented to read a file containing time and value (in this case it is acceleration).
```

No `time_history.h` is defined in this PR, so adding input files and design here helps.

* I'd suggest using the csv parser in io, you can just do it for the time-history files.

ezrayst:
@kks32  I updated the initial comment, and I will change to CSV within today, hopefully. Thanks!

ezrayst:
@kks32, please let me know if you would like to discuss the structure of the code, but currently I have changed to csv file input. I think I see you are concerned about the size of the file - but I think typical earthquake ground motion would be in the order of thousands of entries, not millions. Thanks anyways!

ezrayst:
Hi @kks32, do you still have more comments?

include/solvers/mpm_base.h
```diff
@@ -27,6 +27,9 @@
 #include "particle.h"
 #include "vector.h"
 
+// CSV-parser
+#include "csv.h"
```
kks32: Should be included in `io` class not mpm base

kks32 (changes requested):
create a folder inside external for `csv` and then add their license as well. Follow the convention.

include/solvers/mpm_base.tcc
```diff
@@ -762,16 +762,44 @@ bool mpm::MPMBase<Tdim>::initialise_math_functions(const Json& math_functions) {
       const std::string function_type =
           function_props["type"].template get<std::string>();
 
+      // Initiate another function_prop to be passed
+      auto function_props_update = function_props;
```
kks32, line 766: Why do we need this?

include/solvers/mpm_base.tcc
```diff
@@ -762,16 +762,44 @@ bool mpm::MPMBase<Tdim>::initialise_math_functions(const Json& math_functions) {
...
+
+      // Create a file reader
+      const std::string io_type =
+          io_->json_object("mesh")["io_type"].template get<std::string>();
+      auto reader = Factory<mpm::IOMesh<Tdim>>::instance()->create(io_type);
+
+      // Math function is specified in a file, replace function_props_update
```
kks32, line 773: Read files in `io` class not in `mpm_base`. Create a separate class in `io` or use the `read_mesh_ascii` to read these files. I'd prefer for now if you use `read_mesh_ascii` to read the input file.

include/solvers/mpm_base.tcc
```diff
@@ -762,16 +762,44 @@ bool mpm::MPMBase<Tdim>::initialise_math_functions(const Json& math_functions) {
...
+
+        function_props_update["xvalues"] = xvalues;
+        function_props_update["fxvalues"] = fxvalues;
+      }
+
       // Create a new function from JSON object
       auto function =
           Factory<mpm::FunctionBase, unsigned, const Json&>::instance()->create(
-              function_type, std::move(function_id), function_props);
+              function_type, std::move(function_id), function_props_update);
```
kks32, line 796: why is it called `function_props_update`?

include/solvers/mpm_base.tcc
```diff
@@ -762,16 +762,44 @@ bool mpm::MPMBase<Tdim>::initialise_math_functions(const Json& math_functions) {
...
+
+      // Create a file reader
+      const std::string io_type =
+          io_->json_object("mesh")["io_type"].template get<std::string>();
+      auto reader = Factory<mpm::IOMesh<Tdim>>::instance()->create(io_type);
+
+      // Math function is specified in a file, replace function_props_update
+      if (function_props.find("file") != function_props.end()) {
+        // Make separate arrays
+        std::vector<double> xvalues;
```
kks32, line 776: Having separate vectors makes it decoupled between x and y values, use a vector of arrays instead.

kks32 (changes requested):
Could you please add a test that shows you can read the files and get the correct x and fx values?

kks32:
@ezrayst This PR is a generic PR for reading x and fx values from a CSV file and has nothing to do with acceleration time histories per se. Please update your PR details accordingly, you may see it will be used for acceleration time histories, but there is no need to discuss how it will be implemented at the node level because that's not done here. Provide details specific to this PR, which is missing (an input JSON file and an example CSV file would help). These are the core details and should be included rather than a link to mpm-doc PR.

ezrayst:
Hi @kks32. I will update the PR accordingly then, to just simply update the `math_function`. I will make a separate PR regarding Earthquake one, but can we discuss this quickly sometime this week?

include/solvers/mpm_base.tcc
ezrayst: OK

include/solvers/mpm_base.h
ezrayst: Got it!

ezrayst: Should we move `json` and `catch` to their own respective folders?

include/solvers/mpm_base.tcc
ezrayst: That's what I have tried to use in the beginning, and you requested CSV. Can we agree on which one to use now before I make more changes?

kks32: I'm not asking you to revert back, this is an `io` operation and reading the CSV should be in the `io` class as an additional function and not in the `mpm_base`

kks32 (changes requested):

include/solvers/mpm_base.tcc
ezrayst: I see, so still CSV but within `read_mesh_ascii`. But doesn't it refer to `ascii` not `csv`?

kks32:
#715 supersedes this PR. Closing

PR #702, opened 2020-12-14
Refactor resume to improve performance

kks32:
**Describe the PR**
This PR refactors resume function calls so that the mpm code resumes quickly. The original design used to read and initialize all the particles, do a load balancing and delete all the non-local particles. Once this was done, HDF5 file was read to re-initialize the particles with the data from the previous step. This redundant process makes loading of particles and resuming extremely slow. 

When `resume` is enabled, the code now directly reads the HDF5 particles data and creates the relevant particles. This allows for faster loading of resume and removes the redundant particle data movements and deletion.

**Related Issues/PRs**
https://cb-geo.discourse.group/t/the-cost-per-timestep-after-resuming-increases/211/4

**Additional context**
This PR is to improve the run-time of Oso landslide.

### DIFF
--- a/cmake/FindKaHIP.cmake
+++ b/cmake/FindKaHIP.cmake
@@ -46,12 +46,12 @@ if (MPI_CXX_FOUND)
     DOC "Directory where the KaHIP header files are located"
     )
 
-  find_library(KAHIP_LIBRARY kahip
+  find_library(KAHIP_LIBRARY libkahip.a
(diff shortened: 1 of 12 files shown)


yliang-sn:
I tested it with the OSO landslide case (more than 4 Million particles). The resume speed is good and the results look correct.

yliang-sn (approved):

yliang-sn (approved):

cbgeo (approved):

PR #703, opened 2020-12-15
Add Absorbing Boundary #692

cgeudeker:
**Describe the PR**
Adding an absorbing boundary for earthquake simulations.

**Related Issues/PRs**
Part 2 of #692 (Most of the below information can be found here)

**Additional context**
Dashpots provide normal and shear traction along boundaries to dissipate the ground motion. Traction values are computed according to the following equations from Lysmer and Kuhlemeyer (1969). 

traction

Traction values are included in the governing equation for calculating acceleration. The direction input refers to the normal direction.

```
bool assign_absorbing_constraint(unsigned dir, double pwave_v, double swave_v);
```

### DIFF
--- a/include/node.h
+++ b/include/node.h
@@ -209,6 +209,9 @@ class Node : public NodeBase<Tdim> {
   bool assign_friction_constraint(unsigned dir, int sign,
                                   double friction) override;
 
+  bool assign_absorbing_constraint(unsigned dir, int sign,
+                                  double friction) override;
+
   //! Apply friction constraints
   //! \param[in] dt Time-step
   void apply_friction_constraints(double dt) override;
--- a/include/node.h
+++ b/include/node.h
@@ -202,12 +202,13 @@ class Node : public NodeBase<Tdim> {
   void apply_velocity_constraints() override;
 
   //! Assign absorbing constraint
-  //! \param[in] pwave_v p wave velocity
-  //! \param[in] swave_v s wave velocity
-  bool assign_nodal_absorbing_constraint(double pwave_v,double swave_v);
+  //! \param[in] pwave_v P-wave velocity
+  //! \param[in] swave_v S-wave velocity
+  bool assign_absorbing_constraint(unsigned dir, double pwave_v,
+                                   double swave_v) override;
 
   //! Apply absorbing constraint
-  void apply_nodal_absorbing_constraint(double dt);
+  void apply_absorbing_constraint(double dt) override;
 
   //! Assign friction constraint
   //! Directions can take values between 0 and Dim * Nphases
--- a/include/node.h
+++ b/include/node.h
@@ -208,7 +208,7 @@ class Node : public NodeBase<Tdim> {
                                    double swave_v) override;
 
   //! Apply absorbing constraint
-  void apply_absorbing_constraint(double dt) override;
+  void apply_absorbing_constraint() override;
 
   //! Assign friction constraint
   //! Directions can take values between 0 and Dim * Nphases
@@ -309,7 +309,9 @@ class Node : public NodeBase<Tdim> {
   //! Velocity constraints
   std::map<unsigned, double> velocity_constraints_;
   //! Absorbing Constraints
-  std::map<unsigned, double> absorbing_constraints_;
+  std::tuple<unsigned, double, double> absorbing_constraints_;
+  //! Absorbing Traction
+  Eigen::Matrix<double, Tdim, Tnphases> absorbing_traction_;
   //! Rotation matrix for general velocity constraints
   Eigen::Matrix<double, Tdim, Tdim> rotation_matrix_;
   //! Material ids whose information was passed to this node
--- a/include/node.h
+++ b/include/node.h
@@ -201,15 +201,20 @@ class Node : public NodeBase<Tdim> {
   //! Apply velocity constraints
   void apply_velocity_constraints() override;
 
+  //! Assign absorbing constraint
+  //! \param[in] pwave_v p wave velocity
+  //! \param[in] swave_v s wave velocity
+  bool assign_nodal_absorbing_constraint(double pwave_v,double swave_v);
+
+  //! Apply absorbing constraint
+  void apply_nodal_absorbing_constraint(double dt);
+
   //! Assign friction constraint
   //! Directions can take values between 0 and Dim * Nphases
   //! \param[in] dir Direction of friction constraint
   //! \param[in] sign Sign of normal wrt coordinate system for friction
   //! \param[in] friction Applied friction constraint
   bool assign_friction_constraint(unsigned dir, int sign,
-                                  double friction) override;
-
-  bool assign_absorbing_constraint(unsigned dir, int sign,
                                   double friction) override;
 
   //! Apply friction constraints
@@ -302,6 +307,8 @@ class Node : public NodeBase<Tdim> {
   Eigen::Matrix<double, Tdim, Tnphases> acceleration_;
   //! Velocity constraints
   std::map<unsigned, double> velocity_constraints_;
+  //! Absorbing Constraints
+  std::map<unsigned, double> absorbing_constraints_;
   //! Rotation matrix for general velocity constraints
   Eigen::Matrix<double, Tdim, Tdim> rotation_matrix_;
   //! Material ids whose information was passed to this node
--- a/include/node.tcc
+++ b/include/node.tcc
@@ -16,6 +16,7 @@ mpm::Node<Tdim, Tdof, Tnphases>::Node(
 
   // Clear any velocity constraints
   velocity_constraints_.clear();
+  absorbing_constraints_.clear();
   concentrated_force_.setZero();
   this->initialise();
 }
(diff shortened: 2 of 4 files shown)


jgiven100:
@kks32 Maybe this would be a good PR to start pushing to a new earthquake branch?

include/node.h
```diff
@@ -299,6 +308,10 @@ class Node : public NodeBase<Tdim> {
   Eigen::Matrix<double, Tdim, Tnphases> acceleration_;
   //! Velocity constraints
   std::map<unsigned, double> velocity_constraints_;
+  //! Absorbing Constraints
+  std::tuple<unsigned, double, double> absorbing_constraints_;
+  //! Absorbing Traction
```
kks32, line 313: This variable is not needed

include/node.h
```diff
@@ -201,6 +201,15 @@ class Node : public NodeBase<Tdim> {
   //! Apply velocity constraints
   void apply_velocity_constraints() override;
 
+  //! Assign absorbing constraint
+  //! \param[in] pwave_v P-wave velocity
```
kks32, line 205: p-wave and s-wave should be material properties too, not nodal properties.

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+bool mpm::Node<Tdim, Tdof, Tnphases>::assign_absorbing_constraint(
+    unsigned dir, double pwave_v, double swave_v) {
+  bool status = true;
+  try {
+    if (dir < Tdim) {
+      this->absorbing_constraints_ = std::make_tuple(
+          static_cast<unsigned>(dir), static_cast<double>(pwave_v),
+          static_cast<double>(swave_v));
+    } else
+      throw std::runtime_error("Constraint direction is out of bounds");
```
kks32, line 364: Please be specific on the constraint

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
   }
 }
 
+//! Assign absorbing constraints
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+bool mpm::Node<Tdim, Tdof, Tnphases>::assign_absorbing_constraint(
```
kks32, line 355: Let's change the function to void

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+    unsigned dir, double pwave_v, double swave_v) {
+  bool status = true;
+  try {
+    if (dir < Tdim) {
```
kks32, line 359: Check this condition as assert

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
   }
 }
 
+//! Assign absorbing constraints
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+bool mpm::Node<Tdim, Tdof, Tnphases>::assign_absorbing_constraint(
+    unsigned dir, double pwave_v, double swave_v) {
+  bool status = true;
+  try {
```
kks32, line 358: Remove the try and catch, but throw on dir > Tdim

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
   }
 }
 
+//! Assign absorbing constraints
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+bool mpm::Node<Tdim, Tdof, Tnphases>::assign_absorbing_constraint(
+    unsigned dir, double pwave_v, double swave_v) {
```
kks32, line 356: compression and shear wave are material parameters and can't be passed as functions to nodes.

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+
+  // Phase: Integer value of division (dir / Tdim)
+  const unsigned phase = 0;
+
+  // Create Traction Value
+  double traction;
+
+  for (unsigned dir = 0; dir < Tdim; ++dir) {
+    // Get mass density and velocity
+    double mass_density = mass_(phase) / volume_(phase);
```
kks32, line 393: Move this outside the for-loop

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+    double velocity = velocity_(dir, phase);
+
+    // Calculate Traction Forces
+    if (dir == dir_n) {
+      traction = -velocity * mass_density * pwave_v;
+    } else {
+      traction = -velocity * mass_density * swave_v;
+    }
+    // Apply traction forces
+    absorbing_traction_(dir, phase) = traction;
```
kks32, line 403: This doesn't have to be a private member variable, can be local to the function.

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+  double traction;
+
+  for (unsigned dir = 0; dir < Tdim; ++dir) {
+    // Get mass density and velocity
+    double mass_density = mass_(phase) / volume_(phase);
+    double velocity = velocity_(dir, phase);
+
+    // Calculate Traction Forces
+    if (dir == dir_n) {
+      traction = -velocity * mass_density * pwave_v;
```
kks32, line 398: I'd suggest adding a dimensionless damping co-efficient.

include/node.tcc
```diff
@@ -346,6 +350,62 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+  // Extract S-wave velocity
+  const double swave_v = std::get<2>(this->absorbing_constraints_);
+
+  // Phase: Integer value of division (dir / Tdim)
+  const unsigned phase = 0;
+
+  // Create Traction Value
+  double traction;
+
+  for (unsigned dir = 0; dir < Tdim; ++dir) {
```
kks32, line 391: Instead of a for loop create an `Eigen::Matrix<double, Tdim, 1> wave_velocity = swave_velocity` and set the normal direction to pwave_velocity then use a cwiseProduct to do the multiplication so you don't need a for loop.

kks32: You probably need absorbing boundaries on all phases

include/node.tcc
kks32, line 398: How do we know it's a +/- normal direction?

tests/node_test.cc
```diff
@@ -683,6 +683,93 @@ TEST_CASE("Node is checked for 2D case", "[node][2D]") {
...
+            direction_normal, velocity_pwave, velocity_swave));
+
+        // Calculate mass density
+        double mass_density = node->mass(Nphase) / node->volume(Nphase);
+
+        // Apply absorbing constraint
+        node->apply_absorbing_constraint();
+
+        // Define traction variables
+        double traction_normal;
```
kks32, line 733: No need to predefine variables

tests/node_test.cc
```diff
@@ -683,6 +683,93 @@ TEST_CASE("Node is checked for 2D case", "[node][2D]") {
       // TODO Assert: REQUIRE_NOTHROW(node->update_external_force(false, 1,
       // force_bad));
 
+      SECTION("Check absorbing constraint") {
+        // Set external force to 0
```
kks32, line 687: Could you use an already existing function, so you don't have to redefine and check all the momenta and forces?

kks32 (changes requested):
@cgeudeker Thanks for your first PR. Nice work! 

If the boundary nodes are only supported by the Lysmer-Kuhlmeyer dashpots, it would result in creep because of the stresses from the soil body and will continue to deform. I would suggest adding a spring + dashpot to avoid this effect similar to a Kelvin-Voigt type boundary. Thanks for the unit test. However, I suggest a benchmark problem, create a linear elastic solid block supported on just the dashpots on one end and free on the other. At the free surface, apply an instantaneous vertical load and see what happens to the displacements at the dashpot end, you'll notice creep. This issue needs to be fixed to be able to use a good absorbing boundary.

ezrayst:
Hi @kks32, thanks so much for your thoughtful comments! Really appreciate it. We are making the benchmark as well, but do you have a reference to a problem? I am struggling to make a problem because the answer is not quite analytical. In addition, does it mean you do not want the unit test (I hope you still want it :P). Thanks once again!

kks32:
Hi @ezrayst What benchmark problem are you creating? Could you please elaborate. 
It would be good to see the results of an elastic bar subjected to an instantaneous load, like you mentioned in #692 

I was thinking a linear elastic rectangular block which is 1 m thick and 1 m long supported on dashpots. Apply an instantaneous load and see how the displacement gets damped. You can compare the results against (a) fixed displacement boundary (in this case the stress will get doubled, once the wave reflects), (b) dashpot-alone, I expect this to creep, and (c) spring-dashpot system.

include/node.h
cgeudeker: Hi, so I've been able to understand and make most of your other suggestions barring this one. How would I be able to access these values from the material properties? I had thought it would be through ```property_handle_``` but I'm currently getting an error about segmentation violation when I run ./mpmtest and it calls the function where I do this. Perhaps this is the right method but I need to initialize the properties somewhere. Thanks for your help

kks32:
Closing this PR as #703 supersedes this PR.

Issue #706, opened 2021-02-17
Print the latest git commit's SHA1 upon starting a simulation


SchDvr:
# Print the latest git commit's SHA1 upon starting a simulation

## Summary

This change would allow the user to see exactly which version of CB-Geo is running by printing to the console the latest commit's SHA1 before anything else. The first line displayed by CB-Geo would look like that :

```
[2021-02-17 00:09:02.219] [main] [info] git revision: 1fa330e756a986e332bffef224395d33f8a75d4e
```  

## Motivation

This feature would help users organizing their results and thus improve the reproducibility of their work. It should also be useful when developing a new feature. 

## Design Detail

This change would use [this project](https://github.com/andrew-hardin/cmake-git-version-tracking) which was developed by Andrew Hardin for this purpose. It is also under a MIT License. 

Three files would be added and one would be created by make :
 - `cmake/git_watcher.cmake`, to add 
 - `include/git.h`, to add
 - `include/git.cc.in`, to add
 - `include/git.cc`, generated from `include/git.cc.in` by make

The `CMakeLists.txt` file would be modified to include `cmake/git_watcher.cmake` but also to create and link a new library named `git` (as it is done in the [project mentioned above](https://github.com/andrew-hardin/cmake-git-version-tracking)).

The line that prints the commit's SHA1 would be added in the `src/main.cc` script, right after the [initialization of the logger](https://github.com/cb-geo/mpm/blob/6c6578ae8a7e04fa45971261a2253af7405e4542/src/main.cc#L35).

## Drawbacks

The addition of several files may reduce the code's readability.

## Rationale and Alternatives

The two main alternatives that can be found with a google search are summed up [here](https://github.com/andrew-hardin/cmake-git-version-tracking#wait-doesnt-this-already-exist). Those alternatives either do not see the changes made after running cmake or needs to recompile some files even though no changes were made. 
So according to Andrew Hardin, his solution is better.

jgiven100:
Hi @SchDvr 

Thanks for opening this. The [example from Hardin](https://github.com/andrew-hardin/cmake-git-version-tracking/tree/master/better-example) looks quite straight forward. I agree this would be good in terms of reproducibility and helpful information to include. 

Two thoughts:

* `develop`  should always be the "best" version of the code. So if old results aren't being reproduced on the current `develop HEAD`, the old results probably have some errors and need to be rerun anyway. 

[...]

PR #707, opened 2021-02-19
Feature/print_git_revision

SchDvr:
**PR description**
Add the feature discussed [here](https://github.com/cb-geo/mpm/issues/706): the last git SHA1 is printed when a simulation starts.
Here is an example of the first line that CB-Geo displays with these changes : 

```
[2021-02-19 10:34:20.894] [main] [info] git revision: e4c44e702eca1d5ecb667573a60fee852a119357
```   

**Additional context**
This changes use [Andrew Hardin's project](https://github.com/andrew-hardin/cmake-git-version-tracking) that was developed for this purpose.


.gitignore
```diff
@@ -15,6 +15,8 @@
 !*/src/*
 !*/include/*
 !*/tests/*
+!cmake/git_watcher.cmake
```
kks32: Could you please remove these. They are needed in case we change the files.

kks32 (changes requested):
Thanks for the PR. 

Could you please run from the `build` directory: `sh ../clang-tools/format.sh` to make sure the test passes.

.gitignore
SchDvr: > Could you please remove these. They are needed in case we change the files.

I added these lines because git wasn't tracking the files otherwise. I guess another line made git ignore them so I forced git to **not** ignore them using a "!" in front of their paths.
So I guess these lines should stay right ? Or maybe the lines that made git ignore them should be tweaked ?

SchDvr:
The build fails because of the clang-format: 
 1. in `include/git.cc.in` the clang-format script puts a space in the middle of some variables that begin and end with the `@` character. This makes the generation of `include/git.cc` fail. 
 2. even when these spaces are manually removed, some automatically generated lines in `include/git.cc` get too long and are no longer respecting the clang-format. 
 
Is there a way to make the clang-format script behave differently for these files?

.gitignore
kks32: You could do `git add` those specific files without putting it in gitignore. I can try and fix the clang-format error for you

kks32:
Hi @SchDvr I pushed some changes to your repo, this removes `git.cc` because it will get automatically generated when building so there is no need to add this everytime. So I have ignored it by doing `include/git.cc` in `.gitignore`. Also added a flag to turn-off format checking on this file. Hopefully, this should fix all the build issues. Thanks for your contribution!

kks32 (approved):
I'm happy with this. Thanks @SchDvr !!

jgiven100 (approved):

jgiven100:
@SchDvr and @kks32 

Thanks all for working on these improvements

SchDvr:
Thanks for solving the clang-format problem and for merging the branch.

PR #705, opened 2021-02-10
Add shear and compression wave velocities to Linear Elastic Material Properties for #692

cgeudeker:
**Describe the PR**
Add shear and compression wave velocities to linear elastic material properties related to #692 .

**Related Issues/PRs**
Related to #692, specifically Kelvin-Voigt boundary.

**Additional context**
The following gives information that can also be found in the following [paper](https://elib.uni-stuttgart.de/bitstream/11682/513/1/Thesis_Issam_FinalPrint.pdf ). Kelvin-Voigt boundary elements operate as a dashpot, spring system according to the following equations.

KV_boundary_eqs

Equations for calculating P-wave (c_p) and S-wave (c_s) velocities:

wave_velocities

 is the constrained modulus, calculated as 

Equations for calculating corresponding spring coefficients:

spring_coeffs

Additionally, constrained modulus and shear modulus were calculated. `layer_thickness` was the only value that could not be calculated by the already existent material properties.

Input File Format:

P-wave and S-wave velocities are automatically calculated using values already defined in linear elastic properties. The values are then recorded into the JSON material properties object

```
    properties_["pwave_velocity"] = vp_;
    properties_["swave_velocity"] = vs_;
```

PR #710, opened 2021-03-24
Kelvin-Voigt Boundary Implementation for #692

cgeudeker:
**Describe the PR**
Kelvin-Voigt boundary implementation for #692 

**Related Issues/PRs**
This PR aims to implement the Kelvin-Voigt boundary, or more specifically the spring-dashpot system, as a boundary conditions. This utilizes linear elastic material properties, as added by PR #705, as well as json file input parameters (shown in the Additional context section) to calculate traction values for the following equations:

KV_boundary_eqs

Spring coefficients `kp` and `ks` are calculated according to the following equations:

spring_coeffs

Other relevant to this PR, are the parameters `delta`, `a` and `b`. Each of these will be user input values under the `boundary conditions` as shown in the additional context section below. `a` and `b` are optional input parameters, which if not specified will be equivalent to a value of 1. However, the following relation must be satisfied:

thickness_check

where `h_min` is taken to be the cell height.

**Additional context**
`input.json` file is currently expected to look like the following
```
"boundary_conditions": {
...
   "absorbing_boundary": [
              {
                  "nset_id" : #,
                  "dir" : #,
                  "delta": #,
                  "a": #,
                  "b": #
              }
          ],
...
},

```

The above implementation is based on the work of this [paper](https://elib.uni-stuttgart.de/bitstream/11682/513/1/Thesis_Issam_FinalPrint.pdf).


include/io/io_mesh_ascii.h
```diff
@@ -86,6 +86,12 @@ class IOMeshAscii : public IOMesh<Tdim> {
       read_friction_constraints(
           const std::string& friction_constraints_file) override;
 
+  //! Read absorping constraints file
```
kks32:
```suggestion
  //! Read absorbing constraints file
```

include/io/io_mesh.h
```diff
@@ -98,6 +98,12 @@ class IOMesh {
       read_friction_constraints(
           const std::string& friction_constraints_file) = 0;
 
+  //! Read absorbing constraints file
+  //! \param[in] absorbing_constraints_files file name with tractions
```
kks32:
```suggestion
  //! \param[in] absorbing_constraints_files file name with constrains
```

include/io/io_mesh_ascii.tcc
```diff
@@ -486,6 +486,57 @@ std::vector<std::tuple<mpm::Index, unsigned, int, double>>
   return constraints;
 }
 
+//! Return absorbing constraints of particles
+template <unsigned Tdim>
```
kks32: Why do you need this function, when you already have a, b, delta in the input JSON file?

include/loads_bcs/constraints.h
```diff
@@ -3,6 +3,7 @@
 
 #include <memory>
 
+#include "absorbing_constraint.h"
```
kks32, line 6: This file is missing!

include/loads_bcs/constraints.h
```diff
@@ -48,6 +49,19 @@ class Constraints {
       const std::vector<std::tuple<mpm::Index, unsigned, int, double>>&
           friction_constraints);
 
+  //! Assign nodal absorbing constraints
+  //! \param[in] setid Node set id
+  //! \param[in] absorbing_constraints Constraint at node, dir, delta, a, b
```
kks32: naming doesn't match the documentation

include/loads_bcs/constraints.h
```diff
@@ -48,6 +49,19 @@ class Constraints {
...
+  bool assign_nodal_absorbing_constraint(
+      int nset_id,
+      const std::shared_ptr<mpm::AbsorbingConstraint>& aconstraints);
```
kks32:
```suggestion
      const std::shared_ptr<mpm::AbsorbingConstraint>& absorbing_constraints);
```

include/node.h
```diff
@@ -213,6 +213,19 @@ class Node : public NodeBase<Tdim> {
   //! \param[in] dt Time-step
   void apply_friction_constraints(double dt) override;
 
+  //! Assign absorbing constraint
+  //! Directions can take values between 0 and Dim * Nphases
+  //! \param[in] dir Direction of absorbing constraint
+  //! \param[in] delta Virtual viscous layer thickness
+  //! \param[in] a Equation constant
```
kks32: Please be more descriptive

include/node.tcc
```diff
@@ -345,6 +345,78 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
   }
 }
+//! Assign absorbing constraints
+template <unsigned Tdim, unsigned Tdof, unsigned Tnphases>
+bool mpm::Node<Tdim, Tdof, Tnphases>::assign_absorbing_constraint(
+    unsigned dir, double delta, double a, double b, double h_min) {
+  bool status = true;
+  try {
+    assert(dir <= Tdim);
+    if (delta >= h_min / (2 * a) and delta >= h_min / (2 * b)) {
```
kks32:
```suggestion
    if (delta >= h_min / (2 * a) && delta >= h_min / (2 * b)) {
```

kks32: Can't you make this check at input stage and ask the mean cell length as an input, because this should be known to the user?

include/node.tcc
```diff
@@ -345,6 +345,78 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+  // a coefficient
+  const double a = std::get<2>(this->absorbing_constraint_);
+
+  // b coefficient
+  const double b = std::get<3>(this->absorbing_constraint_);
+
+  // Get material id
+  auto mat_id = material_ids_.begin();
+
+  // Extract material properties and displacements
```
kks32: You may want to check if the node has only one material, otherwise this would be incorrect...

include/node.tcc
```diff
@@ -345,6 +345,78 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+  double pwave_v = this->property_handle_->property("pwave_velocity", prop_id_,
```
kks32: When do we map this information?

include/node.tcc
```diff
@@ -345,6 +345,78 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+  double k_s = (density * pow(swave_v, 2)) / delta;
+  double k_p = (density * pow(pwave_v, 2)) / delta;
+  Eigen::Matrix<double, Tdim, 1> spring_constant =
+      Eigen::MatrixXd::Constant(Tdim, 1, k_s);
+  spring_constant(dir, 0) = k_p;
+
+  // Iterate through each phase
+  for (unsigned phase = 0; phase < Tnphases; ++phase) {
+    // Calculate Aborbing Traction
+    Eigen::Matrix<double, Tdim, 1> absorbing_traction_ =
```
kks32: We don't need to store this variable, right?

include/node.h
```diff
@@ -306,6 +319,10 @@ class Node : public NodeBase<Tdim> {
   //! A general velocity (non-Cartesian/inclined) constraint is specified at the
   //! node
   bool generic_boundary_constraints_{false};
+  //! Absorbing constraints
```
kks32: Could we pass this instead of storing it in the node class? As this would increase the size of each node class by a 256 bytes unnecessarily.

include/node.h
```diff
@@ -306,6 +319,10 @@ class Node : public NodeBase<Tdim> {
...
+  std::tuple<unsigned, double, double, double> absorbing_constraint_;
+  //! Absorbing traction
+  Eigen::Matrix<double, Tdim, 1> absorbing_traction_;
```
kks32: I don't think we need to store this variable in node either

include/solvers/mpm_base.tcc
```diff
@@ -937,6 +940,62 @@ void mpm::MPMBase<Tdim>::nodal_frictional_constraints(
...
+              constraints.at("file").template get<std::string>();
+          bool absorbing_constraints =
+              constraints_->assign_nodal_absorbing_constraints(
+                  mesh_io->read_absorbing_constraints(
+                      io_->file_name(absorbing_constraints_file)));
+          if (!absorbing_constraints)
+            throw std::runtime_error(
+                "Absorbing constraints are not properly assigned");
+
+        } else {
```
kks32: This is not needed, let's not read it from a file

kks32 (changes requested):
Add unit tests and please fix as per suggestions

cgeudeker:
I wasn't able to get the unit test done today, but I wanted to make sure that some of the other edits I was able to make today were what was expected since I was slightly unsure, particularly with regards to `apply_absorbing_boundary`. I also still wasn't quite sure where to map the new material information `pwave_velocity` and `swave_velocity` onto the nodal properties. I thought I had understood it to be in `mpm_explicit.tcc` but I couldn't find anything similar.

jgiven100:
@cgeudeker Would it work to follow along with `nodal_velocity_constraints()` in `mpm_base.tcc` [here](https://github.com/cb-geo/mpm/blob/develop/include/solvers/mpm_base.tcc#L234)? 

Within mpm_base it calls 
```
constraints_->assign_nodal_velocity_constraints()
```
then in constraints.tcc it calls
```
mesh_->nodes(set_id)->assign_velocity_constraint()
```
at which point the velocity constaints are saved as nodal properties. For the absorbing boundary case, we would need to define 2 new variables in `node.h` to save the pwave and swave values.

cgeudeker:
A few notes about the last commit.

I see that I need to put together a test case that will go through the absorbing constraint file, so I'll look at what was done for the others and follow along one of those. I'll probably commit this later today.

There is an issue that resurfaced while finalizing the test case for the mapping of wave velocities to nodes. When 2 particles of the same material were mapped to the nodes, the wave velocities doubled. I thought this would be fixed by having the update_property boolean argument set to false, but the same result occurred. So if there is any advice about how to best solve this it would be much appreciated. 

Finally, some of the sections are poorly commented, so that will be fixed in the next couple of commits, but I wanted to make sure that the function was working as desired.

include/node.tcc
```diff
@@ -346,6 +347,63 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+
+      // Iterate through each phase
+      for (unsigned phase = 0; phase < Tnphases; ++phase) {
+        // Calculate Aborbing Traction
+        Eigen::Matrix<double, Tdim, 1> absorbing_traction_ =
+            this->velocity_.col(phase).cwiseProduct(wave_velocity) * density +
+            material_displacement.cwiseProduct(spring_constant);
+
+        // Update external force
+        this->update_external_force(true, phase, -absorbing_traction_);
```
jgiven100: @cgeudeker @kks32 
`absorbing_traction_` is going to be [ force / (length^2) ] but we should be passing just [ force ] to `update_external_force(..)`. 

Using `h_min` and unit thickness into the page I think we can easily go from traction -> force. But we need to think about how to handle the corner nodes where the nodal influence is only `h_min / 2`

include/node.tcc
```diff
@@ -346,6 +347,63 @@ void mpm::Node<Tdim, Tdof, Tnphases>::apply_velocity_constraints() {
...
+      double pwave_v = this->property_handle_->property(
+          "wave_velocities", prop_id_, *mat_id, Tdim)(0);
+      double swave_v = this->property_handle_->property(
+          "wave_velocities", prop_id_, *mat_id, Tdim)(1);
+      double density =
+          this->property_handle_->property("density", prop_id_, *mat_id)(0);
+      Eigen::Matrix<double, Tdim, 1> material_displacement =
+          this->property_handle_->property("displacements", prop_id_, *mat_id,
+                                           Tdim);
+
```
jgiven100: Based on current implementation, these nodal displacements are weighted by the particle mass. We need to divide by nodal mass. Something simple like
```
material_displacement /= this->mass(*mat_id);
```
will do the trick

jgiven100:
@kks32 2 mains issues have come up. Guidance on these would be appreciated. 

# 1: Nodal location affects how to go from traction -> force

## Problem

The thesis provides an equation of traction for the absorbing boundary. We need to apply this as a nodal force. ~~Based on the nodes location (corner, side, or bottom), the different components of the traction must be multiplied by either half or full cell lengths to become force. Below are the 3 cases to consider.~~
image

## Proposed solution

Add `"position" :` to input JSON to indicate if the specific node set is for the model's `"corner"`, `"side"`, or `"bottom"`. For example:
```
      "absorbing_constraints" : [
        { 
          "nset_id" : 0,
          "dir" : 1,
          "delta" : 0.25,
          "h_min" : 0.001,
          "a" : 1.0,
          "b" : 1.0,
          "position" : "corner"
        }
      ]
```
Then update `apply_absorbing_constraint(...)` to handle the 3 possible cases. 

# 2: How to deal with material points displacing beyond the absorbing boundary

## Problem 
If `delta` is very small, then wave energy at the boundary will be reflected instead of absorbed. So `delta` must be somewhat large to be effective. But this also means that material points will likely displace beyond the absorbing boundary. If the mesh also extends beyond the absorbing boundary, the model will continue to run, but this breaks conservation of mass within the sub-domain defined by the absorbing boundary. 

1. Is extending the mesh beyond the absorbing boundary an accetable work-around?
2. Is conservation of mass within the subdomain defined by the absorbing boundary necessary?
3. If extending mesh beyond the absorbing boundary is ok, would that affect the area of influence for going from traction to force (problem above)? (i.e., displacement and velocity will be mapped to the node form inside and outside the boundary, but the traction would still be only applied to the partial cell within the boundary.) 

*side note: I don't think Q3 is important for geotechnical applications, but rather for keeping the code general. Absorbing boundaries are likely to only be applied to the sides of models with lateral damping being a priority. Area of influence for lateral damping will be along vertical cell edges which wouldn't be affected by material points outside the absorbing boundary.*


# EDITS:
*  (Nov. 23, 2021): cross out wrong scheme

include/solvers/mpm_base.h
```diff
@@ -207,6 +216,15 @@ class MPMBase : public MPM {
   double damping_factor_{0.};
   //! Locate particles
   bool locate_particles_{true};
+  //! Absorbing Boundary Variables
+  bool absorbing_boundary_{false};
+  std::shared_ptr<mpm::AbsorbingConstraint> absorbing_constraint_;
+  unsigned absorbing_dir_{0};
+  int absorbing_nset_id_{0};
```
jgiven100, line 223: Maybe it would be good to make this a vector of node set ids so we can have more than 1 absorbing boundary

include/solvers/mpm_base.tcc
```diff
@@ -937,6 +940,67 @@ void mpm::MPMBase<Tdim>::nodal_frictional_constraints(
...
+
+// Apply nodal absorbing constraints
+template <unsigned Tdim>
+void mpm::MPMBase<Tdim>::nodal_absorbing_constraints() {
+  bool absorbing_constraints = constraints_->assign_nodal_absorbing_constraint(
+      absorbing_nset_id_, absorbing_constraint_);
+  if (!absorbing_constraints)
+    throw std::runtime_error(
+        "Nodal absorbing constraint is not properly assigned");
+}
```
jgiven100, line 1002: If absorbing_nset_id_ is a vector of ids (see above comment), then we need to loop each id here

jgiven100:
@kks32 I've been able to locally implement the proposed solution described above (adding `"position"` to boundary inputs). I'll wait to hear from you if this is an acceptable solution before pushing to the feature branch.

kks32:
@jgiven100 That's a good solution. Thanks!!

include/loads_bcs/absorbing_constraint.h
```diff
@@ -15,9 +15,17 @@ class AbsorbingConstraint {
   //! \param[in] h_min Cell height
   //! \param[in] a Equation constant
   //! \param[in] b Equation constant
+  //! \param[in] position Nodal position along boundary
   AbsorbingConstraint(int setid, unsigned dir, double delta, double h_min,
-                      double a = 1, double b = 1)
-      : setid_{setid}, dir_{dir}, delta_{delta}, h_min_{h_min}, a_{a}, b_{b} {};
+                      double a = 1., double b = 1.,
+                      std::string position = "null")
```
kks32: @jgiven100 could we enum the position labels instead of string? C++ class enums would be the best. Thanks!

kks32 (changes requested):

jgiven100:
@kks Seems like there might have been a Clang update? Clang is failing [here](https://github.com/cb-geo/mpm/blob/develop/include/cell.tcc#L415). Seems to have been passing before. 
image

include/loads_bcs/absorbing_constraint.h
jgiven100: @kks32 is adding the enum class to data_types.h ok?

kks32:
> @kks Seems like there might have been a Clang update? Clang is failing [here](https://github.com/cb-geo/mpm/blob/develop/include/cell.tcc#L415). Seems to have been passing before.
> image

I thought it was `cppcheck` not clang:

```
include/node.tcc:404:37: warning: Found suspicious operator ',' [constStatement]
            absorbing_force_ << dir0, dir1;
                                    ^
include/node.tcc:409:37: warning: Found suspicious operator ',' [constStatement]
            absorbing_force_ << dir0, dir1;
 ```
 
 You can turn-off clang-format temporarily using
 ```
     // clang-format off
     ...code...
     // clang-format on
 ```

File: include/loads_bcs/absorbing_constraint.h (hunk 19 above)
Comment (kks32) [reply include/loads_bcs/absorbing_constraint.h]: Yes that would be a good place to add enum class

File: include/node.tcc [hunk 20]
```diff
@@ -397,19 +397,25 @@ bool mpm::Node<Tdim, Tdof, Tnphases>::apply_absorbing_constraint(
...
+              force1 = 0.5 * h_min * absorbing_traction_(1);
+              break;
+            case mpm::Position::Bottom:
+              force0 = 0.5 * h_min * absorbing_traction_(0);
+              force1 = 1.0 * h_min * absorbing_traction_(1);
+              break;
+            case mpm::Position::Side:
+              force0 = 1.0 * h_min * absorbing_traction_(0);
+              force1 = 0.5 * h_min * absorbing_traction_(1);
+              break;
```
jgiven100: @kks32 @cgeudeker I am longer convinced this scheme is correct. Now, my understanding is that the "side" and "bottom" nodes should simply be
```
force = h_min * absorbing_traction_
```
I was incorrectly thinking about the applied tractions in terms of a frame-like problem instead of a continuum problem. 

If others agree - I will update this section (& greatly simplify whats going on here).

jgiven100: Fixed

include/io/io_mesh_ascii.h
jgiven100: outdate - marked as resolved

include/io/io_mesh.h
jgiven100: outdated - marked as resolved

include/io/io_mesh_ascii.tcc
jgiven100: outdated - marked as resolved

include/solvers/mpm_base.tcc
jgiven100: outdated - resolved

include/loads_bcs/constraints.h
jgiven100: outdated - resolved

include/node.h
jgiven100: `absorbing_constraint_` previously removed

include/node.h
jgiven100: `absorbing_traction_` removed

include/loads_bcs/absorbing_constraint.h
jgiven100: Resolved

include/node.h
jgiven100: more descriptive names added (to be merged later today)

jgiven100:
@kks32 @cgeudeker 
## Feature Updates
* I've locally updated `node.tcc` to handle 3D models. 
## Verification 
* With a 3D column, I've been able to reproduce 1D p-wave results (from Al-Kafaji thesis). 
* I'll work on upward moving shear wave next. Once reproducing the 2D results, I'll push changes here (next few days). 
## Next Steps
- [x] Current 2D benchmark https://github.com/cb-geo/mpm-benchmarks/pull/20 needs some slight updates due to changes in input file naming convention. 
- [x] Add identical 3D benchmark to the above open bencmark PR. 
- [x] Open mpm-doc PR.


## Edits
(24 Nov. 2021) - make next steps into check boxes

include/node.tcc
jgiven100: correct. No need to save this

include/loads_bcs/constraints.h
jgiven100: naming updated. Resolved.

include/node.tcc
jgiven100: Moved.

jgiven100:
@kks @cgeudeker 
This is ready for a re-review. 

I'm happy with unit testing coverage for 18 of 21 changed files. Expections are `mpm_base.tcc`, `mpm_scheme.tcc`, and `mpm_explicit.tcc`. Looking at similar existing functions for the exception files, I don't see too much existing testing... This leads me to believe that the areas with low coverage in this update is ok.

PR #716, opened 2021-07-02
Feature/contact

thiagordonho:
**Describe the PR**
Implementation of the contact-related functions in the `Node` and `Particle` classes and definition of the `ContactFriction` class. These entities were incorporated to `MPMExplicit` and are currently able to deal with frictional contact between two distinct bodies.

**Related Issues/PRs**
This PR is related to RFC #636 .

**Additional context**
This PR closely follows the design detail presented in RFC #636 . In fact, it comprehends the implementations of the **contact-related functions** to the `Node` and `Particle` classes. The functions in the particles are responsible for mapping the properties from the material points to the nodes (started with `map_multimaterial`) and updating the particle's kinematics using the acceleration and velocity from nodes (`compute_contact_updated_position`). The functions in the nodes are responsible for computing the additional nodal properties (`compute_multimaterial_velocity`, `compute_multimaterial_normal_unit_vector`, `compute_multimaterial_relative_velocity`, `compute_contact_acceleration_velocity`) that are required for resolving contact at contact nodes using `apply_contact_mechanics`.

To validate the contact algorithm implemented, two 2D benchmark problems were implemented: a block sliding down an inclined plane, and a cylinder rolling down an inclined plane. Both problems were modeled in the horizontal position with an inclined gravity applied to simulate the inclination of the plane. Both models also considered the Maximum Velocity Gradient, or MVG, as the normal computation type (see RFC #636 ). They were tested for different friction coefficients and angles of inclination. The results for the position of the center of mass were compared to that of the analytic solutions for rigid block and cylinder cases. Figures 1 and 2 show a schematic of the model from both problems and Figures 3 to 6 show the obtained results. As shown in Figure 2, the cylinder was modeled at 1 cell above the plane because the approximation condition would not handle the rolling properly if the cylinder is directly touching the plane. Other approximation conditions must be implemented to solve this issue.

image
Figure 1: Sliding block model in its initial position.

image
Figure 2: Rolling cylinder model in its initial position.

image
Figure 3: Horizontal position of the center of mass of the sliding block for a 30-degree inclination.

image
Figure 4: Horizontal position of the center of mass of the sliding block for a 45-degree inclination.

image
Figure 5: Horizontal position of the center of mass of the rolling cylinder for a 30-degree inclination.

image
Figure 6: Horizontal position of the center of mass of the rolling cylinder for a 45-degree inclination.

### DIFF
--- a/include/contacts/contact.h
+++ b/include/contacts/contact.h
@@ -17,8 +17,28 @@ class Contact {
   //! Intialize
   virtual inline void initialise(){};
 
+  //! Compute nodal kinematics
+  virtual inline void compute_nodal_kinematics(){};
+
   //! Compute contact forces
-  virtual inline void compute_contact_forces(){};
+  //! \param[in] gravity Gravity vector
+  //! \param[in] phase Index to indicate material phase
+  //! \param[in] time Current time in the simulation
+  //! \param[in] concentrated_nodal_forces Boolean for if a concentrated force
+  //! is applied or not
+  virtual inline void compute_contact_forces(
+      const Eigen::Matrix<double, Tdim, 1>& gravity, unsigned phase,
+      double time, bool concentrated_nodal_forces){};
+
+  //! Compute contact nodal kinematics
+  //! \param[in] dt Timestep in analysis
+  virtual inline void compute_contact_kinematics(double dt){};
+
+  //! Update particle position
+  //! \param[in] dt Timestep in analysis
+  //! \param[in] velocity_update Update particle velocity from nodal vel
+  virtual inline void update_particles_contact(double dt,
+                                               bool velocity_update){};
 
  protected:
   //! Mesh object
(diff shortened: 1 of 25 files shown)


jgiven100:
@thiagordonho
Thanks for all of the figures on this. For the cylinder cases, the limiting friction angle and coefficient are given as:
```
\phi_L = atan( sin( \alpha / 3 ) )
\mu_L = tan( atan( sin( \alpha / 3 ) ) )
\mu_L = sin( \alpha / 3 )
```
For the examples you show, this implies the following (please let me know if I'm thinking about this correctly):
## 45 Deg. Incline
```
\mu_L = sin( 45 / 3 )
\mu_L = 0.26
```
so `\mu=0.2` is "sliding" while `\mu=0.4` is "rolling". 

## Question
For the cylinder + 45 deg. incline, I think its good that the `\mu = 0.2` case (red dash, Figure 6) aligns with the analytical solution like the sliding block outputs since in this case it is just sliding, like the blocks. I expect most contact scenarios we will model with CB-Geo will be a sliding mechanism, so great to catpure this so well!

But I'm a bit suprised the "rolling" cases are over-predicting displacement. I would have expected rolling to under-predict. Do you have any ideas why numerical outputs always appear to over-predict here?

Issue #714, opened 2021-06-02
Update NorSand for Jefferies and Been (2019)


jgiven100:
# Update NorSand for Jefferies and Been (2019)

## Summary

Current NorSand formulation is based primarily on Jefferies (1993) with some additions from Jefferies and Shuttle (2012). The 2nd edition of _Soil Liquefaction_ (Jefferies and Been, 2019) provides updates for NorSand. These updates include the introduction of soil state-dilatancy property `\chi`, simplified critical friction ratio `M`, and changes to volumetric coupling property `N`.

Additional internal parameters include `\psi_i`, `\chi_i`, and `M_i` or the state paramter image, state-dilatancy image, and critical state ratio image. The resulting new yield surface is "bullet" shaped, similar to original Cam Clay. 

![image](https://user-images.githubusercontent.com/62029065/120545087-0a9bcb80-c3a3-11eb-8193-4381e5bd5393.png)

## Motivation

NorSand was initially published with Nova's rule:
```
D^p = ( M_{tc} - \eta ) / ( 1 - N )
```
but Dafalias has shown that M_{tc} is strain dependent. Thus stress-dilatancy should be updated to:
```
D^p = M_i - \eta 
```
NorSand should be updated prior to exploring anisotropic stresses within the model as briefly discussed here -> https://github.com/cb-geo/mpm/discussions/713

## Design Detail

Generally, the existing NorSand material class will be updated based on the improved yield surface and hardening rule. This will require:

### New Private Variable
`chi_image_` will be added to the material construtor (beneath `Mtc_`). This will be a constant since `chi_`, `lambda_`, and `Mtc_` are all currently implemented as constants. 
```
chi_image_ = chi_ / (1. - ((chi_ * lambda_) / Mtc_))
```

### New State Parameters
Both `M_imgae` and `psi_image` will be added to the `state_vars` dense_map. They will be initialized as 
```
// M_image
{"M_image", 0.},
 // State parameter image
{"psi_image", void_ratio_initial_ - (gamma_ - lambda_ * log(p_image_initial_ / reference_pressure_))}
```

### New Functions
`M_image` and `psi_image` will be updated via the following new function.
```
//! Compute image parameters
template <unsigned Tdim>
void mpm::NorSand<Tdim>::compute_image_parameters(mpm::dense_map* state_vars,
                                                  const double void_ratio,
                                                  const double e_image,
                                                  const double M_theta) {
  // Compute state parameter image
[...]

PR #717, opened 2021-07-13
Feature/norsand updates

jgiven100:
## Describe the PR
NorSand formulation is updated to be aligned with Jefferies and Been (2019) as described previously here -> https://github.com/cb-geo/mpm/issues/714 . Internal formulation changes include updates to the yield function. There are no proposed changes for user-facing components of the NorSand model. 

Test cases are updated for the new formulation. 

In addition to the NorSand updates, this PR contains a small refactoring of the material utility function `dp_dsigma(...)`. As mentioned in the above issue, passing a vector of the current stress state should not be required. Thus, passing this vector is removed. 

## Validation
A simple FEM code is availabe from the publishers ([found here](https://www.routledge.com/Soil-Liquefaction-A-Critical-State-Approach-Second-Edition/Jefferies-Been/p/book/9781482213683)). Updates to the CB-Geo implementation of NorSand are compared to the FEM code for a suite of triaxial compression tests. These tests include:
```
* Undrained TX,  psi_0 = -0.05
* Undrained TX,  psi_0 = +0.05
* Undrained TX,  psi_0 =  0.00
* Drained TX,    psi_0 = -0.05
* Drained TX,    psi_0 = +0.05
* Drained TX,    psi_0 =  0.00
```
Below is an example of the validation for the undrained, dense sample.
```
* solid black == FEM code
* dashed red == updated CB-Geo MPM code
* dashed purple == previous CB-Geo MPM code
```

image

The other 5 validation plots are are posted [here]().

## Remaining work
Documentation will be updated prior to merging in `develop`. 

Following this NorSand improvement, anisotropic elastic stiffness will be considered. 

### Edit:
Here is the mpm-doc PR -> https://github.com/cb-geo/mpm-doc/pull/62

### DIFF
--- a/include/materials/material_utility.h
+++ b/include/materials/material_utility.h
@@ -43,10 +43,8 @@ inline double lode_angle(
     double tolerance = std::numeric_limits<double>::epsilon());
 
 //! Compute derivative of p in terms of stress sigma
-//! \param[in] stress Stress in Voigt notation where positive is tension
 //! \retval dp_dsigma Derivative of p in terms of stress sigma
-inline const Eigen::Matrix<double, 6, 1> dp_dsigma(
-    const Eigen::Matrix<double, 6, 1>& stress);
+inline const Eigen::Matrix<double, 6, 1> dp_dsigma();
 
 //! Compute derivative of q in terms of stress sigma
 //! \param[in] stress Stress in Voigt notation where positive is tension
--- a/tests/materials/norsand_test.cc
+++ b/tests/materials/norsand_test.cc
@@ -1,6 +1,7 @@
 #include <limits>
 
 #include <cmath>
+#include <iostream>
 
 #include "Eigen/Dense"
 #include "catch.hpp"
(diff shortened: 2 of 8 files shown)


jgiven100:
@ezrayst I would appreciate if you took some time to look at the proposed updates!

include/materials/material_utility.h
```diff
@@ -43,10 +43,8 @@ inline double lode_angle(
     double tolerance = std::numeric_limits<double>::epsilon());
 
 //! Compute derivative of p in terms of stress sigma
-//! \param[in] stress Stress in Voigt notation where positive is tension
 //! \retval dp_dsigma Derivative of p in terms of stress sigma
-inline const Eigen::Matrix<double, 6, 1> dp_dsigma(
-    const Eigen::Matrix<double, 6, 1>& stress);
+inline const Eigen::Matrix<double, 6, 1> dp_dsigma();
```
kks32, line 47: Why do we not have to pass the stress vector here?

tests/materials/norsand_test.cc
```diff
@@ -1,6 +1,7 @@
 #include <limits>
 
 #include <cmath>
+#include <iostream>
```
kks32:
```suggestion
```

Please remove iostream

kks32 (changes requested):
Thanks @jgiven100 a couple of minor changes

include/materials/material_utility.h
bodhinandach: I think that's just because `dp_dsigma` returns a constant vector independent of the input stress vector.

bodhinandach (commented):
I am okay as long as @ezrayst is ok. Thanks @jgiven100

include/materials/material_utility.h
jgiven100: Agreed with @bodhinandach. `dp_dsigma` is defined as `1/3 * identity matrix` so it'll always be a constant. As seen in our code -> [here](https://github.com/cb-geo/mpm/blob/develop/include/materials/material_utility.tcc#L83-L93)

jgiven100: I was double-sure this refactoring is ok when the Mohr-Coulomb and MCC tests continued to pass after refactoring without any updates to those specific test files

tests/materials/norsand_test.cc
jgiven100: @kks32 removed in latest commit

kks32 (approved):

include/materials/material_utility.h
bodhinandach: @jgiven100 I think you need to refactor the rests too. I think MCC and Mohr-Coulomb call this specific function.

jgiven100: @bodhinandach alread did in [this commit](https://github.com/jgiven100/mpm/commit/39012f01f81f93dcf21f4418fcfbd4906cb56b49). Otherwise I don't think all of the test cases would pass since there would be compiler errors

bodhinandach: Okay, I will be waiting for @ezrayst's response, otherwise, I am good!

cbgeo (approved):

PR #721, opened 2021-09-29
:computer: Add a repartition flag  to resume using manual h4 files

kks32:
**Describe the PR**
When using a manually written particles HDF5 files and resuming from step 0, it will typically call `mesh_->resume_domain_cell_ranks()` which then assigns the current MPI_RANK to the cell if the cell has particles in it. This works on the assumption that the current cell has particles only on the current rank - which is true when generated using our MPM code, but not for particles generated manually and written to HDF5 file. When particles are distributed across different MPI ranks for the same cell, MPI_Reduce will result in MPI_RANK larger than the MPI_SIZE (bad!) - assigning to a different MPI rank as long as it's within MPI_SIZE is bad for efficiency, which will be resolved in the subsequent step.

This PR adds a boolean flag called `repartition: true` to `resume` JSON to enable repartitioning when resuming from a HDF5 file. This is an optional flag which is set to false. 

**Additional context**
Input file change:

```json
"analysis": {
		"resume": {
			"resume": true,
			"repartition": true,
			"uuid": "slope",
			"step": 0
		}
}
```

### DIFF
--- a/include/mesh.tcc
+++ b/include/mesh.tcc
@@ -954,8 +954,10 @@ template <unsigned Tdim>
 void mpm::Mesh<Tdim>::resume_domain_cell_ranks() {
   // Get MPI rank
   int mpi_rank = 0;
+  int mpi_size = 0;
 #ifdef USE_MPI
   MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
+  MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
   const unsigned rank_max = std::numeric_limits<unsigned>::max();
   const unsigned ncells = this->ncells();
   // Vector of cell ranks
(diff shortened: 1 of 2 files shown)


include/mesh.tcc
```diff
@@ -954,8 +954,10 @@ template <unsigned Tdim>
 void mpm::Mesh<Tdim>::resume_domain_cell_ranks() {
   // Get MPI rank
   int mpi_rank = 0;
+  int mpi_size = 0;
```
cbgeo, line 957: Not needed anymore!

cbgeo (changes requested):

cbgeo (approved):

jgiven100 (approved):
`bool` flag seems like a simple (and non-breaking) fix for this

PR #731, opened 2022-10-11
:bug::wrench: fix sign error for dtheta_dsigma in NorSand

jgiven100:
**Describe the PR**
I found a bug related to the `dtheta_dsigma` sign I had included in the previous NorSand implemention. 

Previously we were passing *compression positive* stresses from NorSand -> `dtheta_dsigma(...)`, BUT `dtheta_dsigma(...)` returns the same Vector6d regardless of being passed *compression positive* or *tension positive* stresses. So this is the incorrect way to account for the incompatibility of sign conventions between NorSand and material utility. 

Therefore, we need a `-1.` coefficient infront of `dtheta_dsigma(...)` to account for incompatibility of 2 sign conventions

**Numerically Approximated Validation**
Consider following stress state (tension positive):
```
-1.5
-1.6
-1.7
-1.8
0.0
0.0
```

Lode angle is computed via material utility (cosine convention): `0.475541833917983`

Then change the stress state a bit (tension positive):
```
-1.5 - 1e-10
-1.6
-1.7
-1.8
0.0
0.0
```

Lode angle is computed via material utility (cosine convention): `0.47554183393394`

Using approximate derivatives for the first term in `dtheta_dsigma` gives
```
dtheta_dsigma(0) = (0.47554183393394 - 0.475541833917983) / 1e-10 
dtheta_dsigma(0) = 0.1595701348833245
```
which approximately matches the first value of the proposed `-1. * mpm::materials::dtheta_dsigma(stress)` in terms of magnitude and sign
```
dtheta_disigma(0) : 0.159573
dtheta_disigma(1) : 0.160313
dtheta_disigma(2) : -0.319887
dtheta_disigma(3) : 0.0266367
dtheta_disigma(4) : -0
dtheta_disigma(5) : 0
```
Subsequent terms can be likewise computed. 

**Related Issues/PRs**
N/a

**Additional context**
N/a

### DIFF
--- a/include/materials/norsand.tcc
+++ b/include/materials/norsand.tcc
@@ -369,7 +369,7 @@ void mpm::NorSand<Tdim>::compute_plastic_tensor(const Vector6d& stress,
       (3. * std::pow(Mtc_, 2) * sin(3. / 2. * lode_angle)) / (2. * (3. + Mtc_));
 
   // Compute dtheta / dsigma
-  const Vector6d dtheta_dsigma = mpm::materials::dtheta_dsigma(stress);
+  const Vector6d dtheta_dsigma = -1. * mpm::materials::dtheta_dsigma(stress);
 
   // dF_dsigma is in compression negative
   const Vector6d dF_dsigma =


kks32 (approved):
Thank Joel! This looks good :)

PR #742, opened 2023-12-07
Extending Mohr Coulomb state variables

jduriez:
**Description**
This PR includes new state variables for the Mohr-Coulomb material:

- 4 (binary) flags specifying the type of failure (if any). It can now be distinguished whether the material is currently (or has ever) failed in shear and/or tension

- stored elastic energy. Note that it is denoted `E_el`, should it be rather `eEl` ?

- cumulated plastic work, i.e. dissipation, given as a positive number. (well, let's not argue here whether this really is a state variable from a theoretical perspective ;-) ). Which is currently `W_pl` (should it be rather `wPl` ?)

**Additional context**
- I was about to modify mpm-doc as well to document this change, but have not yet fully decided about the most appropriate place to do so. Probably in source corresponding to https://mpm.cb-geo.com/#/theory/material/mohr-coulomb, in a new (final) section "Output state variables" ?

- This code comes from @SchDvr work during [his PhD thesis](https://theses.hal.science/tel-04101270). There is quite a number of other updates that could be proposed to be applied following his work, but this should probably be further discussed elsewhere.

### DIFF
--- a/include/materials/mohr_coulomb.tcc
+++ b/include/materials/mohr_coulomb.tcc
@@ -70,7 +70,19 @@ mpm::dense_map mpm::MohrCoulomb<Tdim>::initialise_state_variables() {
                                // Theta
                                {"theta", 0.},
                                // Plastic deviatoric strain
-                               {"pdstrain", 0.}};
+                               {"pdstrain", 0.},
+                               // Elastic energy
+                               {"E_el", 0.},
+                               // Plastic work
+                               {"W_pl", 0.},
+                               // Flag for current failure because of tensile stress
+                               {"tensile_fail_curr", 0},
+                               // Flag for current failure because of shear stress
+                               {"shear_fail_curr", 0},
+                               // Flag for past or current failure because of tensile stress
+                               {"tensile_fail", 0},
+                               // Flag for past or current failure because of shear stress
+                               {"shear_fail", 0}};
   return state_vars;
 }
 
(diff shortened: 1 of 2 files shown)


jduriez:
Sorry for the initial check failure, my local test took place in a more compliant Ubuntu 22.04 environment (where brace initializing `Vector6d` was possible)

kks32 (approved):

kks32 (changes requested):
Thank you so much @jduriez for the suggested updates. Could we add some mpm-benchmark tests to make sure the previous version will still work and any new tests for this modified MC?

jduriez:
> Thank you so much @jduriez for the suggested updates. Could we add some mpm-benchmark tests to make sure the previous version will still work and any new tests for this modified MC?

Thanks for feedback @kks32 ! Do you mean "mpm-benchmark tests" like in [mpm-benchmark](https://github.com/cb-geo/mpm-benchmarks) ? If yes, I'm afraid that would be too much for us at the moment (and for possibly a long time due to lack of time..) but I just tried to include a greater consideration of these new state variables in the catch test `tests/materials/mohr_coulomb_test.cc` as here:

https://github.com/cb-geo/mpm/blob/f4d183b3c2aebb9d16c7cb2bb1ee6c032a781491/tests/materials/mohr_coulomb_test.cc#L3392

to L3398

As for whether previous version would still work, we somehow relied on the fact existing MC tests still pass and I would mostly stress out we are (aim at, at least ! ;-) ) touching here only post-processing state variables.

Let me know what you think (I should anyway try to come with a `mpm-doc` update about MC state variables next week..)

jduriez:
For the record, the incremental elastic energy expression is actually to improve. We currently multiply here the updated stress value with elastic strain increment, which is an overestimation. Using a mid-point rule between `updated_stress` and `stress` should be much better, so maybe it is indeed better not to merge currently

